PR #63, opened 2013-02-28
Avoid organizational comments

jferris:
- Based on summary from technical research card.


jferris:
For the record, this is the Research card mentioned in the title: https://trello.com/c/QIsZgN0K

PR #146, opened 2014-03-07
Specify method chaining dot style rule

gylaz:
When splitting long lines place the dot on the same line as the method:

``` ruby
foo.bar
  .baz
```


printercu:
Just walked around and seen this issue. That's cool that you decided to use trailing dot. I haven't seen that someone mentioned the main pro. 😃

The biggest thing about this way, is that such code can be simply copy-pasted to repl, while with leading dot this can not be done without concating lines in editor.

PR #185, opened 2014-06-05
Change the restrictions for using self.

halogenandtoast:
Based on in office conversation the use of self to reference the class
has the potential benefit of handling renames in the event that the
class name needs to change. Additionally it makes for easier extraction
into a new class.


jferris:
@halogenandtoast any updates?

jferris:
I'm going to close this, because it seems to have stalled. Feel free to re-open if you're interested.

PR #215, opened 2014-08-06
Use `is_expected.to` instead of `should`

JoelQ:
Even though we've switched to the `expect` syntax, we've still been using
`should` for one-liners with an implicit subject. RSpec 3 introduced
`is_expected.to` as an alternative. Myron Marston explains:

> Some users have expressed confusion about how this should relates to the
> expect syntax and if you can continue using it. It will continue to be
> available in RSpec 3 (again, regardless of your syntax configuration), but
> we’ve also added an alternate API that is a bit more consistent with the
> expect syntax

From: [New API for
one-liners](http://myronmars.to/n/dev-blog/2014/05/notable-changes-in-rspec-3#new_api_for_oneliners_)

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -270,13 +270,13 @@ Testing
 * Separate setup, exercise, verification, and teardown phases with newlines.
 * Use RSpec's [`expect` syntax].
 * Use RSpec's [`allow` syntax] for method stubs.
-* Use `should` shorthand for [one-liners with an implicit subject].
+* Use `is_expected.to` for [one-liners with an implicit subject].
 * Use `not_to` instead of `to_not` in RSpec expectations.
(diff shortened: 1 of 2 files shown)


mcmire:
Do we want to revisit this? I'm finding more and more that people who leave issues on shoulda-matchers tend to prefer the `is_expected` syntax rather than the `should` syntax. I'm considering switching the README and documentation to use `is_expected`... but I'm hesitant to do this if we currently promote the use of `should`.

JoelQ:
We don't promote either form. The guideline was removed in https://github.com/thoughtbot/guides/pull/234. I don't see a problem with converting the shoulda-matchers README to `is_expected`, although others may disagree.

mcmire:
Oh yeah! It sure was. Cool, thanks.

PR #225, opened 2014-09-09
Add convention for time without date columns

geoffharcourt:
Rails returns times without dates from the database as `Time` objects (as a
time on the date `2000-01-01` in UTC), which can lead to confusion with
actual datetimes. Although we can't make the returned value clearer,
uniformly naming these columns with the suffix `_time` can at least remove
confusion while reading the code.

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -217,6 +217,7 @@ Rails
 * Use private instead of protected when defining controller methods.
 * Name date columns with `_on` suffixes.
 * Name datetime columns with `_at` suffixes.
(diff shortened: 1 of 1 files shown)


jferris:
@joannecheng any thoughts here? I have no objection to this.

joannecheng:
:+1: from me. @geoffharcourt and I never discussed this officially when we were working together, but I think it's great to have it established as a best practice.

jferris:
Thanks. Merged as 2038731.

PR #232, opened 2014-09-23
Include `href` or `to_param` when constructing URLs client side

sgrif:
One nice feature that Rails provides is easy changes to URL generation
(assuming the param starts with the ID). All you have to do is

```ruby
def to_param
  [id, name].to_param
end
```

and everything "Just Works". This is lost, however, if you're manually
constructing URLs on the client side. This is avoided, if as a general
rule of thumb, we let the server decide how to construct the URLs, and
pass that information along.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -187,6 +187,8 @@ JavaScript
 ----------
 * Use Coffeescript, ES6 with [babel], or another language that compiles to
   JavaScript
+* Include a `to_param` or `href` attribute when serializing ActiveRecord models,
+  and use that when constructing URLs client side, rather than the ID. Example:
 
 [babel]: http://babeljs.io/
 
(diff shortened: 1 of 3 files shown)


jferris:
@mike-burns can't it be both??!

Any further questions/objections/suggestions?

pbrisbin:
I like this guideline, but the example should be moved as Mike indicated.

jferris:
@sgrif want to move the example into the examples?

jferris:
@sgrif one last ping on updating this

sgrif:
I'll update today sorry

sgrif:
I moved the examples into separate files, but had to create said files. We did not previously have any example files for best practices, only style.

best-practices/README.md
```diff
@@ -177,6 +177,23 @@ JavaScript
 ----------
 
 * Use CoffeeScript.
+* Include a `to_param` or `href` attribute when serializing ActiveRecord models,
+  and use that when constructing URLs client side, rather than the ID. Example:
```
mcmire: Looks like you can remove "Example:" here, now.

mcmire:
:+1: I like this.

best-practices/README.md
jferris: You could also link to the example.

jferris:
@sgrif ready to wrap this up?

PR #235, opened 2014-10-10
Use double colons for pseudo-elements

penge:
Use double colons (::) for pseudo-elements, like ::before, ::after.
Use single colon (:) for pseudo-class, like :hover, :checked.

More at:
https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements
https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes
https://developer.mozilla.org/en-US/docs/Web/CSS/::before

### DIFF
--- a/style/samples/sass.scss
+++ b/style/samples/sass.scss
@@ -17,7 +17,11 @@ section.application {
     attribute: value;
   }
 
-  &:before {
+  &::before {
+    content: "<";
+  }
+
+  &::after {
     content: ">";
   }
 


jferris:
@thoughtbot/design any thoughts?

tysongach:
Yep. :+1: 

I was actually about to put up a PR for this exact thing. Thanks, @penge!

brendastorer:
Great suggestion.

corwinharrell:
Do we care that double colons are not supported in IE8?

joshuaogle:
We [don't officially support](http://playbook.thoughtbot.com/#web-apps) IE8, so I wouldn't worry about it, but that is something to keep in mind if your project has a critical need to support it.

Great addition, even if it's a hard habit to break :+1:

corwinharrell:
Ah ok, thanks @joshuaogle. Didn't even know that was in the playbook.:+1:

kaishin:
It will take some getting used to but I can do with it.

wardpenney:
Good point, I should start doing this. I notice that the syntax guide is missing an example of pseudo-selector, which would still be single-colon. :shipit:

whmii:
:love_hotel: looks good

jferris:
Thanks. I merged this into master as e28b8f7.

penge:
Happy to hear that Joe!
Wishing nice day to thoughtbot! ;)

PR #251, opened 2014-11-14
Add best practice to prefer Lo-dash/Underscore

gylaz:
There are certain utility methods that exist in both Lo-dash/Underscore as well
as in a framework like AngularJS. We should prefer the former, because they are
smaller, well-focused, widely used libraries. This will avoid us from coupling
our code to a framework.

Example:
```js
// prefer

_.merge(object1, object2);

// over

angular.extend(object1, object2);
```

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -178,7 +178,15 @@ Email
 JavaScript
 ----------
 
+* Prefer utility methods from common libraries like [Lo-Dash]/[Underscore.js]
+  over methods that do the same thing from frameworks like
+  [Angular.js]/[Ember.js].
 * Use CoffeeScript.
+
+[Lo-Dash]: https://lodash.com
+[Underscore.js]: http://underscorejs.org
+[Angular.js]: https://angularjs.org
+[Ember.js]: https://emberjs.com
 
 HTML
 ----


jferris:
Does this mean we should add something like Underscore or Lodash to every project?

gylaz:
I would say that projects that utilize JavaScript in moderate-to-heavy fashion should use it. Projects that just sprinkle some jQuery may not have a need. @sgrif Do you have any thoughts?

sgrif:
I would agree with this guideline personally. I like consistency, and we can't actually use the native versions of most of these functions on projects yet

BlakeWilliams:
I disagree from an Ember context. I can't think of anything from lodash or underscore that I would want/need that Ember doesn't already provide.

One example would be `_.map` and Ember's `map`. Ember's `map` is identical parameter wise to ES6's `Array.prototype.map`. Ember extends prototypes so these methods can be called directly on the object instead of through a namespace.

drapergeek:
:-1: for Ember. I don't see a benefit in pulling in another library. If you're using Ember you've already coupled your code to a framework and the difference in a `map` or `each` isn't going to save you much effort if you decide to move out.

gylaz:
Doesn't seem like there's a strong consensus. Closing.

Issue #250, opened 2014-11-14
Define subject when using Shoulda-Matchers' validate_uniqueness_of


c-lliope:
Can we revisit #133 with respect to https://github.com/thoughtbot/shoulda-matchers/pull/567 ?

I just ran into the Non-NULL database column problem, where `validates_uniqueness_of` didn't know how to create a valid record.

Here's what the guides suggested my specs look like:

``` ruby
  describe "Validations" do
    it "validates uniqueness of email" do
      create(:registration)

      expect(Registration.new).to validate_uniqueness_of(:email)
    end

    it "validates uniqueness of mobile phone number" do
      create(:registration)

      expect(Registration.new).to validate_uniqueness_of(:mobile_phone_number)
    end

    it { is_expected.to validate_presence_of(:email) }
    it { is_expected.to validate_presence_of(:mobile_phone_number) }
  end
```

And here's what it looks like with an explicit subject:

``` ruby
  describe "Validations" do
    subject { build(:registration) }

    it { is_expected.to validate_presence_of(:email) }
    it { is_expected.to validate_presence_of(:mobile_phone_number) }
    it { is_expected.to validate_uniqueness_of(:email) }
    it { is_expected.to validate_uniqueness_of(:mobile_phone_number) }
  end
```

I would suggest that the second option is more readable and more manageable, with less duplication. Thoughts on allowing an explicit subject for use with shoulda-matchers?

TorvaldsDB:
```ruby
require 'rails_helper'

RSpec.describe Score, type: :model do
  let(:score) { build(:score) }
  subject { described_class }

  it { is_expected.to validate_numericality_of(:score) }

  it 'belongs to staff' do
    assoc = subject.reflect_on_association(:staff)

    expect(assoc.macro).to eq :belongs_to
  end
end
```
This is my codes,  but I occurred the issues, looks like this: 
```ruby
» rspec spec/models/score_spec.rb                                                                                                                             [ruby-2.5.1][19:04:09]
F

Failures:

  1) Score should validate that :score looks like a number
     Failure/Error: it { is_expected.to validate_numericality_of(:score) }

     Shoulda::Matchers::ActiveModel::AllowValueMatcher::AttributeDoesNotExistError:
       The matcher attempted to set :score on the Class to "abcd", but that
       attribute does not exist.
```
my `Score` looks like this:
```ruby
class Score < ApplicationRecord
  belongs_to :staff

[...]

PR #253, opened 2014-11-15
use custom subjects as needed for one-liners

graysonwright:
See #250 

This PR encourages the use of custom subjects for one-line specs as opposed to breaking the specs into multiple lines.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -107,7 +107,9 @@ Testing
 -------
 
 * Avoid `any_instance` in rspec-mocks and mocha. Prefer [dependency injection].
-* Avoid `its`, `let`, `let!`, `specify`, `before`, and `subject` in RSpec.
+* Avoid `its`, `let`, `let!`, `specify`, and `before` in RSpec.
+* Avoid using `subject` explicitly *inside of an* RSpec `it` block.
+  [Example][subject-example].
 * Avoid using instance variables in tests.
 * Disable real HTTP requests to external services with
   `WebMock.disable_net_connect!`.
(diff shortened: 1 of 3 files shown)


gylaz:
My concern here is that it loosens our recommendation about avoiding `subject` in general. We **only** want to use it in a very specific case, that is less that 10% (totally arbitrary number) of test that we write.

What if we leave `subject` in `best-practices`, and treat the line in `style` as an amendment?

best-practices/README.md
```diff
@@ -107,7 +107,8 @@ Testing
 -------
 
 * Avoid `any_instance` in rspec-mocks and mocha. Prefer [dependency injection].
-* Avoid `its`, `let`, `let!`, `specify`, `before`, and `subject` in RSpec.
+* Avoid `its`, `let`, `let!`, `specify`, and `before` in RSpec.
+* Avoid using `subject` explicitly in RSpec `it` blocks.
```
graysonwright: @gylaz does this extra line help to clarify at all?

gylaz: I'm not sure that the problem is people using `subject` in `it` blocks -- is that even possible?

How about something like this?
```
Avoid using `subject` in RSpec, except for groups of one-line specs.
```

gylaz:
With the example it makes sense! Does it make sense to add that example somewhere?

Or perhaps reward to use `inside`: "Avoid using `subject` explicitly __inside of an__ RSpec `it` block."

graysonwright:
@gylaz @mcmire changes made, one last look?

gylaz:
LGTM

mcmire:
:star:

graysonwright:
Dammit! I dissociated the code from this PR again!
:angry:

Figured out why, though. Creating aliases now so it doesn't happen again.
Merged into master in c694bd0

PR #254, opened 2014-11-18
Prefer `if` over `unless`

gabebw:


jferris:
Are there any objections to moving this into the "Style" section, since the behavior is the same?

*Pedantry: you can override `!` in Ruby, so `if ! foo` and `unless foo` are technically not identical.*

teoljungberg:
Aslong as we just _prefer_ `if` over `unless`, I'll :+1:

gabebw:
If I move this as-is ("Prefer if to unless") to the style section, any objections?

mike-burns:
Works for me.

BlakeWilliams:
:+1: to style section

jferris:
@gabebw looks like this has been open for a week without objection. Ready to merge.

PR #256, opened 2014-11-21
Add guideline for using described_class

georgebrock:
The `described_class` method in RSpec returns the class passed to the outer-most `describe` block. Using it makes it significantly easier to rename classes.

I was using this in some unit specs in gitsh, but not in others. When @mike-burns and I did some restructuring and moved a bunch of classes around, it was much easier for the specs that were already using `described_class` (see: https://github.com/thoughtbot/gitsh/commit/c1dc545c21a8037f47d71bf8ca01e667433f6d6e).

Documentation: https://www.relishapp.com/rspec/rspec-core/docs/metadata/described-class

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -287,6 +287,7 @@ Testing
 * Separate setup, exercise, verification, and teardown phases with newlines.
 * Use RSpec's [`expect` syntax].
 * Use RSpec's [`allow` syntax] for method stubs.
+* Use `described_class` instead of directly referring to the class under test.
 * Use `should` shorthand for [one-liners with an implicit subject].
 * Use `not_to` instead of `to_not` in RSpec expectations.
 * Prefer the `have_css` matcher to the `have_selector` matcher in Capybara assertions.


elle:
I know this is closed and might be too late. I'm still a fan of using `described_class`

Also a few reasons @geoffharcourt likes it (from: https://github.com/thoughtbot/eds/pull/270/files#r53353178)

* it calls out that you're doing something with the system under test
* it allows you to rename the class without renaming references
* if you later extract to a shared example the code just works (it may make it easier to find duplicated test code for similar tests as well)

geoffharcourt:
I was about to get my 2c in, but @elle has done it for me.

geoffharcourt:
From my experience on projects here I've found the response to this to be very bifurcated. Developers seem to either love or hate this practice.

georgebrock:
We currently don't have a guideline either way, which is probably the right thing to do given that some people love it and some people hate it.

jferris:
> I know this is closed and might be too late.

It's never too late, @elle! Don't give up!

> Developers seem to either love it or hate it.

I was the naysayer in this pull request, but I do not hate `described_class`. I have a minor preference for explicitly referencing the class in the spec. Most of my specs end up abstracting the class through a factory method or FactoryGirl factory, though, so I think using `described_class` wouldn't affect me much.

> We currently don't have a guideline either way, which is probably the right thing to do given that some people love it and some people hate it.

I don't hate `described_class`, but it would drive me nuts to open a spec and see it used inconsistently. At the very least, I'd want each project to be either in or out. The easiest way to be consistent within a project is to be consistent across projects, so maybe a guideline would be better after all?

sikachu:
I think "be consistent" is good as well. And yeah, with constraint, sometime it make sense to use `described_class` to reduce the number of characters on the line. I think _use the best judgement_ would work here just fine.

nickrivadeneira:
There was a brief period of time I used `described_class` and found the convenience of renaming classes nice, but ultimately I prefer using the class name for the same reasons Joe describes.

It's only occasionally that I'm changing the name of a class or refactoring into a shared example, but every time I look at a spec, I'll be scanning it to get a quick understanding of what I'm looking at. Syntax highlighting of class names helps with the more common scenario.

mjankowski:
@jferris any more thoughts since you reopened this about a month ago?  Do you still want to push forward, or are you comfortable with leaving this decision as a per-project concern?

tommydangerous:
I like this

jferris:
> Do you still want to push forward, or are you comfortable with leaving this decision as a per-project concern?

I generally prefer to have a company-wide convention vs a per-project convention when we can reach a consensus. I'm not going to push this forward personally, because I don't actually agree with the changes in this pull request. I re-opened it because it was still being actively discussed.

@elle @geoffharcourt @georgebrock any further thoughts on the changes here?

geoffharcourt:
I hadn't considered @sikachu's pro-point as well about often reducing the length of the line if you have classes with long names.

I'd certainly prefer to use it, but I don't feel strongly enough about it if there's a lot of pushback.

elle:
Plus one for consistency across a project. But also, I'd be happy to have this as a company-wide convention with the knowledge that a team on a specific project can decide to take a different approach. 

But then again, if enough people object, I ok leaving it as is.

mjankowski:
I'm closing this because it looks like "lets just be consistent on each project" is a slightly stronger held opinion than "we must have a guideline for this", and because there doesn't appear to be incredible passion here, only mild preference.

PR #264, opened 2014-12-08
Prefer initializers over global variables in Ember

BlakeWilliams:
Prefer to use initializers and dependency injection when dealing with "global" data in Ember.

eg: Storing the current user.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -291,6 +291,8 @@ Ember
 * Don't use jQuery outside of views and components.
 * Prefer to use predefined `Ember.computed.*` functions when possible.
 * Use `href="#"` for links that have an action.
+* Prefer dependency injection through Ember initializers over globals on window
+  or namespaces.
 
 Angular
 -------


jferris:
> It's specific to Ember since you should be injecting them via initializers when possible.

It seems like the "Avoid global variables" part is a generic programming guideline. Maybe just the part about avoiding `window` is useful, since that doesn't *seem* global but it is?

BlakeWilliams:
The point here is to encourage more use of [initializers](http://emberjs.com/api/classes/Ember.Application.html#toc_initializers). I still see a lot of information put on `window` and `App` when Ember lets you inject it where you need it specifically.

BlakeWilliams:
Maybe 
> Prefer dependency injection through Ember initializers over `window` and namespaces?

jferris:
@BlakeWilliams your latest comment seems clearer to me. Want to push an update?

BlakeWilliams:
@jferris Yep, doing it now.

BlakeWilliams:
Any more thoughts before I merge this?

JoelQ:
:+1:

PR #269, opened 2014-12-10
Use 2 spaces for swift

Keithbsmiley:
I think we should use 2 spaces for Swift code. Thoughts?

### DIFF
--- a/style/samples/Swift.swift
+++ b/style/samples/Swift.swift
@@ -8,7 +8,7 @@ typealias CoolClosure = (foo: Int) -> Bool
 
 // Use aliased parameter names when function parameters are ambiguous
 func yTown(some: Int, withCallback callback: CoolClosure) -> Bool {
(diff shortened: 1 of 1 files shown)


jferris:
Anything else to add here?

dazmuda:
My 2cents: If the guides are a guideline for writing code like we write at thoughtbot, I like 2 spaces and think this should be in. However, I like to think of guides as a "best practices" type thing, which makes me think that since this is a preference it shouldn't be in. My gut says the latter.

gfontenot:
If more people like 2 spaces than 4, I'm OK with this. It seems like @jnutting and I are the only dissenters.

Keithbsmiley:
I think, that since this isn't unanimous, I should just close this. Since what will probably happen is people who don't like 2 spaces will keep using 4 and this PR won't mean anything.

PR #270, opened 2014-12-10
Remove outdated swift example

Keithbsmiley:
This API has since been audited. I think we should still have this in
the guide, does anyone know of an API we could use in place of this?


jferris:
@Keithbsmiley Anything holding this up?

PR #272, opened 2014-12-10
Use MARK in swift section comments

Keithbsmiley:
I think this is another thing we should do, it just happens to work in
this guide too. @thoughtbot/ios


jferris:
Is this good to merge?

>  just wondering if there's a way we can also say "but maybe consider whether this section should be extracted instead of commented"

Can you open a pull request for an additional guideline?

PR #273, opened 2014-12-10
Update Objective-C examples

Keithbsmiley:
This was just a quick pass to update the objc examples with what I feel like we actually follow.

I removed the `enumerateObjectsUsingBlock:` portion too. Partially because I feel like that's not actually what we use, and then there's this: http://www.objc.io/issue-7/collections.html Either way I don't feel like we follow a hard and fast rule about that so that portion now just references the spacing.

// @thoughtbot/ios

### DIFF
--- a/style/objective_c/sample.m
+++ b/style/objective_c/sample.m
@@ -19,9 +19,6 @@
 static NSString *const TBConstantName = @"Constant";
 static NSUInteger const TBNumberOfCardsInDeck = 52;
 
-// Prepend constants with 'k' when being used as keys
-static NSString *const kFirstName = @"FirstName";
-
 @implementation ClassName
 
 /*
@@ -31,7 +28,7 @@ static NSString *const kFirstName = @"FirstName";
 */
 #pragma mark - Initialization
 
-- (id)initWithCoder:(NSCoder *)aDecoder
+- (instancetype)initWithCoder:(NSCoder *)aDecoder
 {
     self = [super initWithCoder:aDecoder];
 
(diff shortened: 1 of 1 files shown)


jferris:
Is this good to merge?

style/samples/ObjectiveC.m
```diff
@@ -49,8 +46,8 @@ - (id)initWithCoder:(NSCoder *)aDecoder
 - (void)shuffleCards
 {
     // Objective-C literals are your friend
-    NSDictionary *themeColors = @{ kRedColor : [UIColor redColor], kBlueColor : [UIColor blueColor] };
-    NSArray *robots = @[ @"Ralph", @"Bender", @"The Iron Giant" ];
+    NSDictionary *themeColors = @{redColor: [UIColor redColor], blueColor: [UIColor blueColor]};
```
dazmuda: I prefer keeping `kRedColor` even though we don't declare those constants and it wouldn't compile. `kBlueColor` screams "constant" (and I think that guideline is in the guides) whereas `blueColor` screams "I did something crazy with `copyWithZone:` to use this object as a key."

gfontenot: I think it should be `BlueColor`, not `blueColor`, for the record. I didn't make that clear enough earlier.

dazmuda: Ah, then I agree with Gordon. Since k-prefixing constants isn't our jam anymore, `BlueColor` sounds great.

PR #277, opened 2014-12-19
Sass: Prefer mixins to @extends

creuter:
Extends can cause long, unwieldy selectors if not minded. Compiled file
size is negligible with Gzip. Extending doesn't work across media
queries, and can't gracefully accept arguments.

Further discussion at https://github.com/thoughtbot/bitters/pull/140


jferris:
@creuter seems like this is good to merge?

PR #278, opened 2014-12-22
A pull system means never assigning

mike-burns:
This introduces a new communications section for describing how best to
communicate like a thoughtbotter.

Up first: tickets. We use Trello right now, but the principles of how we
use it are the same as when we used Pivotal Tracker, JIRA, Trajectory,
GitHub Issues, and everything else. Therefore I've left it as just
"tickets."

We follow a pull system, which breaks down like this:

* For client projects, the next designer or developer takes the ticket
  with top priority.
* For discussion topics, such as happens on the Research board, people
  subscribe to tickets they are interested in.

Assigning people to a ticket defeats this and causes confusion.
Therefore, we don't do that.

### DIFF
--- a/protocol/README.md
+++ b/protocol/README.md
@@ -3,6 +3,7 @@ Protocol
 
 Guides for getting things done.
 
+* [Communication](/protocol/communication)
 * [Git](/protocol/git)
 * [Rails](/protocol/rails)
 * [iOS](/protocol/ios)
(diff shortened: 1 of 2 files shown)


mike-burns:
I've added a shorter, simpler rewrite and a longer, more detailed commit message.

jferris:
The latest version reads well to me.

Are there any objections to merging this?

gabebw:
I like it.

teoljungberg:
I concur

PR #279, opened 2014-12-23
Decide when to merge your branch

mike-burns:
The value of a code review is in the feedback from other developers.
Once the developer gets feedback, it's up to hir to decide whether the
branch is ready to merge.

Sign-off is handy for some repos which affect everyone (guides,
playbook, handbook, etc.). Others, such as the blog or code, can be more
lax. As such, remove this from the guides since it does not apply
everywhere.

### DIFF
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -39,9 +39,9 @@ Having Your Code Reviewed
   able to read individual updates based on their earlier feedback.
 * Seek to understand the reviewer's perspective.
 * Try to respond to every comment.
-* Wait to merge the branch until another person signs off.
 * Wait to merge the branch until Continuous Integration (TDDium, TravisCI, etc.)
   tells you the test suite is green in the branch.
+* Merge once you feel confident in the code and its impact on the project.
 
 Reviewing Code
 --------------


JoelQ:
Without completely disagreeing here, I like waiting for a sign off that lets me know that the reviewer has completed looking at the PR and any changes I may have made in response to previous feedback. I've had people ask me for a review in the past and then merge while I was half-way through it.

If I've made a lot of changes to a PR, I like to have someone else take a final look before I merge it. For these reasons, I personally like to wait for a :+1: before merging.

BlakeWilliams:
I agree with @JoelQ. Even if I feel confident in the code I like waiting for someone else to reinforce that opinion.

gabebw:
I agree with everything @JoelQ wrote. I don't really like "Fix these things and then good to merge" comments for similar reasons, because in the process of fixing those problems, I may introduce a new problem.

mike-burns:
The proposed change doesn't affect such a workflow, @gabebw @JoelQ .

drapergeek:
I agree with @mike-burns, I think that if people want to wait on a re-review that is up to them. There are times I find that necessary but a lot of times if the feedback was minor, waiting on a re-review is a waste of time. I think this definitely plays into the 'use your best judgement' idea.

:+1: from me

mike-burns:
OK, I'm going to merge this.

PR #284, opened 2015-01-07
Add guide for Sass color functions

tysongach:
Addition: Use a `%` for the amount/weight, when using Sass’s color functions: `darken($color, 20%)`, not `darken($color, 20)`

I also fixed some grammar and formatting issues.

@thoughtbot/design Thoughts?

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -1,39 +1,36 @@
-Sass
-====
+# Sass
 
 [Sample](sample.scss)
 
-Formatting
-----------
+## Formatting
 
-* Use the *Scss* syntax.
+* Use the SCSS syntax.
 * Use hyphens when naming mixins, extends, classes, functions & variables: `span-columns` not `span_columns` or `spanColumns`.
-* Use space between property and value: `width: 20px` not `width:20px`.
-* Use a blank line above selector that has styles.
-* Prefer hex color codes `#000`.
+* Use one space between property and value: `width: 20px` not `width:20px`.
+* Use a blank line above a selector that has styles.
+* Prefer hex color codes `#fff`.
 * Use `//` for comment blocks not `/* */`.
-* Use a space between selector and `{`.
+* Use one space between selector and `{`.
 * Use double quotation marks.
 * Use only lowercase, including colors.
 * Don't add a unit specification after `0` values, unless required by a mixin.
 * Use a leading zero in decimal numbers: `0.5` not `.5`
 * Use space around operands: `$variable * 1.5`, not `$variable*1.5`
 * Avoid in-line operations in shorthand declarations (Ex. `padding: $variable * 1.5 variable * 2`)
-* Use parentheses around individual operations in shorthand declarations: `padding: ($variable * 1.5) ($variable * 2)`
+* Use parentheses around individual operations in shorthand declarations: `padding: ($variable * 1.5) ($variable * 2);`
 * Use double colons for pseudo-elements
+* Use a `%` unit for the amount/weight when using Sass's color functions: `darken($color, 20%)`, not `darken($color, 20)`
 
-Order
------
+## Order
 
 * Use alphabetical order for declarations.
-* Place @extends and @includes at the top of your declaration list.
+* Place `@extends` and `@includes` at the top of your declaration list.
 * Place media queries directly after the declaration list.
 * Place concatenated selectors second.
 * Place pseudo states and elements third.
 * Place nested selectors last.
 
-Selectors
----------
+## Selectors
 
 * Don't use ID's for style.
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
@@ -48,12 +45,11 @@ Selectors
 * Avoid using comma delimited selectors.
 * Avoid nesting within a media query.
 
-Organization
-------------
+## Organization
 
-* Use Bourbon for a Sass Library.
+* Use Bourbon for a Sass library.
 * Use Neat for a grid framework.
-* Use Bitters / Base folder for style on HTML tags, global variables, global extends and global mixins.
-* Use Normalize as a browser reset.
+* Use Bitters/`base` directory for styling element selectors, global variables, global extends and global mixins.
+* Use [Normalize](https://github.com/necolas/normalize.css) for browser rendering consistency, rather than a reset.
 * Use HTML structure for ordering of selectors. Don't just put styles at the bottom of the Sass file.
 * Avoid having files longer than 100 lines.


creuter:
:+1: I think it's more readable, and better explains what's happening.

style/README.md
```diff
@@ -82,25 +82,26 @@ Sass
...
+* Use one space between property and value: `width: 20px` not `width:20px`.
+* Use a blank line above a selector that has styles.
+* Prefer hex color codes `#fff`.
 * Use `//` for comment blocks not `/* */`.
-* Use a space between selector and `{`.
+* Use one space between selector and `{`.
 * Use double quotation marks.
 * Use only lowercase, including colors.
-* Don't add a unit specification after `0` values, unless required by a mixin.
+* Don’t add a unit specification after `0` values, unless required by a mixin.
```
pbrisbin: Is this smart quote intentional?

tysongach: Yes, it’s intentional. It’s proper, typographically.

style/README.md
```diff
@@ -82,25 +82,26 @@ Sass
...
 * Use a leading zero in decimal numbers: `0.5` not `.5`
 * Use space around operands: `$variable * 1.5`, not `$variable*1.5`
 * Avoid in-line operations in shorthand declarations (Ex. `padding: $variable * 1.5 variable * 2`)
 * Use parentheses around individual operations in shorthand declarations: `padding: ($variable * 1.5) ($variable * 2)`
+* Use a `%` for the amount/weight, when using Sass’s color functions: `darken($color, 20%)`, not `darken($color, 20)`
```
kylefiedler: :-1: The `%` is not needed and doesn't add any value to me.

style/README.md
gabebw: I don't think we should use it in this repo. Here's why:

* It's the only UTF-8 character in the whole repo,
* difficult to type, and
* straight quotes are rendered into typographic quotes when viewing on GitHub anyway

pbrisbin: I agree and would extend the rule to any markdown-to-html scenario.

tysongach: No worries, it’s a small thing.

This is the second time I’ve heard that GitHub is supposedly rendering proper smart quotes, but that doesn’t seem to be the case.

gabebw: Even if GitHub isn't rendering smart quotes, I don't think it's worth changing this in the repo.

kaishin:
Looking good.

Magnus-G:
I agree with @kylefiedler here. I think the color adjustments are clear as they are and I'd love to avoid those awkward key presses.

brendastorer:
I also agree with @kylefiedler. I don't think the "%" is necessary. But I think it's a good idea to have a guideline to specify it one way or the other for consistency.

creuter:
It isn't necessary, but neither are the leading 0 in decimals, or "button" instead of "btn" classes. We generally prefer explicitness and verbosity for clarity, I don't see why this should be an exception.

cc @joshuaogle

joshuaogle:
I would argue that it is necessary. All of the examples in the [Sass docs](http://sass-lang.com/documentation/Sass/Script/Functions.html) use `%`. It still compiles without it, but that's just a circumstance of this mixin only accepting percentages.

kaishin:
I haven't read @kylefiedler's comment before since it was collapsed.
There is no way I will write `darken($color, 20)`.  Twenty what? Amperes?

kylefiedler:
I’m guessing all of you know the value of this unit:

`line-height: 1;`

Or don’t care about the unit here:

`margin: 0;`

And can tell me exactly what this is without any indication what it is:

`0, 0, 0, 0.5`

If you value verbosity for clarity wouldn’t this be your preference:

`adjust-color(#fff, lightness: 20%)`

`lighten` and `darken` are shortcuts for that function. They are both less clear that they are adjusting the lightness spectrum in hsl. Especially `darken` which is giving `lightness` a negative value.

`%` already is the symbol for silent extends and for a unit of measurement. If possible I don’t want to justify another meaning because I believe makes Sass harder to glance through.

And yes I’m lazy and I don’t want to type something that isn’t necessary.

tysongach:
Outside of Sass, 20 ≠ 20%, but 0.2 = 20%.

Using Sass’s color functions, this:

```scss
color: darken(red, 0.2); // returns #fe0000
```

…is very different from this:

```scss
color: darken(red, 20%); // returns #990000
```

For me, the addition of the `%` makes it very clear what is actually going to happen to the color.

whmii:
To me, requiring a `%` on these units is like saying that if you had a declaration like `margin: 0;` then you were required to write `margin: 0em;` which doesn't make any sense to me.

Had I written this function I would have preferred it to be calculated based of a 0 - 1 integer  like `darken(red, 0.2)`.

joshuaogle:
@kylefiedler Line-height's unitless number is a little different because it is a multiplier. `1` means `1x`, `1.75` means `1.75x` and so on. If you used `1.75em` it would give you the same thing because ems work the same way, but that's why the unit is missing there. I can't think of another case in CSS where they use multipliers, so I wish they hadn't since it confuses some people. `0` is also a special case because `0%` = `0em` = `0px`.

If there is this much dissent, then we could avoid having a guideline. I would want to side with the Sass docs though. Units are important.

> Twenty what? Amperes?

Ha. Exactly

kaishin:
To counter the counter-arguments:
- `0` is a zero, so the unit does not matter. `0%`, `0em`, `0px` are all going to be equal to nothing,
- `line-height` is a relative value coming from the CSS spec and has thus nothing to do with Sass types.

Now, some additional arguments:
- Adding `%` makes it clear that it's a percentage. It'll guarantee that no one will write `200%` without knowing what it entails. `color: darken(red, 255)` looks logical to me, whereas `color: darken(red, 255%)` doesn't.
- There has been a counter-intuitive trend in Sass where everything is becoming more and more ambiguous: unknown length argument lists, weak typing, versatile function signatures, etc. Other languages have been there already and are now heading towards stricter typing and less versatile syntaxes. This results in less bugs and better readability.

kylefiedler:
The `%` here is a relative number as well. It's adding `+` or `-` `$value` relative to the inputed color's lightness. HSL lightness only takes 100% so if you put in `255%` it would give you white no matter the color. But in most cases we're not on that scale because we're starting off in-between 0 and 100. So doing `darken(red, 100%)` is just as meaningless to me as putting in `darken(red, 255%)` since you are in effect giving the lightness a value `-150%`. So in most cases we're not dealing with a 0 – 100 scale but a custom scale in both directions for each individual color. In this case with `red` it just so happens it's `50` in both directions, so putting any number higher than `50` is useless. By adding the unit you aren't gaining anything except it's relationship to lightness which, I'll give you, is a `%`.

No one took my suggestion of `adjust-color(#fff, lightness: 20%)` and not use the `lighten` `darken` seriously. Since those are more ambiguous and you all are vying for clarity I don't understand why that's not what you are arguing for. It's the most explicit and descriptive of what each of these functions are doing. 

> If there is this much dissent, then we could avoid having a guideline.

I'm fine with this.

brendastorer:
What if the guideline is just to keep its use or "not use" consistent within a project?

kaishin:
@kylefiedler We are not arguing its necessity from a result point of view; we're arguing it from making Sass functions and mixins, in general, clear about what type of `Number` value they expect: unit, unitless, percentage, etc. This is the offical documentation:

> Makes a color darker. Takes a color and a number between 0% and 100%, and returns a color with the lightness decreased by that amount.

Just because Sass doesn't have strong type checks, yet, it doesn't mean we should strip all unnecessary things. What if Sass 4.0 introduces a `<Percentage>` type and [this line](https://github.com/sass/sass/blob/bd1270c8c6b46219d879bf79f5b72874aeddefcd/lib/sass/script/functions.rb#L2624) is replaced with `assert_type amount, :Percentage, :amount`?

By removing the unit, we are exploiting a weakness in the current Sass implementation, not a feature.
 
> Since those are more ambiguous and you all are vying for clarity I don't understand why that's not what you are arguing for. 

There is a fine balance between clarity and verbosity. Your suggestion is way too verbose but it is perfectly okay if you ask me. I'd rather see than the other extreme.

I don't want to write `darken(red, 20)` regardless of the circumstances.

kaishin:
Also `darken(red, 255)` and `darken(red, 255%)` are both illegal. So clarity of intent does matter here.

kylefiedler:
So we're clear I am not arguing for us to make everyone not use `%` I'm arguing that we shouldn't have this specified. 

> we're arguing it from the point of view of making Sass functions and mixins, in general, clear about what type of Number value they expect: unit, unitless, percentage, etc

I know, I'm arguing that in this case they shouldn't take a value. That it should be unitless. It might just be my mental model and how I read it the code.  `darken $color by 20` or `$color lightness value - 20`. When I see `darken $color by 20%` I think this `$color lightness value * - 20%".

> What if Sass 4.0 introduces a <Percentage> type and this line is replaced with assert_type amount, :Percentage, :amount?

Couldn't they also come to the conclusion that this should be unitless?

Either way this has been fun discussion but I'm throwing in my towel. I've spent more time arguing about the `%` than I would typing it out all the time. Decide what you will.

kaishin:
If typing less was our primary concern, we should all be writing `.sass` ;)

tysongach:
Wow.

When I put up this PR, I thought it would be a shoo-in. I had no idea people were so particular about this! 

The Sass docs use percentages and the [documentation for the functions](http://sass-lang.com/documentation/Sass/Script/Functions.html#darken-instance_method) themselves say “Takes a color and a number between 0% and 100%…”. I’m actually *surprised* you can pass a value without the percentage unit. If you write `darken(red, 255%)`, the exact error you receive is `Amount 255% must be between 0% and 100% for 'darken'`.

That being said, I’m also fine with leaving this unspecified. There are better things to do.

Is everyone good with the rest of my minor grammatical fixes? If so, I’ll extract out of this PR and merge them in.

kaishin:
I want to push for the `%` guideline. Leaving it out is clearly a bad practice and should not be tolerated. I'd say this is more important than leading 0 in decimals and other guidelines we have.

Magnus-G:
FYI this is implemented in Refills now.

style/README.md
jferris: Want to open a pull request for Markdown and smart quotes? (Or something similar)

whmii:
@tysongach LGTM

tysongach:
We have more “Yes” than “No”, right now…

jferris:
Are people still divided on this? Should I create a poll?

jasonramirez:
Poll please.

On Thursday, January 22, 2015, Joe Ferris <notifications@github.com> wrote:

> Are people still divided on this? Should I create a poll?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/thoughtbot/guides/pull/284#issuecomment-71106736>.
>


-- 

Jason Ramirez

Designer | thoughtbot, Inc.

347-703-8617

www.thoughtbot.com

twitter.com/thoughtbot

tysongach:
@thoughtbot/design Here’s a quick poll, please fill it out: http://goo.gl/forms/IkR1yln0EL

tysongach:
It’s been 24 hours. The results:

| Timestamp          | Username                 | Should our Sass guides include adding a percentage unit to color functions? |
|--------------------|--------------------------|-----------------------------------------------------------------------------|
| 1/29/2015 11:36:39 | christian@thoughtbot.com | Yes                                                                         |
| 1/29/2015 11:37:06 | brenda@thoughtbot.com    | No                                                                          |
| 1/29/2015 11:42:37 | tyson@thoughtbot.com     | Yes                                                                         |
| 1/29/2015 11:49:29 | joshua@thoughtbot.com    | Yes                                                                         |
| 1/29/2015 12:31:45 | jason@thoughtbot.com     | Yes                                                                         |
| 1/29/2015 12:56:33 | steven@thoughtbot.com    | Yes                                                                         |
| 1/29/2015 13:23:40 | don@thoughtbot.com       | Yes                                                                         |

I would consider this a success and mergable.

tysongach:
I’m going to merge this soon. Last call.

PR #285, opened 2015-01-07
Remove guideline for recommended css file structure.

brendastorer:
I don't think this is considered a best practice anymore, and this isn't a guideline I follow. File structure for CSS is moving towards styling modules/components that can be reused anywhere vs. per page or view. This can cut down on repeated code for similar styles across views, and also helps maintain consistency in design across a site.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -55,5 +55,4 @@ Organization
 * Use Bitters / Base folder for style on HTML tags, global variables, global extends and global mixins.
 * Use Normalize as a browser reset.
 * Use HTML structure for ordering of selectors. Don't just put styles at the bottom of the Sass file.
-* Prefer the same file structure that is found in app/views.
 * Avoid having files longer than 100 lines.


carolanncan:
Agreed. :+1:

kylefiedler:
@brendastorer I agree but I want you to replace it another (or multiple) guideline on how we should be structuring our Sass files.

kylefiedler:
This might be a good starting point: http://sass-guidelin.es/#architecture

brendastorer:
Good idea @kylefiedler Thanks!

jferris:
If it's not possible to swap in a new recommendation quickly, it seems like it would be worth merging this as-is. It seems to me that it's worse to provide bad advice than to provide no advice at all.

jferris:
@kylefiedler @brendastorer any updates here?

brendastorer:
I would like to merge it in as is, because we definitely don't need this guideline, and then take some time to make a researched and collaborative PR with an update to how we should organize the file structure.

joshuaogle:
:+1: For merging. I would love to see a card on the Research board for suggested organization if someone has a system in mind.

brendastorer:
Thanks all. @joshuaogle great suggestions. Card created and can be found here: https://trello.com/c/0YAPIBso/448-best-practices-for-css-sass-file-structure Merging in as is.

PR #287, opened 2015-01-08
Update guide to reflect background-color style

ryancoughlin:
I was talking to @creuter yesterday about this. When only specifying a background color use `background-color` vs `background` - seems more "proper." `background` feels more appropriate when using background shorthand.

I wasn't sure where else to make the note. I added it as a comment (that may be a bad thing).
Thoughts on this?

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -9,6 +9,7 @@
 * Use one space between property and value: `width: 20px` not `width:20px`.
 * Use a blank line above a selector that has styles.
 * Prefer hex color codes `#fff`.
+* Avoid using shorthand properties for only one value: `background-color: #ff0000;`, not `background: #ff0000;`
 * Use `//` for comment blocks not `/* */`.
 * Use one space between selector and `{`.
 * Use double quotation marks.


tysongach:
I agree with this. :+1: 

Maybe alongside our Sass guide should be a CSS guide, for little tidbits about the vanilla CSS spec? I don’t think we should comment on the code here; this is more of an example, rather than the place where the guides are noted.

ryancoughlin:
@tysongach Ok, thanks. I agree it is more CSS related than SASS. The comment I wasn't 100% on since its not the best practice to use that in code (with this being out style guide). I didnt know if it would be clear enough by having:

```
background-color: $thoughtbot-red
```
Aka use that not `background`

kylefiedler:
I agree with this too. I think you need to specify in https://github.com/thoughtbot/guides/tree/master/style#sass with a line about what this means while keeping the example. Basically your github comment above. Also I think this should be for `font` too.

kylefiedler:
I don't mind that it is in Sass for me the two are one in the same now and it makes it easier for me to look at all the rules when they are grouped under on language.

creuter:
^ Ditto, keep it with the Sass style guide.

kaishin:
:+1: 

I stopped using the shorthand version altogether because I can never remember which comes first, position, size or repeat.

ryancoughlin:
@kaishin Same here. With background I usually will use:

```
background: {
    color: $thoughtbot-red;
    position: ...;
    image: ...;
    size: ...;
}
```

kylefiedler:
I don't like the look of the Sass shorthand because to me it looks like a selector. I just use regular CSS.

jferris:
Are there any objections to merging this as-is?

> I don't like the look of the Sass shorthand because to me it looks like a selector. I just use regular CSS

Want to open a pull request for a separate guideline?

tysongach:
The addition itself is good, but it needs to be moved, per the comments above.

Should this go in “[Best Practices](https://github.com/thoughtbot/guides/tree/master/best-practices#sass)”? What’s the difference?

jferris:
> Should this go in “Best Practices”? What’s the difference?

The "Style" guide is reserved for guidelines that only affect aesthetics. For example, `a{color:white;}` and `a { color: white; }` are identical and will be interpreted exactly the same by any browser.

The "Best Practices" guide is for guidelines that stray beyond aesthetics. For example, a guideline like "Prefer flexbox to floats" has a meaningful impact on how you write the rest of the CSS and on browser support.

Some examples are obvious, but many are in an unfortunate gray area. I tend to err on the side of something being a "Best Practice" and reserving the "Style" section for things I feel sure about.

tysongach:
Nice. Thanks for the clarification, @jferris! That makes sense. Maybe a little one-liner describing the purpose could go under the heading of each of these sections?

jferris:
> Maybe a little one-liner describing the purpose could go under the heading of each of these sections?

Yeah, I think that could be nice. Want to open a pull request?

jferris:
It looks like the current diff only updates the example, and does not have a corresponding guideline in the Style section. Can we move the comment into the guide?

jferris:
ping @ryancoughlin

ryancoughlin:
@jferris Totally missed this thread.

Will this be added as a note in "Best Practices" under "Sass" - want to clarify?

jferris:
@ryancoughlin I'm guessing it should go here: https://github.com/thoughtbot/guides/tree/master/style/sass

ryancoughlin:
I thought this would be enough instead of adding: "when only specifying a background-color.."

Thoughts?

jferris:
ping @thoughtbot/design

joshuaogle:
:+1: to rewording this to abstract it a bit and using the more specific properties when you don't need the shorthands for `background` and `font`. It sounds like would make the same case to use `background-image` over `background` if you are just specifying an image, etc.

How about:
"Avoid using shorthand properties for only one value: `background-color: #ff0000;`, not `background: #ff0000;`"

smharley:
i like @joshuaogle 's wording ^

ryancoughlin:
@joshuaogle @smharley Agreed. Makes it less specific. Will update.

ryancoughlin:
@joshuaogle @smharley Updated!

joshuaogle:
:rocket:

PR #288, opened 2015-01-15
Do not put timestamps at the top of a migration

gabebw:
`rails generate` puts timestamps at the bottom of a migration by default, 
meaning we have to change every migration even if the migration is otherwise 
generated perfectly.

Original discussion: https://github.com/thoughtbot/guides/pull/159

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -30,9 +30,7 @@ Migrations
 
 * Set an empty string as the default constraint for non-required string and text
   fields. [Example][default example].
-* List timestamps first when creating a new table. [Example][timestamps example].
 
-[timestamps example]: migration.rb
 [default example]: migration.rb#L6
 
 Routes


style/rails/README.md
```diff
@@ -28,9 +28,7 @@ Migrations
 
 * Set an empty string as the default constraint for non-required string and text
   fields. [Example][default example].
-* List timestamps first when creating a new table. [Example][timestamps example].
```
gabebw: I removed this line entirely (instead of replacing it with the opposite guideline) because, as the commit message mentions, `rails generate` puts timestamps at the bottom by default.

jessieay:
Disagree! I still like timestamps at top - looks so much better!

gabebw:
Excellent, I knew this would lead to spirited discussion!

> looks so much better!

Why does it look better? Does it add a readability benefit? Why go against the Rails default?

jessieay:
Timestamps are a given for most database tables. As such, they are not very interesting or telling about the nature of a database table.

Because I don't much are about these fields, I prefer to know where they are so I can ignore them. When they are listed in the middle, I am forced to consider them part of the table I am looking at the same as any other field.

There is also an aesthetic thing here - it looks more consistent to have them at the top of each table. The screenshot I included in https://github.com/thoughtbot/guides/pull/159 shows this

gylaz:
I'm confused by the title of this PR and its content.

`timestamps` seem like a given, and don't add much to the contents of the table. I do prefer bottom, since it's not likely what I want to see. I'm not sure that we even need a guildeline on this, as long as there's consistency in the project.

gabebw:
> When they are listed in the middle, I am forced to consider them part of the table I am looking at the same as any other field.

Whoa, who said put them in the middle? I think they should be at the bottom of the `create_table`, regardless of alphabetization.

jyurek:
I'm for this change. This REALLY does not need to be in the style guide.

jessieay:
@gylaz @gabebw 

the reason for including them at the top is that putting them at the bottom is actually putting them at the middle, since most tables will receive migrations beyond the initial `create` which will put new fields below `created_at` and `updated_at`

BlakeWilliams:
I'm a fan of putting timestamps at the top so you don't have them in the middle of your table after you add a few migrations. I do always forget to move them though.

gylaz:
> putting them at the bottom is actually putting them at the middle

I see now that we are talking about the end-result of `schema.rb` and not just the migration file. I agree that it seems better to put them at the top, because of additional migrations will add columns after the timestamps.

gabebw:
> I see now that we are talking about the end-result of schema.rb 

Actually, `schema.rb` is always alphabetized by column name, no matter the order in the migration file. Nothing we write in the migration file will affect column order in `schema.rb`. I opened this PR just for migration style.

drapergeek:
>Actually, schema.rb is always alphabetized by column name, no matter the order in the migration file. Nothing we write in the migration file will affect column order in schema.rb. I opened this PR just for migration style.

Can you provide an example of this? I don't see this happening in any of my projects.

derekprior:
It seems like a silly thing to have a guide for, IMO. I'm fine with removing them and having them included in your table migration wherever you see fit. It's not like you read migrations every day.

gabebw:
Ugh, I'm ill and misread my `schema.rb`. It is not alphabetized. I agree with Derek, though:

>  It's not like you read migrations every day.

drapergeek:
:+1: I don't see enough benefit here in overriding the rails defaults.

jferris:
This comes up most often when looking at `ActiveRecord` instances in console, particularly in production. If you add a column to the end of the table in a later migration, the timestamps will show up in the middle.

Example:

```
> User.last
=> #<User id: 1234, email: "user@example.com", encrypted_password: nil, salt:
nil, confirmation_token: nil, remember_token: "secret", email_confirmed: true,
created_at: "2015-01-15 19:15:50", updated_at: "2015-01-15 19:16:14", reference:
nil, admin: false, stripe_customer_id: "cus_secret", github_username: "wup",
auth_provider: "github", auth_uid: 12345, organization: nil, address1: nil,
address2: nil, city: nil, state: nil, zip_code: nil, country: nil, name: "Billy
Persona", bio: nil, mentor_id: nil, team_id: nil>
```

I think these are an unreadable blob either way, though.

teoljungberg:
I'll :+1: to remove this guideline, I don't see the real value of it

gabebw:
If you want to put timestamps at the top of your migration, that's fine. But I don't think we should have a guideline *preferring* timestamps at the top.

mike-burns:
I'm in favor of the diff stats of this PR.

jferris:
Are people still opposed to removing this guideline?

BlakeWilliams:
I still prefer the timestamps up top, but I'm not heavily opposed to removing it.

gylaz:
I'm also, now, a fan of timestamps at the top, since it will unsure consistency for when additional fields are added to the table.

gabebw:
I don't understand why it matters at all. How often do we really look at the schema, or at every single attribute on an AR model?

derekprior:
> I'm also, now, a fan of timestamps at the top, since it will unsure consistency for when additional fields are added to the table.

Honestly curious as to why the order matters at all and why it would be timestamps that matter more than any other field in this manner.

jferris:
I see the output for `ActiveRecord::Base#inspect` fairly often in test failures and from Rails console.

gabebw:
> I see the output for ActiveRecord::Base#inspect fairly often in test failures and from Rails console.


Whenever I see that output, I give up on reading it and write something like `puts user.name` so I get actual useful input. Like Derek said - why should timestamps be first, too? Do they help in debugging test failures?

jessieay:
> Why go against the Rails default?

I do not add fields to my rails migrations with the generator, so I always write my migrations from scratch. This means that I don't *change* the Rails default, I just write my migrations in the way I think they are best written.

It is helpful to have a standard way of writing migrations so that I do not forget anything. 

My standard: writing timestamps first, them all of the database fields in order (by attribute name). 

I think this is a useful format to standardize on *and* has the benefit of making the schema look more uniform.

gabebw:
Just so we don't stray too far from the original point: I believe that we should have no guideline. If there is a project that puts timestamps at the top, that's fine. If there's a project that puts them at the bottom: also fine. And not doing anything about timestamp position is also fine.

Some people use the generators, and some people don't, which definitely influences which way their opinion goes.

But I believe that we should not have an official guideline about this at all, which allows projects to pick what they do about this issue.

pbrisbin:
> I believe that we should have no guideline. If there is a project that puts timestamps at the top, that's fine. If there's a project that puts them at the bottom: also fine. And not doing anything about timestamp position is also fine

Yup. This.

croaky:
I prefer the timestamps at the top so it's easier to see when they are missing entirely or missing `null: false`. It's an easy guideline to follow and generates consistency in the file, which a large part of our guidelines are about. Since timestamps added to the bottom of a migration often end up in the middle of a column as further columns are added in future migrations, "put timestamps at the bottom of a migration" doesn't add consistency, but at the top does.

I believe our existing guideline is worth continuing to follow and therefore vote against this particular PR.

croaky:
> I don't see the real value of it

The problem I see "top of migration timestamps" solving is preventing confusing bugs down the line because timestamps were forgotten or `NULL` constraints on the timestamps were forgotten. I have seen that happen in projects before. Visually scanning a `db/schema.rb` file which follows this guideline helps avoid those problems or find them when they crop up because you know where to look for the timestamps or notice them when they are missing.

teoljungberg:
> Just so we don't stray too far from the original point: I believe that we should have no guideline. If there is a project that puts timestamps at the top, that's fine. If there's a project that puts them at the bottom: also fine. And not doing anything about timestamp position is also fine.

> Some people use the generators, and some people don't, which definitely influences which way their opinion goes.

> But I believe that we should not have an official guideline about this at all, which allows projects to pick what they do about this issue.

I couldn't have said it better myself. I totally agree.

mcmire:
Just to offer my two cents, I am "one of those people" who care about the order of columns in a table. Although I'm not so anal about it as I used to be, I find that I often place important columns in key areas (for me, it's foreign keys first and timestamps last). Scanning through the schema to locate a column comes up often enough (especially early on in a project) that making any improvements to how things are organized tends to have a noticeable impact.

I think Dan's point makes sense -- if you place important columns last, as soon as you add another column you'll screw up the order -- but if you put them first, they'll always stay there. I would be okay if this gets removed, because I can see how others wouldn't feel so strongly about it, but I am down to try this out, at least.

JoelQ:
> Just so we don't stray too far from the original point: I believe that we should have no guideline. If there is a project that puts timestamps at the top, that's fine. If there's a project that puts them at the bottom: also fine. And not doing anything about timestamp position is also fine.
> 
> Some people use the generators, and some people don't, which definitely influences which way their opinion goes.
> 
> But I believe that we should not have an official guideline about this at all, which allows projects to pick what they do about this issue.

This ^^ :+1:

jferris:
I created a poll to address this important question: https://docs.google.com/a/thoughtbot.com/forms/d/1qjcIIy_r4qRid25bJ-Jq6uH6KXXAD4HBVOF7g7iaHU4/formResponse

style/rails/migration.rb
```diff
@@ -1,9 +1,10 @@
 class CreateClearanceUsers < ActiveRecord::Migration
   def change
     create_table :users  do |t|
-      t.timestamps null: false
       t.string :email, null: false
       t.string :name, null: false, default: ''
+
+      t.timestamps null: false
```
jferris: Since we removed the guideline, should we remove this entirely, so we don't imply that timestamps should be at the bottom?

jferris:
The poll results are fairly strongly in favor of removing the guideline.

style/rails/migration.rb
croaky: If it's removed, it can look like we advocate no timestamps, which is more problematic than its position.

How about remove the guideline but leave the timestamps in the example at the top?

nickrivadeneira:
:+1: for removing the guideline.

style/rails/migration.rb
jferris: I think that would be fine, too. So, we'd leave the example as it was before this pull request, and remove the guideline entirely.

gabebw: Can do.

gabebw:
I'm merging.

PR #290, opened 2015-01-22
Added IE8 to list of browsers to not support

galenandrew:
Considering [Microsofts announcement](http://blogs.windows.com/bloggingwindows/2015/01/21/the-next-generation-of-windows-windows-10/) for a new [web experience](http://gizmodo.com/microsofts-spartan-browser-is-here-to-save-you-from-int-1680879010) in Windows 10, IE8 is becoming even more deprecated.

Included IE8 to the list of browsers not to support.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -210,7 +210,7 @@ Browsers
 --------
 
 * Don't support clients without Javascript.
-* Don't support IE6 or IE7.
+* Don't support IE6, IE7, or IE8.
 
 Objective-C
 -----------


jferris:
@thoughtbot/design any thoughts here?

According to the [Playbook](http://playbook.thoughtbot.com/), we already don't support IE8:

> We support Internet Explorer 9.0+ and the latest versions of Firefox, Chrome, and Safari. We do not support Internet Explorer 6, 7, or 8.

smharley:
:+1: I think it's a great idea

joshuaogle:
:+1:

kaishin:
:100:

mcmire:
Go for it.

best-practices/README.md
```diff
@@ -210,7 +210,7 @@ Browsers
 --------
 
 * Don't support clients without Javascript.
-* Don't support IE6 or IE7.
+* Don't support IE6, IE7, or IE8.
```
JoelQ: Instead of listing all versions we don't support, how about something like:

> Don't support versions of Internet Explorer before IE9

gabebw: I like that idea.

whmii:
:+1:

jferris:
Looks like we're also dropping support for IE9: https://github.com/thoughtbot/playbook/pull/147

jferris:
Thanks for the pull request. I revised as suggested in comments and merged as 17fedd9143b6a06a62b825bc25ddeee5a68188e8.

galenandrew:
:+1: Thanks everyone!

PR #291, opened 2015-01-23
Support clients without JavaScript.

calleerlandsson:
Since it's a trade-off, not supporting clients without JavaScript capabilities should be a concious decision instead of the default.

Developing with [progressive enhancement](http://en.wikipedia.org/wiki/Progressive_enhancement) in mind makes it easier to build accessible web applications with semantic markup that can be used by more people in more web browsers. Not depending on JavaScript for core functionality often makes testing easier and progressively enhancing an application by applying new layers on top of existing ones is in line with agile practices.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -212,7 +212,6 @@ Sass
 Browsers
 --------
 
-* Don't support clients without Javascript.
 * Don't support IE6 or IE7.
 
 Objective-C


mike-burns:
I have nothing more to add besides: yes, definitely.

teoljungberg:
I agree, applications should still be usable without JavaScript. They might not be as feature-rich, but they are still usable

(This also goes without saying, if we build a JS application in ember or angular we won't support clients without Javascript in that case)

kaishin:
Agreed. This shouldn't be in the guides because it's case by case thing.

BlakeWilliams:
Agreed. Unless it's a special case or we're using a front-end framework we should support clients without JavaScript.

mcmire:
If it's not in guides, then, should it be in best practices?

joshuaogle:
:+1: For having it in best practices. It is a good idea, even if it doesn't work for every project.

mike-burns:
This isn't even a best practice?

BlakeWilliams:
I agree with @mike-burns. It's such a case by case issue that it doesn't seem like it's a best practice.

drapergeek:
:+1: Kill it.

calleerlandsson:
I agree that the reverse is more in line with what I think are best practices but I'd prefer no guideline. I see little point in single sentence guidelines unless they are strict rules that serve a specific purpose.

gabebw:
I vote no guideline.

jferris:
If we can get general agreement on removing the guideline, I think it's worth merging this as-is rather than trying to replace it with a new one immediately.

If this gets merged, it will make it easier to have a separate discussion a new pull request if somebody wants to add a new guideline.

mjankowski:
This is a weird guideline to have at all in a "Browsers" section, without any more context on what the guideline is for.

If you're building a Web Application, you don't need to worry about javascript support because you won't need to use JS, but if you're building a Single Page Application, of course you need to have JS support because that's how it works.

I'd vote to remove this as it's currently written, and call it a day -- or add something back with more context in an appropriate spot.

mcmire:
I'm fine with that.

Addvilz:
I completely agree that supporting (or not) users without JavaScript in their browsers must be a conscious decision. This must be a decision made by developers based on their applications purpose, clientele and many other factors.

Good read: http://www.punkchip.com/why-support-javascript-disabled/

jferris:
Seems like this is ready to merge.

PR #292, opened 2015-01-23
[Swift] Add guidelines for computed properties

kaishin:
@thoughtbot/ios Thoughts?


gfontenot:
:+1: Yes. I agree completely.

jferris:
@kaishin this seems like we're good to merge, yeah?

PR #294, opened 2015-01-26
Update Hound's language support

croaky:
We now support Ruby, JavaScript, CoffeeScript, and SCSS.
Rather than needing to update this each time we add languages,
don't be specific.


gabebw:
:+1:

JoelQ:
:+1:

PR #295, opened 2015-01-26
Fix broken link for RSpec subject style

croaky:


best-practices/README.md
```diff
@@ -109,7 +109,7 @@ Testing
 * Avoid `any_instance` in rspec-mocks and mocha. Prefer [dependency injection].
 * Avoid `its`, `let`, `let!`, `specify`, and `before` in RSpec.
 * Avoid using `subject` explicitly *inside of an* RSpec `it` block.
-  [Example][subject-example].
```
gabebw: It looks like this might work if you change the link to "explicit subject example" instead of "subject-example"? Or, keep this the same and change that link's URL on line 128?

croaky: Good catch. That "explicit subject example" also references an out of date file. Updating it but using that style.

PR #296, opened 2015-01-27
Replace mail preview statement for Rails

rob-murray:
Nice project; handy list of best practices :+1: 

This PR just updates the Rails mail preview point to say that ActionMailer Preview is available in Rails 4.1.0 or later now and you can still use the MailView gem if on an older version of Rails.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -173,12 +173,13 @@ Email
 
 * Use [SendGrid] or [Amazon SES] to deliver email in staging and production
   environments.
-* Use a tool like [MailView] to look at each created or updated mailer view
-  before merging.
+* Use a tool like [ActionMailer Preview] to look at each created or updated mailer view
+  before merging. Use [MailView] gem unless using Rails version 4.1.0 or later.
 
 [Amazon SES]: http://robots.thoughtbot.com/post/3105121049/delivering-email-with-amazon-ses-in-a-rails-3-app
 [SendGrid]: https://devcenter.heroku.com/articles/sendgrid
 [MailView]: https://github.com/37signals/mail_view
+[ActionMailer Preview]: http://api.rubyonrails.org/v4.1.0/classes/ActionMailer/Base.html#class-ActionMailer::Base-label-Previewing+emails
 
 JavaScript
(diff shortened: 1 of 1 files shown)


mcmire:
Anyone use ActionMailer Preview? Is it good/bad?

jessieay:
@mcmire it's good - was using it before when it was the mail_preview gem; now that it's part of Rails, I tend to use it whenever I am not using mandrill templates :)

rob-murray:
Same here - when you're using Rails to build mail templates its a great, quick way to preview the mail without having to send an email after each change.

joshuaclayton:
@jsteiner @halogenandtoast anything to add here re: ActionMailer::Preview?

jsteiner:
Nope. Looks good to me!

best-practices/README.md
```diff
@@ -173,12 +173,13 @@ Email
...
 * Use [SendGrid] or [Amazon SES] to deliver email in staging and production
   environments.
-* Use a tool like [MailView] to look at each created or updated mailer view
-  before merging.
+* Use a tool like [ActionMailer Preview] to look at each created or updated mailer view
+  before merging. Use [MailView] gem unless using Rails version 4.1.0 or later.
 
 [Amazon SES]: http://robots.thoughtbot.com/post/3105121049/delivering-email-with-amazon-ses-in-a-rails-3-app
 [SendGrid]: https://devcenter.heroku.com/articles/sendgrid
 [MailView]: https://github.com/37signals/mail_view
```
gabebw: Can we remove this link? Is it used anywhere now?

drapergeek: It looks to be used in the note about Rails version prior to 4.1.0.

rob-murray: I can remove the reference and link if you like? But Rails 4.1 isn't *really* old so previous version will be around and MailView is pretty useful tool still - don't want them to feel left out :innocent:

gabebw: No, I just missed it. Let's keep it in.

jferris:
Thanks! I merged this into master.

PR #297, opened 2015-01-28
Add stricter nesting rules for sass

kylefiedler:
I've been following the below and trying to keep my nesting no larger than 4 deep.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -38,8 +38,7 @@
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
 * Use ID and class names that are as short as possible but as long as necessary.
 * Avoid using the direct descendant selector `>`.
-* Avoid nesting more than 4 selectors deep.
-* Don't nest more than 6 selectors deep.
+* Avoid nesting more than 3 selectors deep.
 * Use HTML tags on vague classes that need a qualifier like `header.application` not `.main`.
 * Avoid using the HTML tag in the class name: `section.news` not `section.news-section`.
 * Avoid using HTML tags on classes for generic markup `<div>`, `<span>`: `.widgets` not `div.widgets`.


brendastorer:
Yes. The less nesting the better. :+1:

smharley:
Yeah, I've been using the new Sass 3.3+ `&` syntax and have basically stopped indenting all together. but :+1: for stricter guidelines

ryancoughlin:
Yes, yes yes to this.

ryancoughlin:
Keep those rules flexible and modular!

tysongach:
:+1:

style/sass/README.md
```diff
@@ -38,8 +38,8 @@ Selectors
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
 * Use ID and class names that are as short as possible but as long as necessary.
 * Avoid using the direct descendant selector `>`.
-* Avoid nesting more than 4 selectors deep.
-* Don't nest more than 6 selectors deep.
+* Avoid nesting more than 3 selectors deep.
```
wardpenney: Can this be 1 selector deep?

style/sass/README.md
```diff
@@ -38,8 +38,8 @@ Selectors
...
+* Don't nest more than 5 selectors deep.
```
wardpenney: And this just be removed?

wardpenney:
I just made some comment about 1 selector deep because I think any unnecessary increase in declaration specificity is a setup for pain down the road.

joshuaogle:
I would be against changing to 1 selector deep. I'm all about keeping things flat, but there's nothing wrong with some responsible "molecule" and "organism" level styles. I think 3/5 is a good rule.

wardpenney:
Either way, you would want stronger selector names. If you're nesting an `a` tag or a `.active` generic class name, you're setting yourself and the team up for trouble down the road.

jferris:
Any further discussion to have here?

jferris:
ping @thoughtbot/design

jessieay:
For the record, Hound already hounds you for using selectors more than 2 levels deep (or least it did for me today, not sure if that was a project setting or the default).

teoljungberg:
@jessieay It's 4 as the default https://github.com/thoughtbot/hound/blob/17a4ddcd82534455e8d4331b331d0e43754684f2/config/style_guides/thoughtbot/scss.yml#L67

kaishin:
`2` as a strict limit sounds good to me. `1` is leaving a good Sass feature on the table, and `3` is getting in over-specificity territory.

wardpenney:
I like the word "avoid" and the limit at 2, although I would prefer 1.

jasonramirez:
I think that `2` is a reasonable suggestion.

jessieay:
Hmmm here are screenshots of the comments I got and the entire `hound.yml` file. (would share links, but it's a private repo)...seems like there is a default of 2 being set somewhere.

screen shot 2015-06-08 at 10 32 15 am

screen shot 2015-06-08 at 10 32 30 am

teoljungberg:
@jessieay Huh, could you file an issue? And add the thoughtbot/hound team to the repo and we'll help debug

creuter:
@jessieay @teoljungberg The [`SelectorDepth`](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#selectordepth) setting (depth of applicability) is not quite the same as nesting.

They're very similar, though.

:+1: to the changes.

style/sass/README.md
jasonramirez: Personally, I *try* to keep it to no more than 1 level of nesting, but I do break that rule sometimes. To me, I feel that less nesting means stronger class names, more searchable class names, less complexity, and overall more thoughtful use of css. This makes me feel like the quality of my pattern goes up, and my code is *easier* to read. If we are making a style suggestion, I can get behind `Avoid nesting more than 3 selectors deep`, but I would prefer even `2 selectors deep`.

style/sass/README.md
jasonramirez: +1 - I think we can remove this as the previous line sets the recommended limit.

jasonramirez:
I would love to see:
> +* Avoid nesting more than 3 selectors deep.

And I don't think we need:
> -* Don't nest more than 5 selectors deep.

PR #299, opened 2015-01-28
Prefer `+` over `\` for concatenating long strings

gabebw:
In Ruby, `"a" "b" == "ab"`, i.e. whitespace can be used to concatenate strings. Due to this, the following can be used to concatenate strings:

```ruby
"hello" \
"world"
```

It is equivalent to `"hello" "world"`.

But we don't ever write `"hello" "world"`, we write `"hello" + "world"` and rely on `String#+` instead of niche syntax. Therefore we should use `+` to concatenate multi-line strings as well:

```ruby
"hello" +
  "world"
```

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -38,6 +38,7 @@ Ruby
 * Use heredocs for multi-line strings.
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
+* Prefer `+` over `\` for concatenating strings.
 
 [trailing comma example]: /style/ruby/sample.rb#L49
 [required kwargs]: /style/ruby/sample.rb#L16


gylaz:
I'm :+1: for this.

BlakeWilliams:
:+1:

drapergeek:
:+1:

mcmire:
Did not know this. But totally :+1:

calleerlandsson:
I think both methods of concatenation are equal from a readability standpoint and would prefer no guideline.

mike-burns:
I'd rather see:

* Avoid concatenating strings.

Instead of a new rule on which punctuation we should use.

jyurek:
"Try not to concatenate strings over, especially over multiple lines. But if you have to, use `+`. Don't just let whitespace handle it. Be explicit."

mike-burns:
Out of left field:

* Concatenate strings by joining an array of strings.

pbrisbin:
I would prefer no guideline here. I think having to do this is a minor smell anyway. When I do have to do it, I'd like freedom to choose between `+`, `\`, `<<`, HEREDOC, or `%{ }` on a case-by-case.

teoljungberg:
I vote for having no guideline aswell, guidelines are there to help you write more readable code. And I consider both ways to be equally readable

derekprior:
> I would prefer no guideline here. I think having to do this is a minor smell anyway. When I do have to do it, I'd like freedom to choose

This. Just about the only time I'm forced to do something like this is when printing a deprecation message and it comes up so infrequently that I don't think it matters what I choose to do.

mcmire:
This guideline isn't saying that you can't use arrays to join strings or use heredocs or whatever. You are still have freedom to do that, and I don't think there should be a guideline that says you can't. It just says that if you choose to use `\`, don't do that, use `+` instead, because `\` is antiquated syntax and relies on the fact that `"foo" "bar"` is equal to `"foobar"`, and no one would write that. 

Taking a step back, I would also pose a question, are we seeing this enough that we need a guideline around this?

gabebw:
People raise some good points:

* I hadn't considered the other ways to concatenate strings, like joining arrays (which I like) and `<<` (which, ehhh). I don't think heredocs are relevant here - this is more for creating a long single line of text, like an HTTP header. Heredocs are for creating a paragraph of text.
* I agree that concatenating strings over multiple strings should generally be avoided, but sometimes it can't be, like when creating a single long HTTP header line.
* I've seen `\` used for concatenating multi-line strings a few times this month, and I don't think it's a good idea. Why not just use `+`?

jyurek:
Point of clarification: `\` does not actually concatenate anything. Using `\` on a newline simply tells the parser there's more to the line, and lets Ruby's natural whitespace-concatenation happen. That's another reason I don't like this guideline: it's technically incorrect.

jyurek:
That said, I think we can safely not have this guideline, because it falls under our general guideline to be explicit. We can call this out in any relevant PRs where it's used with that as the reasoning.

pbrisbin:
Good point Jon. I wouldn't oppose a guideline like:

> Prefer explicit over implicit string concatenation

Or

> Avoid Ruby's implicit string concatenation

But agree, that overall guideline about explicitness may be enough to cover this case.

gabebw:
Jon, I think I explained the white-space-concatenation in the commit message?

I do like "Prefer explicit over implicit string concatenation", while keeping (in the commit message) the explanation of what `\` is actually doing.

jyurek:
You more or less did, but we're still using the "Use `+` instead of `\` for concatenation" wording while discussing it, implying `\` does concatenation. I just wanted to clarify.

I think if we change the guideline to that, it will just be confusing, because "implicit string concatenation" just sounds awkward if you didn't even know it was there.

dideler:
Another option is "prefer `<<` (appending) over `+` (concatenating) for long strings".

Concatenation returns a new string whereas appending mutates the same string. Sometimes you don't need a new string. This distinction matters more when performance matters, e.g. an often called tight loop where a long string is being built.

gabebw:
I haven't seen string concatenation be the bottleneck enough times to make `<<` into a full-fledged guideline.

mike-burns:
http://tinyrobots.thoughtbot.com/post/19782457/im-against-string-but-id-feel-bad-taking-it

jferris:
I personally prefer `\` to `+` for long, literal strings, because:

* `\` can only be used with literal strings. When you see `\` used in this way, you know nothing dynamic is going on.
* It's one of the few things Ruby can do at compile-time. You end up with one literal string.
* Since it happens at compile time, most mistakes result in a syntax error instead of going all the way to test failures.

gylaz:
FWIW, RuboCop recommends using `\` for concatenation:
```
Style/LineEndConcatenation: Use \ instead of + or << to concatenate those strings.
```

jferris:
I created a poll for this: https://docs.google.com/a/thoughtbot.com/forms/d/12EO32bHISw_XoQeIahNnmJXuFp5pep3wIA1nHurwkNY/viewform?usp=send_form

gabebw:
The first option is pro-backslash, right? Because it says:

```
Everyone should use (follows the bbatsov/rubocop recommendation)
```

jferris:
> The first option is pro-backslash, right?

Yes; fixed.

gabebw:
There were 8 responses:

* 5 said no guideline
* 1 said `+`
* 2 said `\`

Therefore I'm closing this without merging.

JuanitoFatas:
FYI, backslash could be much faster (~ 11x) for long string literals, see: http://thepugautomatic.com/2015/01/backslashy-ruby/#comment-1780720281 :blush:

mcmire:
There are probably several other things you should check first before concatenation if you're concerned about performance, but that's good to know.

PR #300, opened 2015-01-30
Add foreign key constraints Rails in migrations

croaky:


jferris:
This seems like a good idea to me.

derekprior:
:+1:

sikachu:
:+1:

mcmire:
I approve of this in theory, but I haven't actually tried this yet, so I can't say whether I'd approve of this in practice. Are there any pain points with adding constraints like this?

sikachu:
@mcmire no, but there's pain point when this constraint is _missing_.

derekprior:
When I add these some times shoulda-matcher type specs fail. Like if you're validating uniqueness on a FK column or something. You just have to create a record beforehand.

jferris:
Let's merge this!

srt32:
:+1:

croaky:
Okay!

PR #302, opened 2015-02-11
Add JavaScript style guide

BlakeWilliams:
This is the first shot at a JavaScript style guide to get us talking about what should go in it. I've also replaced "Use CoffeeScript" from best practices with "Use ES6" (but sightly longer). A good bit of this echoes the CoffeeScript guides, but that's expected.

I also need to try and expand on `sample.js`.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -185,8 +185,10 @@ Email
 
 JavaScript
 ----------
+* Use Coffeescript, ES6 with [babel], or another language that compiles to
+  JavaScript
 
-* Use CoffeeScript.
+[babel]: http://babeljs.io/
 
 HTML
 ----
--- a/style/javascript/README.md
+++ b/style/javascript/README.md
@@ -1,0 +1,24 @@
+JavaScript
+==========
+
+[Sample](sample.js)
+
+* Prefer ES6 classes over prototypes.
+* Use strict equality checks (`===` and `!==`) except when comparing against
+  (`null` or `undefined`).
+* Prefer [arrow functions] `=>`, over the `function` keyword except when
+  defining classes or methods.
+* Use semicolons at the end of each statement.
+* Prefer double quotes.
+* Use `PascalCase` for classes, `lowerCamelCase` for variables and functions,
+  `SCREAMING_SNAKE_CASE` for constants, `_singleLeadingUnderscore` for private
+  variables and functions.
+* Prefer [template strings] over string concatenation.
+* Prefer promises over callbacks.
+* Prefer array functions like `map` and `forEach` over `for` loops.
+* Use `const` for declaring variables that will never be re-assigned, and `let`
+  otherwise.
+* Avoid `var` to declare variables.
+
+[template strings]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings
+[arrow functions]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions
--- a/style/javascript/sample.js
+++ b/style/javascript/sample.js
@@ -1,0 +1,7 @@
+object = { spacing: true }
+
+class Cat {
+  canBark() {
+    return false;
+  }
+}
--- a/style/README.md
+++ b/style/README.md
@@ -20,6 +20,7 @@ detailed, language/framework-specific style guides:
 * [Git](git)
 * [Haskell](haskell)
 * [HTML](html)
+* [JavaScript](javascript)
 * [Objective C](objective_c)
 * [Python](python)
 * [Rails](rails)


style/javascript/README.md
```diff
@@ -0,0 +1,18 @@
+JavaScript
+==========
+
+[Sample](sample.js)
+
+* Prefer ES6 classes over library provided classes.
+* Prefer `===` and `!===` to `==` and `!=`.
+* Avoid using arrow functions when not necessary.
```
christoomey: Is there any more guidance you can can provide to clarify "when necessary" or why to avoid? Is it just for binding `this` with fat arrow?

BlakeWilliams: I've seen a few people take advantage of `=>` being shorter which makes code misleading since you assume it will be using the defining scope's `this` value.

`Avoid using arrow functions when not using the defining scope's `this` value`?

christoomey: I see the value in being explicit about using `this`, but I don't like giving up on the clarity arrow functions can provide with implicit return and shorter syntax

``` es6
users.map(user => user.name)

// vs
users.map(function(user) { return user.name })
```

drapergeek: > which makes code misleading since you assume it will be using the defining scope's this value.

Is this true? I don't make that assumption. I've worked on projects where the rule was to simply always use the fat arrow. For some reason I always thought that was a bad idea but I have never actually ran into a time where it caused an issue.

Is there ever a case where this would actually cause a problem?

style/javascript/README.md
```diff
@@ -0,0 +1,18 @@
...
+* Use semicolons and the end of each statement.
+* Prefer double quotes.
+* Use `PascalCase` for classes, `lowerCamelCase` for variables and functions,
+  `SCREAMING_SNAKE_CASE` for constants, `_single_leading_underscore` for private
```
seanpdoyle: How do you feel about `_singleLeadingUnderscore` for private variables and functions?

I would prefer to stick with camelCasing all around.

style/javascript/README.md
BlakeWilliams: Not a technical problem, but I like being able to visually distinguish the two.

@christoomey That is an edge case to the rule and I would definitely prefer to see that. 

Any ideas on how we can word this and make sense?

style/javascript/README.md
drapergeek: agreed.

BlakeWilliams: That's a typo on my part, `_singleLeadingUnderscore`.

seanpdoyle:
How do we feel about something like https://twitter.com/raganwald/status/564792624934961152

> ES6 Conventions:

> 1. use `const` by default.
> 2. use `let` if you have to rebind a variable.
> 3. use `var` to signal untouched legacy code.

style/javascript/README.md
drapergeek: > Not a technical problem, but I like being able to visually distinguish the two.

This sounds a lot like the discussion around single vs double quotes. If we can default to always using the fat arrow and never have to come back and change it, why not? I don't think the readability actually changes that much.

Are there any technical downsides to using the fat arrow everywhere?

BlakeWilliams:
That has a very Swift/Rust/Scala feel to it.

I'm not opposed to it, but I'm curious how others feel about it.

style/javascript/README.md
BlakeWilliams: Besides a tiny bit more code generated when transpiled, no.

drapergeek: I'm in favor of changing this to always use the fat arrow then. That prevents possible confusion over things not working as expected and it is one less thing to have to worry about.

BlakeWilliams: > Prefer arrow functions `=>`, over the `function` keyword

Better wording?

drapergeek: That sounds good to me.

best-practices/README.md
```diff
@@ -183,7 +183,9 @@ Email
 JavaScript
 ----------
 
-* Use CoffeeScript.
+* Use ES6 when possible with a tool like [6to5].
```
gabebw: How about using "Prefer", since we define it in the README?

* Prefer ES6 with a tool like 6to5.
* If 6to5 is not possible, use CoffeeScript (?)

style/javascript/README.md
```diff
@@ -0,0 +1,18 @@
+JavaScript
+==========
+
+[Sample](sample.js)
+
+* Prefer ES6 classes over library provided classes.
```
gabebw: I don't understand - like what classes?

style/javascript/README.md
gabebw: What's an arrow function? Are there docs?

style/javascript/README.md
```diff
@@ -0,0 +1,18 @@
...
+* Prefer `===` and `!===` to `==` and `!=`.
+* Avoid using arrow functions when not necessary.
+* Use semicolons and the end of each statement.
```
gabebw: AT the end

style/javascript/sample.js
```diff
@@ -0,0 +1,7 @@
+object = {spacing: false}
```
gabebw: Why no spacing?

gabebw: Er. Why not put spaces around the curly braces? I like  those spaces, because it makes it easier to read. Same reason I like 'em in Ruby.

BlakeWilliams: I've been skipping the spacing on the latest project because it can look/feel weird when nested inline. Most of the time it will look like this:

```
{
  spacing: false
}
```

So I don't think it's that bad. I don't feel strongly about this one, so if we prefer the space I'm fine with that too.

gabebw: I was actually referring to single-line spacing. 

I prefer this:

```js
{ spacing: true }
```

to this:

```js
{spacing: false}
```

style/javascript/README.md
BlakeWilliams: `React.createClass` (unless you need mixins), `_.extend`, `Backbone.Router.extend`, etc.

BlakeWilliams: Maybe `Use ES6 classes` would be a better cover all since it includes `function Cat() {}` and etc.

best-practices/README.md
mike-burns: Ideally, a best practice won't change often (for example, "avoid global variables" and "keep classes small"). Will 6to5 stand the test of time? Or is it just a current solution?

BlakeWilliams: The best practice is `Use ES6`, I just point to the best solution (I know of) currently instead of keeping it ambiguous.

style/javascript/README.md
BlakeWilliams: Well, `Prefer` would probably fit better.

mike-burns:
Is it too soon to have an ES6 styleguide? Do we have enough experience with it? How does it compare to the community's styleguide?

style/javascript/README.md
gabebw: But if I'm using Backbone, I'll be using "library-provided classes" no matter what, right? So does this mean "avoid backbone because it's not in ES6"?

BlakeWilliams: No, I'd say it means do things like `class Router extends Backbone.Router` instead of `var Router = Backbone.Router.extend`.

I don't think we should be using Backbone anymore, but I chose it since it should be familiar to a lot of us.

BlakeWilliams: Another example:

```
class Cat {
  canBark() {
    return false;
  }
}
```

vs:

```
var Cat = function() {
  this.canBark = function() {
    return false;
  }
}
```

style/javascript/sample.js
BlakeWilliams: I'll change it up then.

BlakeWilliams:
I feel like I've written enough ES6 to at least start a guide. I think we'll end up adding/changing it as more of us use ES6.

best-practices/README.md
BlakeWilliams: > Use ES6 when possible with a tool like 6to5, otherwise use CoffeeScript

Better?

derekprior: This sounds like it's changing ES6 to our default in projhects (over coffeescript). Is that what we've determined?

style/javascript/README.md
```diff
@@ -0,0 +1,19 @@
+JavaScript
+==========
+
+[Sample](sample.js)
+
+* Prefer ES6 classes over library provided classes.
+* Prefer `===` and `!===` to `==` and `!=`.
```
derekprior: `!==`, right?

best-practices/README.md
BlakeWilliams: I'd like it to be a default, but `Use CoffeeScript` isn't really the answer anymore.

style/javascript/README.md
BlakeWilliams: Thoughts on changing it to `Use ES6 classes` or removing it?

style/javascript/README.md
BlakeWilliams: Nice catch, pushed a fix.

style/javascript/README.md
```diff
@@ -0,0 +1,22 @@
+JavaScript
+==========
+
+[Sample](sample.js)
+
+* Prefer ES6 classes over library provided classes.
+* Prefer `===` and `!==` to `==` and `!=`.
```
derekprior: The use cases for `==` are very specific (IMO). Should we just specify them:

* Use strict equality checks (`===` and `!==`) except when comparing against (`null` or `undefined`).

BlakeWilliams: Yeah, I like that wording a lot better.

best-practices/README.md
jferris: We have an ongoing experiment related to ES6: https://trello.com/c/EGpNLWB1/393-es6-instead-of-coffeescript

It seems like the winds are blowing this way, but I think we should try out ES6 on a JavaScript-heavy project before determining it's ready to replace Coffeescript on projects going forward.

I think it would be fine to get guidelines in place for HOW to use ES6 in the event that you are using it. I just think we should have a JavaScript ES6 section instead of replacing the existing Coffeescript recommendation.

It could also be reasonable to soften this to something like, "Use Coffeescript, ES6 with 6to5, or another language that compiles to JavaScript." Basically, make the stance that JavaScript is a compile target.

style/javascript/README.md
jferris: What about: "Prefer ES6 classes over prototypes"?

best-practices/README.md
BlakeWilliams: I'm a fan of "Use Coffeescript, ES6 with 6to5, or another language that compiles to JavaScript.". I think it does a good enough job of saying use JavaScript but avoid ES5.

style/javascript/sample.js
```diff
@@ -0,0 +1,7 @@
+object = { spacing: true }
+
+class Cat {
+  canBark() {
+    return false;
```
jferris: What's the deal with semicolons in ES6?

whatsthedeal

But seriously forks, do people use semicolons everywhere or can we eliminate this noise?

BlakeWilliams: I use semicolons to avoid having to know the [rules of ASI](http://bclary.com/2004/11/07/#a-7.9.1). In practice it's usually not a big deal but can sometimes lead to confusing bugs.

style/javascript/README.md
BlakeWilliams: Yeah, that works for me.

style/javascript/sample.js
derekprior: I don't think automatic semicolon insertion has changed at all in ES6. So unless you're willing to live and die by its edge cases, then they have to stick around.

BlakeWilliams: It hasn't AFAIK.

jferris: boo-hiss

BlakeWilliams: semicolons

gylaz:
@BlakeWilliams What do you think about including JSHint dependency to this JS style guide? It might be even good to provide a `.jshintrc` file.

BlakeWilliams:
Would that go under best practices? I can definitely see JShint being a best practice.

Also, I'm not sure if we should include a jshintrc since it's sometimes environment dependent.https://github.com/jshint/jshint/blob/master/examples/.jshintrc

gylaz:
My thoughts around providing tool configs, is that a lot of folks that want to follow our style guide could just grab the config file, plug it into their project and have automated checks against thoughtbot style guide. I've had a client request thoughtbot's RuboCop config file, and I think this could helpful across the board.

BlakeWilliams:
Would it make sense to point them towards Hound in those instances if applicable?

I think a lot of our guides aren't really enforceable with tools like jshint. eg: Use `const` by default, `let` when mutating, avoid `var`.

style/javascript/README.md
```diff
@@ -0,0 +1,24 @@
...
+  defining classes or methods.
+* Use semicolons at the end of each statement.
+* Prefer double quotes.
+* Use `PascalCase` for classes, `lowerCamelCase` for variables and functions,
+  `SCREAMING_SNAKE_CASE` for constants, `_singleLeadingUnderscore` for private
+  variables and functions.
+* Prefer [template strings] over string concatenation.
+* Prefer promises over callbacks.
+* Prefer array functions like `map` and `forEach` over `for` loops.
+* Use `const` to declare variables.
```
mcmire: Since nothing in JavaScript is truly immutable (except for numbers), I wouldn't go so far as to use `const` (except for actual constants). However, I *would* use `let` over `var` -- the block scope works like Ruby so it's familiar, the block scope also prevents scope issues with things like `for`, and `let`s aren't hoisted to the top of functions so you can't declare them out of order.

mcmire: I think it *is* interesting, though, that `const`s can't be reassigned. That might be useful (I have pretty much never needed to use the same name for a variable).

BlakeWilliams: I know they're not *really* mutable, but it's mostly intent. You should strive to have as little mutability in your code as possible and I think this helps reinforce that with warnings and convention.

BlakeWilliams: Also, Chrome thinks they're immutable I guess.
screen shot 2015-02-11 at 4 16 23 pm

BlakeWilliams: Not really immutable, but can't be reassigned*

BlakeWilliams:
It's been a while, anyone have more thoughts on this?

teoljungberg:
I dislike like that we're having a styleguide based on a new technology that we haven't used extensively. And it's definately not a best practice

BlakeWilliams:
That's fair, but "Use CoffeeScript" isn't a best practice either.

derekprior:
> I dislike like that we're having a styleguide based on a new technology that we haven't used extensively. 

Whether you have a guide or not, the code we right has a style. We can certainly change things as we write more ES6, but putting a stake in the ground seems to make sense.

mike-burns:
Does the ES6 community have an existing styleguide that we can refer to?

BlakeWilliams:
https://github.com/mozilla/addon-sdk/wiki/Coding-style-guide is the best example I could find of one.

teoljungberg:
> That's fair, but "Use CoffeeScript" isn't a best practice either.

Fair point.


> Whether we have a guide or not, the code we write has a style. We can certainly change things as we write more ES6, but putting a stake in the ground seems to make sense.

Dito, can we do something like @mike-burns suggested and refer to the community based styleguide and then "override" things in our own?

BlakeWilliams:
I don't think there's a "blessed" ES6 style guide out there that people reference often. I assume the closest would be what I linked above, which is similar to what we already have here in a lot of cases.

I'm hesitant to refer to it because it suggests things like not using braces for one liners in if statements. I also think it's much easier to start with our own guides telling what *to* do rather than reference another guide and saying what *not* to do from it.

style/javascript/README.md
mcmire: They can be reassigned, but they can't be redeclared... isn't that the distinction?

BlakeWilliams: No, you get `Uncaught TypeError: Idnetifier 'b' has already been declared'` when you do this:

```javascript
const b = 1
// undefined

b = 2
// 2

b
//2

const b = 5
//  Uncaught TypeError: Identifier 'b' has already been declared
```

BlakeWilliams: If it's an object you can modify its properties, but you can't reassign it or re-declare it.

jferris: If you can't assign a new value, I would say this is not a "variable."

How about just "Prefer `const` to `var`?"

BlakeWilliams: Ideally we shouldn't be using `var` in any code written in ES6. In cases where you would use `var` we now use `let` due to the scoping differences.

BlakeWilliams: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let has some good information on the differences

mcmire: "Prefer `const` to `var` or `let`"?

BlakeWilliams:
This has been sitting here for a while, any objections to merging?

style/javascript/README.md
BlakeWilliams: I think it's more "Use `const` when you don't have to rebind the value, otherwise use `let` and avoid `var`" so I'm having a hard time making it a single rule.

mcmire: What about:

* Use `const` for declaring variables that will never be re-assigned, and `let` otherwise.
* Avoid `var` to declare variables.

mcmire:
I had one tweak above for the `const` rule, otherwise LGTM.

style/javascript/README.md
BlakeWilliams: I'm a fan.

PR #304, opened 2015-02-11
Clarify how Rails associations should be ordered

gylaz:
There are two options how the current wording can be interpreted. This changes
the wording to favor the B option.

```rb
# A
has_many :apples
belongs_to :tree
has_many :watermelons

# B
belongs_to :tree
has_many :apples
has_many :watermelons
```

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -8,7 +8,8 @@ Rails
 * Name time columns (referring to a time of day with no date) with `_time`
   suffixes.
 * Name initializers for their gem name.
-* Order ActiveRecord associations alphabetically by attribute name.
+* Order ActiveRecord associations alphabetically by association type, then
+  attribute name.
 * Order ActiveRecord validations alphabetically by attribute name.
 * Order ActiveRecord associations above ActiveRecord validations.
 * Order controller contents: filters, public methods, private methods.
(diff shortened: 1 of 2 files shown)


gabebw:
I agree with this.

jferris:
Do we want an example to go along with this?

gylaz:
Example added.

style/rails/sample.rb
```diff
@@ -0,0 +1,6 @@
+class SomeClass
+  belongs_to :tree
+  has_many :apples
+  has_many :watermelons
+end
+
```
gabebw: extra line

style/rails/README.md
```diff
@@ -8,7 +8,8 @@ Rails
 * Name time columns (referring to a time of day with no date) with `_time`
   suffixes.
 * Name initializers for their gem name.
-* Order ActiveRecord associations alphabetically by attribute name.
+* Order ActiveRecord associations alphabetically by association type, then
+  attribute name. [Example][order-associations]
```
gabebw: What if we put the link on " Order ActiveRecord associations alphabetically" so the link flows as part of the text?

dgalarza:
:+1:

teoljungberg:
:+1:

style/rails/README.md
gylaz: I'm not opposed to that idea, thought that would be inconsistent with how we've been providing examples so far. I'm okay with it if we decided to do examples in this manner going forth.

gabebw: Let's keep it consistent! Therefore, let's put a period after "Example".

PR #305, opened 2015-02-13
Use only one Haskell pragma per line

jferris:
* With too many on one line, it gets ugly and you'll be forced to wrap.
* You can be consistent if there's only one per line.

### DIFF
--- a/style/haskell/README.md
+++ b/style/haskell/README.md
@@ -5,13 +5,16 @@ Haskell
 
 * Break long expressions before operators or keywords.
 * Break long type signatures between arguments.
+* Use a blank line between pragma and module statements.
 * Order imports in three sections, separating each by a blank line:
   [standard libraries], third-party libraries, application modules.
   Within each section, alphabetize the imports and place qualified
   imports last.
+* Order pragma statements alphabetically.
 * Use comma-first style exports, records, and lists.
 * Use four-space indentation except the `where` keyword which is
   indented two spaces. [Example].
+* Use only one pragma statement per line.
 
 [standard libraries]: http://www.haskell.org/ghc/docs/latest/html/libraries/index.html
 [Example]: /style/haskell/sample.hs#L41
--- a/style/haskell/sample.hs
+++ b/style/haskell/sample.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+
 module Haskell
     ( SomeType(..)
     , someFunction


style/haskell/README.md
```diff
@@ -12,6 +12,7 @@ Haskell
 * Use comma-first style exports, records, and lists.
 * Use four-space indentation except the `where` keyword which is
   indented two spaces. [Example].
+* Use only one pragma statement per line.
```
pbrisbin: I'd also add that they be alphabetized, or an additional rule for that?

style/haskell/sample.hs
```diff
@@ -1,3 +1,6 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+
```
pbrisbin: I don't know that we need a rule about this, but I *don't* prefer this extra whitespace.

style/haskell/README.md
jferris: Agreed that they should. I added another bullet in the "Order" section.

style/haskell/sample.hs
jferris: I think that it would be good to be consistent. I prefer the whitespace, but you've written and read more Haskell than I have, so I'll defer to you.

Updated with the space removed and a guideline not to include it.

jferris:
Ready for re-review.

pbrisbin:
Awesome. I feel so authoritarian.

LGTM

jferris:
Cool. I'm going to leave this open for a bit for feedback from others.

Note: I probably won't leave this open the usual full week, because there aren't as many people interested in Haskell formatting and I think it's better to move faster at the beginning.

style/haskell/README.md
```diff
@@ -5,13 +5,16 @@ Haskell
 
 * Break long expressions before operators or keywords.
 * Break long type signatures between arguments.
+* Don't use a blank line between pragma and module statements.
```
BlakeWilliams: I don't have much of an opinion on it, but why no blank line? I usually add one when I write Haskell.

pbrisbin: There's absolutely no reason whatsoever, just something to decide for
consistency.

To me, having the space just looks weird, like putting a space between a
method and the comment documenting it.

Just based on the code I see around, no space is more common. No idea what
the real numbers could be though.

mike-burns: I couldn't find any code on GitHub that formats pragma statements without a subsequent blank line. What code are you looking at?

(On GitHub search results, ghc was the biggest user of language pragmas.)

pbrisbin: > What code are you looking at?

Anything Yesod-related. Maybe it's a Snoyman thing?

mike-burns: Snoyman!!!! I hadn't gotten to his repos in the GitHub search I did. I just noticed that Snoyman did this to my github library, too. Bah. (Relatedly: GH is really bad at uniquing their search results.)

How about dropping this guide suggestion for now; we can revisit it after we write more code? The pragmas are distinct from the modules, afterall.

pbrisbin: > How about dropping this guide suggestion for now; we can revisit it after we write more code?

Fine with me.

jferris: > How about dropping this guide suggestion for now; we can revisit it after we write more code?

I'm not terribly opposed to this, but I'd rather have a guideline and be consistent. Neither way is actually better. I feel like it will just be harder to pick one or the other later, as we'll have more code following different formats.

It looks like ghc-base (probably the largest repository of blessed Haskell code) uses a space after the pragma lines: https://github.com/ghc/packages-base

Any objection to using ghc as a reference for formatting when there's no obvious winner?

mike-burns: I'm fine with just picking a thing and being consistent, and following ghc makes sense to me.

pbrisbin: Yup, same here.

jferris: Updated.

jferris:
Any last comments here?

mike-burns:
LGTM.

PR #306, opened 2015-02-16
Expand on why let should be avoided

gylaz:
Explain that re-implementing `let` is not the answer.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -108,7 +108,9 @@ Testing
 -------
 
 * Avoid `any_instance` in rspec-mocks and mocha. Prefer [dependency injection].
-* Avoid `its`, `let`, `let!`, `specify`, and `before` in RSpec.
+* Avoid `its`, `specify`, and `before` in RSpec.
+* Avoid `let` (or `let!`) in RSpec. Prefer extracting helper methods,
+  but do not re-implement the functionality of `let`. [Example][avoid-let].
 * Avoid using `subject` explicitly *inside of an* RSpec `it` block.
   [Example][subject-example].
 * Avoid using instance variables in tests.
@@ -127,6 +129,7 @@ Testing
 
 [dependency injection]: http://en.wikipedia.org/wiki/Dependency_injection
 [subject-example]: ../style/testing/unit_test_spec.rb
+[avoid-let]: ../style/testing/avoid_let_spec.rb
 [`Delayed::Job` matcher]: https://gist.github.com/3186463
 [stubs and spies]: http://robots.thoughtbot.com/post/159805295/spy-vs-spy
 [assertions about state]: https://speakerdeck.com/skmetz/magic-tricks-of-testing-railsconf?slide=51
--- a/style/testing/avoid_let_spec.rb
+++ b/style/testing/avoid_let_spec.rb
@@ -1,0 +1,52 @@
+# BAD
+describe ReportPolicy do
+  let(:user) { User.new(report_ids: [1, 2]) }
+  let(:report) { Report.new(id: 2) }
+
+  describe "#allowed?" do
+    context "when user has access to report" do
+      it "returns true" do
+        policy = ReportPolicy.new(user, report)
+
+        expect(policy).to be_allowed
+      end
+    end
+
+    context "when user does not have access to report" do
+      it "returns false" do
+        report.id = 3
+        policy = ReportPolicy.new(user, report)
+
+        expect(policy).not_to be_allowed
+      end
+    end
+  end
+end
+
+# GOOD
+describe ReportPolicy do
+  describe "#allowed?" do
+    context "when user has access to report" do
+      it "returns true" do
+        policy = build_policy(report_id: 2, allowed_report_ids: [1, 2])
+
+        expect(policy).to be_allowed
+      end
+    end
+
+    context "when user does not have access to report" do
+      it "returns false" do
+        policy = build_policy(report_id: 3, allowed_report_ids: [1, 2])
+
+        expect(policy).not_to be_allowed
+      end
+    end
+  end
+
+  def build_policy(report_id:, allowed_report_ids:)
+    user = instance_double("User", report_ids: allowed_report_ids)
+    report = instance_double("Report", id: report_id)
+
+    ReportPolicy.new(user, report)
+  end
+end


style/testing/avoid_let_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    context "don't re-implement" do
+      it "uses a re-implementation of let" do
+        some_variable
+
+        expect(true).to be_true
+      end
+
+      def some_variable
+        # BAD -- just use let, or better extract a function that allows passing
+        # the thing that is different accross the examples.
```
graysonwright: @gylaz instead of explaining this in a comment, can you add an example of the function you mention?

I'm not entirely sure what you mean by `a function that allows passing the thing that is different accross the examples`, so I feel an example would help here.

mcmire: Agreed.

best-practices/README.md
```diff
@@ -110,6 +110,10 @@ Testing
 
 * Avoid `any_instance` in rspec-mocks and mocha. Prefer [dependency injection].
 * Avoid `its`, `let`, `let!`, `specify`, and `before` in RSpec.
+  * Instead of using these, extract a method that you can pass the thing that
```
BlakeWilliams: I get a bit tripped up when reading this sentence, thoughts on `Instead of using using the above methods, extract a method that accepts the differing data as an argument` or something similar?

mike-burns:
The Best Practices README already says "These are not to be blindly followed; strive to understand these and ask when in doubt."

If people are blindly following a best practice, maybe we should just remove it.

style/testing/avoid_let_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        # BAD
+        user.name = "Jane"
+        result = SomeClass.new(user).run
+
+        expect(result).to be_present
+      end
+
+      it "does not use lets" do
+        # GOOD
+        user = build_user(name: "Jane")
```
iwz: is `build_user` a method that would go into this spec somewhere? is it factory_girl? something else?

best-practices/README.md
jferris: These are the only sub-bullets in the guides. Is there a way we can flatten this?

style/testing/avoid_let_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe SomeClass do
+  describe "#let" do
+    let(:user) { User.new(name: "John") }
+
+    context "avoid using" do
+      it "uses lets" do
+        # BAD
```
iwz: Why is this bad?

style/testing/avoid_let_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        user.name = "Jane"
+        result = SomeClass.new(user).run
+
+        expect(result).to be_present
+      end
+
+      it "does not use lets" do
+        # GOOD
```
iwz: Why is this good?

best-practices/README.md
```diff
@@ -110,6 +110,10 @@ Testing
...
+    is different accross the tests. [Example][avoid-let].
```
jessieay: typo: `across`

style/testing/avoid_let_spec.rb
mcmire: I feel like you don't need the rest of this sentence. It appears that this context assumes that you're re-implementing let and the value on the right hand side is simple so it will never have to be customized. So I would just say something like

> BAD -- it's okay to use let for simple values that you want to memoize

style/testing/avoid_let_spec.rb
mcmire: Here I would put the wording about extracting a helper method in order to customize the thing that's being shared across examples.

best-practices/README.md
mcmire: This point only applies to `let` and `let!`. It could also apply to using `before` to set an instance variable. I've also noticed that `let` is used poorly specifically when building objects or factories. So, what about:

> Using `let` (or `let!`) particularly to build objects prevents those objects from being customized in tests that need to so. Instead, prefer extracting a helper method.

gylaz:
I've flattened and simplified the wording. Also, provided a more specific example. Ready for another look.

mcmire:
:+1:

best-practices/README.md
```diff
@@ -109,7 +109,9 @@ Testing
 -------
 
 * Avoid `any_instance` in rspec-mocks and mocha. Prefer [dependency injection].
-* Avoid `its`, `let`, `let!`, `specify`, and `before` in RSpec.
+* Avoid `its`, `specify`, and `before` in RSpec.
+* Avoid `let` (or `let!`) in RSpec. Prefer extracting helper method,
```
jferris: "helper method**s**?"

gylaz: Done.

nickrivadeneira:
I've generally disliked this guideline. While it's easy to abuse `let` and `before`, the guideline and "Let's Not" blog post are often used to say "never use `let` or `before` under any circumstance" without any understanding of what uses of the helpers are bad and why.

I do think we should have some guideline to prevent abuse, but I think it would be better if we gave examples of appropriate uses of `let` and `before` alongside inappropriate uses.

gylaz:
@nickrivadeneira I'm not sure on your overall assessment of this PR. Are you against this improvement to the general statement we had before? Do you think that the provided examples are lacking and can be improved?

nickrivadeneira:
@gylaz I agree with the sentiment of the PR (Expand on why `let` should be avoided), but I don't think the guideline change or examples serve that purpose. The changes show an alternative solution to `let`, but don't provide any clarity on the reasoning behind the guideline. Looking at the outdated diff comments, it looks like the original changes might have provided more clarity.

style/testing/avoid_let_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    context "when user does not have access to report" do
+      it "returns false" do
+        policy = build_policy(report_id: 3, allowed_report_ids: [1, 2])
+
+        expect(policy).not_to be_allowed
+      end
+    end
+  end
+
+  def build_policy(report_id:, allowed_report_ids:)
```
mcmire: Taking a second look at this I actually agree with what @nickrivadeneira is saying -- maybe we add some comments here on how is this is preferred?

gylaz:
Is anyone opposed to merging as is? The way I see it, it expands on our thoughts around `let`. It can be perfected further, yes. This PR has been sitting around for a while, and frankly I don't have a concrete idea on how to improve it further. Maybe add some of the good stuff from https://github.com/thoughtbot/guides/issues/322#issuecomment-86574235, but I would propose doing that in a separate PR.

Thoughts @jferris @mcmire @nickrivadeneira?

mcmire:
Nah, ship it.

PR #309, opened 2015-02-17
Prefer Sub-Routes over Maintaining State

seanpdoyle:
http://emberjs.jsbin.com/puyude/2/edit

Canonical modal example:

Instead of maintaining a `showModal` flag in a controller or route that wraps
markup or a component, prefer introducing a new sub-route with its own model,
transitions, etc.

Benefits:

* updates the URL
 * deep linkable
 * back button closes the modal (for free)
* new route -> new model -> less scattered state
* simple controllers
* fewer actions to handle
* introduces the possibility to `link-to` to open a modal, instead of
 being forced to handle an `action`

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -298,6 +298,7 @@ Ember
 * Use `href="#"` for links that have an action.
 * Prefer dependency injection through Ember initializers over globals on window
   or namespaces.
+* Prefer sub-routes over maintaining state.
 
 Testing
 


BlakeWilliams:
I agree with your points, but I think it's too vague and situational to have a guideline on it.

I'm :+1: for no guideline.

seanpdoyle:
I agree, the actual change is prose is pretty vague. I was hoping that the discussion surrounding the change would bring about better wording.

BlakeWilliams:
How about this: `Prefer stateful routes over stateful controllers`?

seanpdoyle:
@BlakeWilliams in the nested approach, the route doesn't maintain any state, it only provides a model. 

The "state" (is the modal open or not) is maintained by the URL.

seanpdoyle:
Also, general disclaimer to my responses to feedback:

> Prefer sub-routes over maintaining state.

is pretty awful and desperately needs improvement.

nickrivadeneira:
@seanpdoyle Also agree with your points, and I think it's worth having a guideline for. If you write the blog post first, it may provide some insight into a better guideline point.

nickrivadeneira:
Then you could link to the blog post in the guideline, similar to what we do with `let` and `before` guidelines.

endash:
Wouldn't this necessarily create a bifurcation amongst your modals if, for instance, some can be modeled as routes but others (login?) cannot be? It would seem preferable to maintain uniform behaviour for the user even at the expense of a less than ideal development technique.

seanpdoyle:
@endash could you elaborate a little more?

What would stop login from being managed by a `LoginRoute`?

endash:
What I meant by that example wasn't clear: a modal that could be accessed from virtually any route, or at least enough routes that duping the subroute all over the place would be prohibitive

seanpdoyle:
Also, it might be worth noting the connotation of `Prefer` within in our guides (https://github.com/thoughtbot/guides/blob/master/README.md#guides)

If a certain circumstance calls for stateful modals, by all means use stateful modals.

> "Use" is a positive instruction.

Would `Use` be a better fit?

seanpdoyle:
The purpose of this guideline is to encourage developers to think twice, and strongly consider adding a new nested route instead of overloading an existing route / controller / view trio.

seanpdoyle:
I've altered the jsbin to include an example with the modal state being persisted to a `?showModal` query param. 

This affords deep-linkability, back button closing, and all the benefits of the deeply nested routes approach, but with all the drawbacks of guard clauses and managing variable state.

The more I expand on this, the more I want to blog about it.

jferris:
@seanpdoyle any updates here?

seanpdoyle:
I've published https://robots.thoughtbot.com/embracing-ember-routes, which was the inspiration behind this guide.

I've also mentioned the post to @tomdale @rwjblue and @trek. Barring any negative feedback or corrections, I think this would be good to go into the guides.

jferris:
Any objections to merging this?

drapergeek:
:+1:

seanpdoyle:
Have we linked to blog posts explaining specific guidelines in the past?

https://robots.thoughtbot.com/embracing-ember-routes was a byproduct of this PR

trek:
Sorry, I've been way too busy since EmberConf to look :(

PR #310, opened 2015-02-18
Drop the branch prefix requirement

mike-burns:
It sometimes makes sense to prefix git branch names with your initial,
and it sometimes doesn't. Drop the requirement and let people do what
they deem right.

### DIFF
--- a/protocol/git/README.md
+++ b/protocol/git/README.md
@@ -24,8 +24,6 @@ Create a local feature branch based off master.
     git pull
     git checkout -b <branch-name>
 
-Prefix the branch name with your initials.
-
 Rebase frequently to incorporate upstream changes.
 
     git fetch origin
--- a/style/git/README.md
+++ b/style/git/README.md
@@ -2,7 +2,6 @@ Git
 ===
 
 * Avoid merge commits by using a [rebase workflow].
-* Prefix feature branch names with your initials.
 * Squash multiple trivial commits into a single commit.
 * Write a [good commit message].
 


teoljungberg:
:+1:

dgalarza:
:+1:

mcmire:
When doesn't it make sense to prefix the branch with your initials?

gylaz:
Prior discussion on Trello: https://trello.com/c/N1eKbFhL/308-drop-branch-prefix-guideline

Keithbsmiley:
:-1:

drapergeek:
:-1:

BlakeWilliams:
:-1:

jyurek:
I see no reason to not use initial-prefixes, but OTOH I see no reason to force them. I believe they are a best practice, but don't think they necessarily need to be required.

gabebw:
I believe they are a best practice, and make following up on stale branches easy, instead of impossible. When does it not make sense to prefix the branch?

tabfugnic:
@gabebw I don't know if it becomes impossible to follow up on stale branches. I think you can look at commit history pretty easily. Though I will admit it does make it easier at a glance. I have been voluntarily omitting prefixes from branches on some projects and it doesn't seem to have a great effect.

mike-burns:
If a branch is stale, delete it.

GitHub lists stale branches for you (though it/we had some confusion about tags, maybe?): https://github.com/thoughtbot/paperclip/branches/stale

stramel:
:+1:

pbrisbin:
As long as I can continue to prefix my branches, I'm fine with no guideline.

mike-burns:
Yup; name your branch anything you want.

joshuaclayton:
:+1:

mcmire:
I just read over the original discussion and it sounds like most of us are okay with the prefix, and only a few object. I guess I'm fine with no guideline, though. :skull:

mike-burns:
Typically when there is no consensus, we have no guideline.

iwz:
:+1:

iwz:
I've created a PR to add a `git branches` alias that gives a nice report of who is working on which branches https://github.com/thoughtbot/dotfiles/pull/347

graysonwright:
I'm :+1: for this PR because I don't feel strongly about prefixes. They're appropriate in some cases, but not in all cases.

calleerlandsson:
I'm in favor of dropping the requirement.

On the last two client projects I have been on, either me or a designer would create a branch and then the other would finish the story/feature and open a PR. I do not see how prefixing the branches with initials improved our workflow.

mike-burns:
@Keithbsmiley @drapergeek @BlakeWilliams do you still want to mandate an initials prefix requirement on every git branch name?

drapergeek:
I like branch prefixing. I believe that removing it from the guides will mean that as the company grows that workflow will fall completely the wayside because new employees will not know about it. I don't think that is the intention of removing it from the guides (or maybe it is).

I don't believe that requiring it on every project makes sense but I also think that we're professional enough to decide when we want to ignore a rule.

tldr; yes I think we should keep the guide but using the word mandate is a bit strong.

BlakeWilliams:
I agree with what Draper said. I don't think every project *has* to have it, but I think there is value in keeping it.

mike-burns:
As-is, this PR removes the requirement for it. If you can think of a phrasing that makes you happier, just let me know.

iwz:
> as the company grows that workflow will fall completely the wayside

@drapergeek when you say "that workflow", do you mean the practice of prefixing branches, or are you referring a larger workflow related to the prefixing?

drapergeek:
@iwz in this case I'm referring to prefixing branches. Removing this guideline doesn't just mean that people have the option to stop using the prefix, it means that we will all stop using eventually as the "tradition" falls away. I don't like that being the case.

I like the features provided by having the prefix. We don't have a "nice things you can do" section so in my opinion leaving it in the guides makes sense.

iwz:
@drapergeek gotcha. I'm on the other side. I don't see the "tradition" as a valuable one, or one that should be kept. I think that the problem the prefix is trying to solve (identifying who is working on a branch) is better solved by other tools, such as GItHub's Branches page, and the git history. I also am concerned that prefixing branches with one's name leads to a stronger feeling of individual code ownership, which I think is a bad thing. Of course, following @jferris's technique of creating a new branch with your own initials could alleviate that a bit, but it seems like more work.

Are there other features provided by having the initials prefix other than helping to identify who is working on a branch?

jferris:
I use the initials primarily to document who should force push to a branch, but that's realistically not too useful. Common sense largely dictates whether or not it's okay to force push to a feature branch, and given Git's distributed nature and requirements that `push` and `pull` are fast forward by default, it's very unlikely somebody would lose their work.

mike-burns:
What's the deal with branch prefix requirements?

jferris:
> What's the deal with branch prefix requirements?

I have created a form to settle this discussion: https://docs.google.com/a/thoughtbot.com/forms/d/1ApuDUcsuFse4icT3SSiBD7QxjodEJcx2aeiBjBaDb9U/viewform?usp=send_form

jferris:
There doesn't seem to be a strong interest in removing this guideline. I'm going to leave this open for a while longer, but if a stronger majority doesn't emerge, I suggest we close this without merging.

nickrivadeneira:
Could we just use "Prefer", as in "Prefer prefixing feature branch names with your initials", like we do on other guidelines that are suggested but optional?

The pros and cons as I see it:
##### Pro:
- Easy to see who is working on branches at a glance
- When people are pairing on a feature but have not set up a paired commit author, that pairing is only evident in the branch name
- If people have multiple branches open, they are grouped together alphabetically in the branch list
- The alphabetical grouping by initials makes it clear if someone needs to tidy up their stale branches

##### Con:
- Have to write 3 extra characters...

Needless to say I'm in the "keep the guideline" camp, but with a "Prefer" modification.

jferris:
Here are the current results from the poll:



There isn't a clear majority, but I think the numbers are strong enough from the combined groups of "I don't want initials" and "I don't think we need a guideline" that the correct thing to do would be remove the guideline and decide it on a project-by-project basis.

PR #313, opened 2015-02-23
Introduce security guide

jferris:
I'm hoping to start a discussion about our security practices. I'm also hoping to set a baseline of what we consider the minimal security we accept as a team, and to raise that baseline over time.

### DIFF
--- a/security/README.md
+++ b/security/README.md
@@ -1,0 +1,57 @@
+Security
+========
+
+A guide for practicing safe web.
+
+Think
+-----
+
+Security is important, and you can't practice these guidelines without
+understanding them. Make sure you understand each guideline, why it exists, and
+how to follow it.
+
+Failing to follow these guidelines will likely put you, your team, and your
+deployed services at risk of compromise or loss of privacy.
+
+Using Passwords
+---------------
+
+* Use a unique password for every account you create.
+* Use a tool like [pwgen](https://github.com/jbernard/pwgen) or
+  [1password](https://agilebits.com/onepassword) to generate random passwords.
+* Use a tool like GnuPG to encrypt passwords if you need to share them with
+  somebody.
+
+Accepting Passwords
+-------------------
+
+* Don't store passwords in plain text.
+* Don't hash passwords using a reversible cipher.
+* Don't hash passwords using a broken cipher, such as MD5 or SHA1.
+* Don't accept passwords or session tokens over HTTP.
+
+HTTPS
+-----
+
+* Use HTTPS for all web traffic.
+* Use HTTPS in the beginning; it's harder to introduce later.
+* Use HTTPS redirects for HTTP traffic.
+* Use [HSTS](http://tools.ietf.org/html/rfc6797) headers to enforce HTTPS
+  traffic.
+* Use secure cookies.
+* Avoid protocol-relative URLs.
+
+PGP and GnuPG
+-------------
+
+* Use a PGP signature in an email if you want somebody to trust that you wrote
+  it.
+* Use PGP to check email signatures if you want to know who wrote it.
+* Use PGP to encrypt emails if you want to be sure nobody but the recipient is
+  reading it.
+* Use ultimate trust for your own keys.
+* Use full trust for keys you have verified in person or via a secure video
+  chat.
+* Don't share your private key with anyone, including services like Keybase.
+* Keep at least one backup of your private key and revocation certificate in a
+  secure location, such as a thumb drive.


jferris:
cc @thoughtbot/security

security/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Security is important, and you can't practice these guidelines without
+understanding them. Question each guideline before following it yourself and
+start a discussion whenever you're in doubt.
+
+Passwords
+---------
+
+* Use a unique password for every account you create.
+* Don't store passwords in plain text.
```
mike-burns: The first bullet point seems like a guide for a user, while the second is for a developer who is building software. I'd find this easier to follow if this guide was split by audience, if so.

mike-burns: More than not storing it in plain text: do not hash it using a reversible cipher, do not encrypt it using a broken cipher (e.g. SHA1).

mike-burns:
I'm definitely interested in contributing to a guide on avoiding rookie security blunders for developers. I think the HTTP section is a great start there.

I'm less interested in having guides around how to be a more secure user, except in one place: 1Password. I want to trust that my coworkers know how to treat the shared 1Password database.

Some things that might be interesting to mention:

- Timing attacks, especially around `==`. Maybe: "Think twice when performing an equality comparison (e.g. `a == b`). Does the comparison leak data; for example, does it take longer to produce `true` than to produce `false`?"
- A list of people to read. My list (off the top of my head): SwiftOnSecurity, __apf__, bcrypt, hacks4pancakes, fugueish, thegrugq, moxie, Bruce Schneier, Ted Unangst.

BlakeWilliams:
Might be a bit early for it, but thoughts on adding encryption to your drives? I think it would make sense to  suggest encrypting at least your home folder.

OS X has Filevault which is easy enough, and  Linux has [eCryptfs](https://wiki.archlinux.org/index.php/ECryptfs#Encrypting_a_home_directory) which can be setup to be pretty transparent since it decrypts on login.

mike-burns:
I was just reminded by Blake's comment that I have a blog post I started in January, 2014 about this. Here it is, as I left it unfinished:

> password-protect BIOS
> disable booting off external media
> Prey
> separate PGP subkeys + expiration
> lock screen + autolock
> 2FA
> 1password
> encrypt FS
> 
> 
> You gotta lock that down
> 
> Theft happens. It's part of life, and it's definitely part of business.
> Laptop theft can cost you more than just a new Thinkpad and the time
> spent running the laptop script; there's also the source code,
> passwords, and other private material that you store locally.
> 
> Let's make it harder for the thief to get at that.
> 
> Track it
> 
> There are various tools for tracking your missing laptop. Prey is an
> open source device tracker for the truly paranoid. Once a device is
> marked as missing using their Web site, your laptop sends its location,
> network details (including nearby wifi SSIDs), a screenshot, and a
> picture from the webcam.
> 
> Apple's Find My Mac is a similar concept, though limited to Apple
> devices and you cannot verify the source code.
> 
> Both of these services depend on your laptop connecting to the Internet.
> For this to be most effective, make sure your operating system stays in
> tact.
> 
> Don't let them wipe your OS
> 
> If they install over your operating system, they can sell it. More
> importantly for them - and for you - if they install over your OS they
> have more of a chance of defeating your device tracking. (Find My Mac
> stays in the boot loader and cannot be removed, but if you think that
> means it's infallible then I have a bridge I'd like to sell you too.)
> 
> Make it harder for the thief: disable booting off external media,
> password-protect your firmware, and give them a password-free user
> account.
> 
> When the thief wants to reinstall the OS, they need to boot off external
> media (USB, CD, PXE, etc.). Disable this from your firmware options.
> 
> However, to turn this back on one merely needs to open the firmware
> settings. Therefore, put a password on your firmware settings.
> 
> The best thing the thief

security/README.md
sikachu: +1 on the first comment on @mike-burns. This is confusing.

If this is for dev, I'd revise the word on 2nd bullet to say something like "don't store it in a plaintext file on your machine" or something.

If that's the case, you could also adds that you should only send password to other user via end-to-end encrypted channel (WhatsApp, iMessage, Telegram, PGP, ...)

derekprior: both of these strike me as things end users and developers should follow. As an end user I should have unique passwords not stored in plain text. As a developer I should have unique passwords not stored in plaintext...

security/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+HTTPS
+-----
+
+* Use HTTPS for all web traffic.
+* Use HTTPS in the beginning; it's harder to introduce later.
+* Use HTTPS redirects for HTTP traffic.
+* Use HSTS headers to enforce HTTPS traffic.
```
derekprior: I think this will be new for a lot of people, so a link would seem useful. Maybe http://tools.ietf.org/html/rfc6797

security/README.md
gabebw: I think these should be dev-focused, so I'd kill the first bullet.

jferris: Expanded and separated in 4c4cf4a.

security/README.md
jferris: Linked in f1ab982.

security/README.md
jferris: > I think these should be dev-focused, so I'd kill the first bullet.

I think it's really important that developers practice good personal security. Otherwise, attackers can use a developer's carelessness to circumvent good security on deployed services.

I'd like to have this guide contain enough personal security recommendations that developers don't compromise the applications that they're deploying. Basic password recommendations seem like an easy place to start.

jferris:
@mike-burns after this is merged, I think it would be great to add things like timing attacks and other potential issues we should generally be aware of.

@BlakeWilliams I think hard drive encryption is probably a good idea, but it's not something we do right now. I think we'd need to make a concerted effort to get everybody to convert their drive to be encrypted (probably with blog posts about how to achieve this). At that point, I think we could add it to our guidelines.

jferris:
@sikachu added a bit about sharing passwords in 3cdcdbf.

mike-burns:
My only complaint with the first section is that it is best practices -- if you don't follow the advice you're making a terrible mistake, but it's your own choice. The subsequent sections are more mandatory; they affect others, and it is something that we can all help each other with.

That said: LGTM.

jferris:
> My only complaint with the first section is that it is best practices -- if you don't follow the advice you're making a terrible mistake, but it's your own choice. The subsequent sections are more mandatory; they affect others, and it is something that we can all help each other with.

Is there a way you'd rephrase the beginning? Think I should drop that section entirely, at least for now?

mike-burns:
How about moving the first section to `/best-practices/README.md` instead?

security/README.md
```diff
@@ -0,0 +1,53 @@
...
+  [1password](https://agilebits.com/onepassword) to generate random passwords.
+* Use a tool like GnuPG to encrypt passwords if you need to share them with
+  somebody.
+
+Accepting Passwords
+-------------------
+
+* Don't store passwords in plain text.
+* Don't hash passwords using a reversible cipher.
+* Don't hash passwords using a broken cipher, such as MD5 or SHA1.
```
derekprior: Can we sum all these up (and other related items we're not listing here) with: "Use a vetted, tested library for password authentication" ?

derekprior: other things I'd add:

* Don't transmit passwords or session tokens over HTTP.

jferris: > Can we sum all these up (and other related items we're not listing here) with: "Use a vetted, tested library for password authentication" ?

Maybe; my instinct is that it's useful to know what's important to look for in a password implementation, but I may be off.

> Don't transmit passwords or session tokens over HTTP.

Think this is redundant with the HTTPS section below?

derekprior: It seems important enough to call out on its own to me because we're likely to still work with sites available via HTTP. You can have a secure website that uses HTTP but you can't have secure password/session token authentication with HTTP.

jferris: Okay, I added it in cebac3c. We'll see if anybody complains about the overlap.

jferris:
> How about moving the first section to `/best-practices/README.md` instead?

We already have a similar section at the top of that file. I'll try rewording the Security heading.

jferris:
@mike-burns updated the intro for your review

security/README.md
```diff
@@ -0,0 +1,57 @@
...
+-----
+
+Security is important, and you can't practice these guidelines without
+understanding them. Make sure you understand each guideline, why it exists, and
+how to follow it.
+
+Failing to follow these guidelines will likely put you, your team, and your
+deployed services at risk of compromise or loss of privacy.
+
+Using Passwords
```
mike-burns: Oh really sorry: I meant that this section is more of a set of best practices, whereas the others are basically mandatory.

My wording ("that first section") was confusing and vague and I'm really sorry about that.

derekprior:
Do we have a recommendation for running a rails app locally via HTTPS? Should we be running via HTTPS in dev mode?

security/README.md
jferris: Ah, I see.

I think I see this section as also basically mandatory. If you don't follow any of these guidelines, your accounts may become compromised. As your team probably has granted you access to privileged data, that means your team will suffer from your lack of personal security.

Basically, as a member of a developer team, you are given the keys to secure services and asked to keep those keys safe. If you reuse passwords, you're exposing the whole team to risk.

jferris:
> Do we have a recommendation for running a rails app locally via HTTPS? Should we be running via HTTPS in dev mode?

I *think* this is unnecessary. You can do this reasonably easily with nginx or puma, but it doesn't seem valuable to encrypt transmissions on a local machine. There's no way to snoop traffic on a loopback device.

derekprior:
Right. I was approaching it from the perspective of eliminating dev/prod differences. For instance, the recommended configuration of clearance includes setting secure cookie to false if you are in dev or test. So if there's a problem that is only present when secure cookie is on, you won't see it until you go to staging/production.

A better example might be protocol relative URLs. There are times when these will be the best option for integrating with other local services, for example.

security/README.md
mike-burns: OK, your explanation does match with my desire: just enough personal requirements to keep the team/organization safe. I no longer have an issue with this section in this README.

jferris:
>  I was approaching it from the perspective of eliminating dev/prod differences.

I think it would be cool to investigate this in the interest of improving dev/prod parity. Until we have a commonly working setup, though, I feel like we can't have a guideline about it.

security/README.md
tute: I like the three different bullets rather than a condensed one, makes you think of what could leak info which is a good exercise.

We should probably mention authentication tokens (any others?) should be treated the same way.

security/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+* Use a PGP signature in an email if you want somebody to trust that you wrote
+  it.
+* Use PGP to check email signatures if you want to know who wrote it.
+* Use PGP to encrypt emails if you want to be sure nobody but the recipient is
+  reading it.
+* Use ultimate trust for your own keys.
+* Use full trust for keys you have verified in person or via a secure video
+  chat.
+* Don't share your private key with anyone.
```
tute: I'd add "not even 'secure' services". There's been two legit organizations that ask to store your private keys (can dig which they were). It's "obvious", and these guides should state it too I think.

security/README.md
```diff
@@ -0,0 +1,57 @@
...
+* Keep at least one backup of your private key and revocation certificate in a
+  secure location.
```
tute: What are a secure locations?

tute:
@calebthompson might be interested in this thread.

tute:
Can we advice guidelines, without saying we all need to follow them? I think "you should encrypt your hard drive" would be nice to add.

security/README.md
mike-burns: Keybase was one org that wanted to store your private key.

mike-burns:
This is a good start. Once it's merged we can all pile on with more PRs.

security/README.md
jferris: Added an example in 417e4a5.

security/README.md
jferris: Updated.

jferris:
> Can we advice guidelines, without saying we all need to follow them? I think "you should encrypt your hard drive" would be nice to add.

This thread is sort of long, so I know it's easy to miss, but this was [discussed already](https://github.com/thoughtbot/guides/pull/313#issuecomment-75579905). Please read my [previous response](https://github.com/thoughtbot/guides/pull/313#issuecomment-75607360).

tute:
> This thread is sort of long, so I know it's easy to miss, but this was discussed already. Please read my previous response.

Got it. I did see it, but I thought of it as a nice to have. Agree that after a blog post it's easier for everyone.

PR #315, opened 2015-02-24
Revert hex value capitalization guideline

kaishin:
Lowercase hex colors proved to be a pain more than I had expected. Most tools we use to get color values default to uppercase, and the work involved in the process of transforming these values just for the sake of cross-project consistency seems disproportional to me.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -8,12 +8,12 @@
 * Use hyphens when naming mixins, extends, classes, functions & variables: `span-columns` not `span_columns` or `spanColumns`.
 * Use one space between property and value: `width: 20px` not `width:20px`.
 * Use a blank line above a selector that has styles.
-* Prefer hex color codes `#fff`.
+* Prefer hex color codes `#fff` or `#FFF`.
 * Avoid using shorthand properties for only one value: `background-color: #ff0000;`, not `background: #ff0000;`
 * Use `//` for comment blocks not `/* */`.
 * Use one space between selector and `{`.
 * Use double quotation marks.
-* Use only lowercase, including colors.
+* Use only lowercase, except for hex or string values.
 * Don't add a unit specification after `0` values, unless required by a mixin.
 * Use a leading zero in decimal numbers: `0.5` not `.5`
 * Use space around operands: `$variable * 1.5`, not `$variable*1.5`


Magnus-G:
:+1: I totally agree. Copying hex codes from chrome always gives uppercase and I think we should stick to that. Why did we decide on lowercase in the first place?

kaishin:
@Magnus-G I couldn't find the relevant PR. @kylefiedler probably knows more.

jeffpersonified:
Completely agree.

kylefiedler:
@kaishin @Magnus-G Chrome hex codes are lowercase (
screen shot 2015-02-24 at 11 19 45 am
), so are colors from photoshop. Most other color grab tools I've used have a default for lowercase or allow you to switch. I'm not sure what tools you are all using that don't let you change the format.

If you look closely Google (https://google-styleguide.googlecode.com/svn/trunk/htmlcssguide.xml#Hexadecimal_Notation) and GitHub (https://github.com/styleguide/css) both use lowercase too. 

I personally like lowercase because it's more readable. I can easily discern letters from numbers. The all uppercase reminds me of the days where HTML was uppercase as well and makes me cringe. 

I only have strong opinions to argue with @kaishin, I honestly don't think this needs to be declared in the guides though it's super nitty gritty.

tysongach:
:+1:

I still prefer lowercase and will keep doing that. Inspector and Sip, the color picker I use, output lowercase values. There is no real benefit having this strict of a guide for this, let’s loosen it up a bit; just try to be consistent within a project.

Magnus-G:
@kylefiedler is that a setting then? Chrome has uppercase here:

image

teoljungberg:
Just to weigh in here as someone who hasn't configured Chrome at all

teoljungberg:
There's actually a setting in Chrome that does this, if you force the Color format to be "HEX" it's uppercased. But if you use the default it will use it as it's authored in the CSS

screenshot 2015-02-25 10 12 37

kaishin:
For Chrome, you need to actually bring up the color picker. It's always uppercase there.
screenshot-2015-03-02-1

jferris:
Any developing preferences here? @thoughtbot/design

smharley:
I agree with being consistent within a project

kaishin:
I will go ahead and merge this.

PR #317, opened 2015-02-26
Ember Testing: Prefer `findWithAssert` over `find`

seanpdoyle:
`findWithAssert` will raise an meaningful exception, whereas 
`find` is just a scoped version of `$()`, which will return an empty jQuery enumerable.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -299,6 +299,10 @@ Ember
 * Prefer dependency injection through Ember initializers over globals on window
   or namespaces.
 
+Testing
+
+* Prefer `findWithAssert` over `find`
+
 Angular
 -------
 


drapergeek:
How about adding a note that explains this is only effective when you know the element will be there. For example, I often have specs that expect something to not be visible, in those cases using `findWithAssert` will not work.

Overall I'm :+1: on this, I just want to make it clear that in some cases you can't use `findWithAssert`.

seanpdoyle:
@drapergeek that's a good point. 

How would you word that?

> Prefer `findWithAssert` over `find` when asserting the presence of an element.

?

drapergeek:
`Prefer findWithAssert over find when looking for an element you expect to exist.` ?

seanpdoyle:
`looking` -> ...`fetching`?

`looking` sounds so ... unscientific

drapergeek:
That sounds good to me.

nickrivadeneira:
:+1:

Issue #308, opened 2015-02-17
Readme URL not resolving on LN 17


ajkamel:
Readme for url "Exceptions should be exceptional" does not resolve.   

https://github.com/thoughtbot/guides/blame/master/best-practices/README.md#L17

jferris:
Fixed in c95ba094b5.

PR #319, opened 2015-03-05
Fix exceptions should be exceptional url

ajkamel:
I raised an Issue #308 hoping you may have had the correct one, instead of removing it I found a relevant blog post from Thoughtbot that may address this point. This is addressing c95ba09 @jferris


jferris:
Thanks for the pull request. I think I'd prefer to leave this one without a URL, as doing a Google search for "exceptions should be exceptional" (or asking a team mate) seems like the best way to figure out what this means.

The canonical resource on this is the Pragmatic Programmer, but only highlights of it are online, so we can't link to the actual chapter.

PR #321, opened 2015-03-09
Ruby: Handle case for long method definitions

jakecraige:
Generally I think we want to prefer method definitions under 80 chars, but
there may be times where it "has" to be so I think we need how to handle that
documented.

This style was chosen to match how we handle long argument lists
[here](https://github.com/thoughtbot/guides/tree/master/style#formatting)

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -40,7 +40,8 @@ Formatting
 * Don't misspell.
 * Don't vertically align tokens on consecutive lines.
 * If you break up an argument list, keep the arguments on their own lines and
-  closing parenthesis on its own line.
+  closing parenthesis on its own line. [Example 1][multi-line arguments]
+  [Example 2][long method signature]
 * If you break up a hash, keep the elements on their own lines and closing curly
   brace on its own line.
 * Indent continued lines two spaces.
@@ -57,6 +58,9 @@ Formatting
 * Use [Unix-style line endings] (`\n`).
 * Use [uppercase for SQL key words and lowercase for SQL identifiers].
 
+
+[multi-line arguments]: /style/ruby/sample.rb#L69
+[long method signature]: /style/ruby/sample.rb#L16
 [dot guideline example]: /style/ruby/sample.rb#L11
 [uppercase for SQL key words and lowercase for SQL identifiers]: http://www.postgresql.org/docs/9.2/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
 [Unix-style line endings]: http://unix.stackexchange.com/questions/23903/should-i-end-my-text-script-files-with-a-newline
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -39,5 +39,5 @@ Ruby
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
 
-[trailing comma example]: /style/ruby/sample.rb#L49
-[required kwargs]: /style/ruby/sample.rb#L16
+[trailing comma example]: /style/ruby/sample.rb#L57
+[required kwargs]: /style/ruby/sample.rb#L24
--- a/style/ruby/sample.rb
+++ b/style/ruby/sample.rb
@@ -13,6 +13,14 @@ class SomeClass
       each_method_lives_on_its_own_line
   end
 
+  def method_with_long_signature(
+    the_first_arg:,
+    the_second_one:,
+    and_another_one:
+  )
+    # code here
+  end
+
   def method_with_required_keyword_arguments(one:, two:)
   end
 


style/ruby/sample.rb
```diff
@@ -13,6 +13,14 @@ def method_with_arguments(argument_one, argument_two)
       each_method_lives_on_its_own_line
   end
 
+  def method_with_args_list_over_80_chars(
+    the_first_arg:,
+    the_second_one:,
+    and_another_one:
```
jessieay: trailing comma :tangerine:

style/ruby/README.md
```diff
@@ -38,6 +38,9 @@ Ruby
 * Use heredocs for multi-line strings.
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
+* Break up method definitions over 80 characters long by
```
jessieay: I wonder if this can just be extrapolated based on `Use a trailing comma after each item in a multi-line list, including the last item.` aka - it seems similar enough that you might just know what to do in this case based on the general rule...but I am not 100% sure. interested in what others think

mcmire: I don't think it can be extrapolated from that, however, I think this wording could be improved. What about:

> When defining methods whose signature exceeds 80 characters, list arguments on separate lines.

jessieay: yeah sorry that was not the rule I meant to highlight (copy-pasted too quickly). These are the related rules I was thinking of :

`Break long lines after 80 characters.`
 `If you break up an argument list, keep the arguments on their own lines and closing parenthesis on its own line.` 

from https://github.com/thoughtbot/guides/blob/803a13409e72ceb7e42c6a887922e359a0fd8da5/style/README.md

mcmire: Ah... yes, those two rules already handle this. Maybe we just add this example to the second rule? 

`If you break up an argument list, keep the arguments on their own lines and closing parenthesis on its own line. [example]`

jakecraige: I think it could be. 

Having the example seems valuable so that this case is covered clearly, but we may not need to actually list it in the list of styles here. 

This PR comes from me not knowing how we handle this, and then getting two differing answers in Slack, so it seems we do need to define this somehow.

jakecraige: I've updated the PR to have the the breaking up argument list section to link to the two examples of it.

style/ruby/sample.rb
jakecraige: If only... but it causes a syntax error here so no can do

style/README.md
```diff
@@ -40,7 +40,9 @@ Formatting
 * Don't misspell.
 * Don't vertically align tokens on consecutive lines.
 * If you break up an argument list, keep the arguments on their own lines and
-  closing parenthesis on its own line.
+  closing parenthesis on its own line. [Example 1][multi-line arguments]
+  [Example 2][long method signature]
+
```
mcmire: I don't think you need this blank line, it causes a paragraph tag to be added to this li.

mcmire:
One little tweak but otherwise looks good to me :+1:

jferris:
This looks good to me.

Just a reminder, since I don't think you've submitted a pull request to the guides before: our [guidelines](https://github.com/thoughtbot/guides/blob/master/CONTRIBUTING.md) ask that we leave pull requests open for about a week to allow everyone on the team a chance to review changes to this repository.

style/README.md
jakecraige: Thanks. fixed it

jakecraige:
@jferris Sounds good. Thanks for the heads up

sgrif:
Out of curiosity, isn't this already required by the "trailing comma" guideline?

mcmire:
@sgrif I don't think so... although I guess for consistency there should be a trailing comma after the last argument in the example. This is saying if you have a long argument list in a method signature break it up over multiple lines.

jessieay:
@sgrif we aren't using trailing commas for args (yet). I predict we will start soon since it's hard to use them for arrays and hashes but not args, in my opinion. A list is a list! :link:

jkrmr:
Also out of curiosity: What's the rationale behind the trailing comma (after the last element in a list, that is)?

Given that [it's arguably best avoided in JavaScript][airbnb-on-js-commas] (can break older browsers), and [with certainty can't be applied uniformly in Ruby as of 2.1][breaks-required-keyword-list] (produces a syntax error on argument lists composed of required keywords), I'm surprised this a guideline. 

Is the idea solely, as I suspect, to guard against the possibility of someone appending to the list but forgetting to add a comma? Or is there another reason that I'm missing?

[airbnb-on-js-commas]: https://github.com/airbnb/javascript#commas
[breaks-required-keyword-list]: https://github.com/thoughtbot/guides/pull/321#discussion_r26088816

derekprior:
@jkrmr : https://github.com/thoughtbot/guides/pull/167

jkrmr:
@derekprior Much obliged! I searched for that but must have overlooked it. 
Had not considered the implications for diff noise.

PR #324, opened 2015-04-07
Prefer dependency injection through `Ember.inject`

seanpdoyle:
[Ember 1.10.0](ember) adds:

* `Ember.inject.service()`
* `Ember.inject.controller()`

[ember]: http://emberjs.com/blog/2015/02/07/ember-1-10-0-released.html#toc_injected-properties

[Ember-Data](ember-data) supports injecting the `store` as a service.

This makes injecting services via initializers a bad idea.

[ember-data]: http://emberjs.com/blog/2015/03/23/ember-data-1-0-beta-16-released.html#toc_store-as-a-service

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -299,7 +299,7 @@ Ember
 * Don't use jQuery outside of views and components.
 * Prefer to use predefined `Ember.computed.*` functions when possible.
 * Use `href="#"` for links that have an action.
-* Prefer dependency injection through Ember initializers over globals on window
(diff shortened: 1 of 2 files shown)


BlakeWilliams:
:+1: from me. Any chance we can get an example of this?

MattMSumner:
:+1:

seanpdoyle:
I've added sample code.

This is ready for another look

drapergeek:
:shipit:

PR #328, opened 2015-04-17
Specify sass order for nested elements

conchan:
Since nested element selectors can be so broad for what they touch, they should sit above any nested classes.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -28,8 +28,9 @@
 * Place `@extends` and `@includes` at the top of your declaration list.
 * Place media queries directly after the declaration list.
 * Place concatenated selectors second.
-* Place pseudo states and elements third.
-* Place nested selectors last.
+* Place pseudo-states and pseudo-elements third.
+* Place nested elements fourth.
+* Place nested classes fifth.
 
 ## Selectors
 
--- a/style/sass/sample.scss
+++ b/style/sass/sample.scss
@@ -25,6 +25,10 @@ section.application {
     content: ">";
   }
 
+  h1 {
+    attribute: value;
+  }
+
   .news {
     attribute: value;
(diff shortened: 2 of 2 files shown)


style/sass/README.md
```diff
@@ -28,8 +28,9 @@
 * Place `@extends` and `@includes` at the top of your declaration list.
 * Place media queries directly after the declaration list.
 * Place concatenated selectors second.
-* Place pseudo states and elements third.
-* Place nested selectors last.
+* Place pseudo states and pseudo elements third.
```
kylefiedler: do these need a `-` between pseudo and states/elements? eg `pseudo-states`

kylefiedler:
Small comment otherwise looks good to me. @thoughtbot/design?

smharley:
:+1: i like it

donokuda:
:+1: Go for it @conchan!

wardpenney:
I almost never nest elements (except for void elements like IMG).

conchan:
Sure, but this is for those who do :wink:

wardpenney:
Yep!

ryancoughlin:
This is a nice change, nice value. No reason to make a class for an `img` if its not repeated as pattern. Nice @conchan - easier to scan as well having it promoted to the top.

wardpenney:
@ryancoughlin I meant that nesting elements is a rare use case for me, and I believe encouraging it by enshrining it in the syntax guide perpetuates a bad practice, IMO. But @conchan was saying some people do do that, which is understandable. It's just one of those things.

ryancoughlin:
@wardpenney I agree. I try not to when I dont have to, but for things like `img`, etc. Its nice to have that nesting ability. Happy balance.

PR #330, opened 2015-04-29
Correct order of android:layout_width and android:layout_height

aenain:
Hey guys, those two were reversed in comparison to the order from https://github.com/thoughtbot/guides/blob/master/style/android/README.md.


sikachu:
Good catch! Thanks for your patch.

PR #332, opened 2015-05-07
Avoid `class A::B` definitions

teoljungberg:
`ctags` is an extremely useful tool for navigating quickly to a class,
module, or method definition. Sadly it has a few limitations, one of
them is how namespaced classes/modules are defined.

If we define a class `B` like so: `class A::B; end`, `B` will never be
found nor indexed by `ctags`.

By defining classes like:

```ruby
module A
  class B
  # body omitted
  end
end
```

We can use `ctags` to navigate to `B` (and `A` for that matter aswell`)

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -13,6 +13,8 @@ Ruby
 * Avoid bang (!) method names. Prefer descriptive names.
 * Don't use `self` explicitly anywhere except class methods (`def self.method`)
   and assignments (`self.attribute =`).
+* Prefer nested class and module definitions over the shorthand version
+  [Example][class definition example]
 * Prefer `detect` over `find`.
 * Prefer `select` over `find_all`.
 * Prefer `map` over `collect`.
@@ -42,3 +44,4 @@ Ruby
 
 [trailing comma example]: /style/ruby/sample.rb#L57
 [required kwargs]: /style/ruby/sample.rb#L24
+[class definition example]: /style/ruby/sample.rb#L121
--- a/style/ruby/sample.rb
+++ b/style/ruby/sample.rb
@@ -117,3 +117,8 @@ class SomeClass
     part_one? && part_two?
   end
 end
+
+module A
+  class B
+  end
+end


derekprior:
I'm torn on this. Does `ctags` pick up nested classes like so?

```ruby
class A
  class B
  # body omitted
  end
end
```

or does it always have to be a class within a module? There are times when Ruby won't let you do this without nesting (or re-opening) a class. For instance, if you have a top level `Comment` class and want a class in that namespace you have to nest it inside a class definition. You can't declare a class and then declare that class as a module elsewhere.

Are there any other advantages at all to explicitly declaring the module like that? I love `ctags` but I'm not sure I'm willing to make structural concessions to support it.

teoljungberg:
> I'm torn on this. Does ctags pick up nested classes like so?

All nested classes and modules will be picked up.

I don't really follow your example @derekprior, could you clarify?

calleerlandsson:
How we define classes and modules also affect constant lookup:

```ruby
module A
  CONSTANT = "a constant"
end

module A::B
  def self.test
     CONSTANT
  end
end

module A
  module C
    def self.test
      CONSTANT
    end
  end
end

A::B.test #=> NameError: uninitialized constant A::B::CONSTANT
A::C.test #=> "a constant"
```

teoljungberg:
good example @calleerlandsson, that one has bit me once or twice

derekprior:
> All nested classes and modules will be picked up.

Okay, thats cool. I was just saying that sometimes it's necessary to nest a class inside of another class (rather than a module) to get the name you want (i.e. your top level is already a class rather than a module). I mistook your PR description for saying I could *only* nest into modules.

@calleerlandsson interesting, thanks.

:+1:

JoelQ:
This seems reasonable. I knew this was a general best practice due to the constant lookup issues mentioned by @calleerlandsson but wasn't aware this also made things easier for CTags.

:+1:

calleerlandsson:
Would the guideline be clearer if it was more than an example? Maybe something like "Avoid defining classes and modules using shorthand nesting (`class A::B`)"?

teoljungberg:
@calleerlandsson Agreed, updated in 5779588

style/ruby/README.md
```diff
@@ -11,6 +11,8 @@ Ruby
 * Avoid explicit return statements.
 * Avoid using semicolons.
 * Avoid bang (!) method names. Prefer descriptive names.
+* Avoid defining classes and modules using shorthand nesting (class A::B)
```
gylaz: Should we flip this wording to match the example?
Something like: `Prefer nested module definitions instead of the shorthand version`

gylaz:
:+1:

style/ruby/README.md
teoljungberg: I sort-off want to keep `Avoid` in there to emphesize that we couldn't use `A::B` unless it's actually needed

drapergeek:
:+1:

gabebw:
Sure. I've been split 50/50 in the past, basically however I feel that day. The `A::B.test` example is convincing. :+1:

mcmire:
All the times I've nested modules (and granted I haven't done it that much) have been for convenience. There's no real *need* to do this though. This seems like a good guideline to me.

teoljungberg:
This has been open for more than a week now, and there doesn't seem to be a strong argument against this. Does anyone else have any objections before I squash and merge?

style/ruby/README.md
```diff
@@ -13,6 +13,8 @@ Ruby
 * Avoid bang (!) method names. Prefer descriptive names.
 * Don't use `self` explicitly anywhere except class methods (`def self.method`)
   and assignments (`self.attribute =`).
+* Prefer nested class and module definitions instead of the shorthand version
```
calleerlandsson: Should "instead of" be "over" or "to"?

teoljungberg: Good call, fixed in 7b95965

TheLonelyGhost:
:+1: for merging the guideline. I usually nest the modules instead of the single-line definition, at least until load order begins to matter. Then it can be used as an early warning system with an obvious solution.

```ruby
# file a
module A
  class B
    # ...
  end
end


# file 2
module A
  class C
    # ...
  end
end


# file 3
class A::D
  # ...
end
```

In the above code, load order only matters with file 3 since it requires `module A` to be defined before it can define `class A::D`. In most circumstances, load order does not matter so long as all appropriate files are actually loaded.

Using syntax in file 3 adds an additional requirement that it be loaded after file 1 or 2, which shouldn't be necessary.

teoljungberg:
No objections made, merging

PR #334, opened 2015-06-03
Fix unix style line ending link

JSalig1:
Link was behaving as a relative link because of the `(` immediately following the `]` sending to:

https://github.com/thoughtbot/guides/blob/master/style/%60n%60

Added reference text so the link will behave as a reference-style link and also display the (`\n`).

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -55,7 +55,7 @@ Formatting
 * Use spaces around operators, except for unary operators, such as `!`.
 * Use spaces after commas, after colons and semicolons, around `{` and before
   `}`.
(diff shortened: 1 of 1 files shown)


JSalig1:
I meant to include the link to where I was pulling my definitions from for 'relative' and 'inline' links in Markdown.  Just to make the above a bit clearer:

http://daringfireball.net/projects/markdown/syntax#link

graysonwright:
Looks good to me.

sikachu:
:+1:

mike-burns:
Merged, thank you.

PR #335, opened 2015-06-16
Prefix unused variables and arguments with `_`

gylaz:


gfontenot:
Fwiw, we typically use `_foo` to indicate an internal (private) variable, not an ignored variable in Swift/ObjC.

georgebrock:
What's the benefit of following this convention?

* We aim to have short methods, if there's an argument that we need to accept but don't use, that should be clear from the method body without any special naming convention.
* It doesn't make the code clearer to someone who isn't familiar with our style guide. As Gordon mentioned, and `_` prefix is often used to mean "private" rather than "unused" (not just in iOS, but also in JavaScript and Python)
* It can only be applied to positional arguments, not keyword arguments, which means we can't apply it consistently.

gylaz:
FWIW, this is in the Ruby specific section.

One situation I can think of:
Often, in specs folks create fixture data that isn't referenced in the test. The data is stored into a well-named clarity. Many of us use Syntastic in vim, and it will highlight an unused variable.

I'm not saying ^ is a good reason, but it does provide additional context.

georgebrock:
> FWIW, this is in the Ruby specific section.

Sorry, I wasn't clear: what I meant was, I'm worried about taking a naming convention that's already going to be familiar to developers who know Swift/JavaScript/Python as meaning one thing, and using it to mean something different in Ruby.

mcmire:
I have always thought that what Hound (or, rather, Rubocop) recommended to resolve an unused variable was a little unconventional. That said, I haven't had any issues with it.

A case in which I think it's useful is when overriding a method that takes an argument, but you never use that argument. The underscore, I think, is a good way of communicating that the variable is actually never used.

jferris:
Rubocop and Syntastic recommend this because it's what Ruby recommends. Running Ruby with warnings will complain about unreferenced variables, but not if you prefix with an underscore, because that's a Ruby convention.

JoelQ:
Haskell uses an underscore to denote unused arguments when pattern matching. It's pretty common to use a plain underscore to denote an unused block argument in Ruby.

georgebrock:
> Running Ruby with warnings will complain about unreferenced variables, but not if you prefix with an underscore, because that's a Ruby convention.

I didn't know this, and before Hound started recommending it I'd never seen it in practice. Since it's the convention the interpreter wants us to follow, I'm fine with adding it to the guide.

gabebw:
> I didn't know this, and before Hound started recommending it I'd never seen it in practice. 

I'm fine with it, but, yes, I had never seen this before Hound recommended it.

jyurek:
Underscore in Ruby has always meant that "this variable doesn't matter". I honestly don't care what a `_` prefix means in other languages, because we're not using those languages here.

That said, the only time I've ever actually used it is in tests where I want something labelled but I also don't want to use it. This keeps syntastic (and Hound) from giving a warning.

jferris:
> It's pretty common to use a plain underscore to denote an unused block argument in Ruby.

Underscore methods are also treated specially in Ruby. You're allowed to use the same underscore name for multiple arguments:

```
> [1].each_with_index {|ignore,ignore| puts "hello" }
SyntaxError: (irb):2: duplicated argument name
[1].each_with_index {|ignore,ignore| puts "hello" }
                                   ^
	from /home/jferris/.rbenv/versions/2.2.2/bin/irb:11:in `<main>'
> [1].each_with_index {|_,_| puts "hello" }
hello
=> [1]
> [1].each_with_index {|_ignore,_ignore| puts "hello" }
hello
=> [1]
```

mcmire:
Well, I learned something new today, too.

style/ruby/README.md
```diff
@@ -26,6 +26,7 @@ Ruby
   calls.
 * Prefer `if` over `unless`.
 * Use `_` for unused block parameters.
+* Prifix unused variables, or method arguments with `_`.
```
mike-burns: No comma.

Instead of "method arguments", how about "parameters"?

And "Prefix" is misspelled.

> Prefix unused variables or parameters with underscore (`_`).

gylaz: Thanks, fixed.

thorncp:
:+1:

gylaz:
It sounds like we should merge this. Any objections @georgebrock?

georgebrock:
@gylaz No objections: after everything I've learnt in this PR discussion I'm actively in favour of merging it.

PR #336, opened 2015-06-22
Prefer single quotes for javascript

iancanderson:
See related PR in ember: https://github.com/emberjs/ember.js/pull/11142

ember-cli generates files with single quotes, so preferring single quotes for ember projects makes it easier to work with the de-facto toolchain, i.e. ember-cli.

Our javascript guide prefers double quotes, and I'm not sure how we want to deal with that. Maybe for now, we can simply recommend single quotes for ember projects, and double quotes for non-ember? Thoughts welcome https://github.com/thoughtbot/guides/tree/master/style/javascript

### DIFF
--- a/style/ember/sample.js
+++ b/style/ember/sample.js
@@ -1,18 +1,18 @@
 App.Post = DS.Model.extend({
-  author: DS.belongsTo("user"),
-  tags: DS.hasMany("tag"),
+  author: DS.belongsTo('user'),
+  tags: DS.hasMany('tag'),
 
-  createdAt: DS.attr("date"),
-  title: DS.attr("string"),
+  createdAt: DS.attr('date'),
+  title: DS.attr('string'),
 });
 
-test("checks the box", function() {
-  visit("/");
-  click(".check-box");
+test('checks the box', function() {
+  visit('/');
+  click('.check-box');
 
   andThen(function() {
-    const checkBox = find(".check-box");
+    const checkBox = find('.check-box');
 
-    ok(checkBox.prop("checked"), "box is checked");
+    ok(checkBox.prop('checked'), 'box is checked');
   });
 });
--- a/style/javascript/README.md
+++ b/style/javascript/README.md
@@ -9,7 +9,7 @@ JavaScript
 * Prefer [arrow functions] `=>`, over the `function` keyword except when
   defining classes or methods.
 * Use semicolons at the end of each statement.
-* Prefer double quotes.
+* Prefer single quotes.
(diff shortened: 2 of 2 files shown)


BlakeWilliams:
I personally prefer double quotes, but I think this makes sense and most JavaScript frameworks/libs tend to lean towards single quotes. Thoughts on making this JavaScript specific instead of Ember specific?

MattMSumner:
Also related: https://github.com/emberjs/ember.js/pull/11142

I think syncing up with the community makes sense. If it really is just a matter of preference then I think single quotes make sense to stay in line with the toolchain.

seanpdoyle:
I'm in favor of this change.

In the future, `ember-cli` might accept a generator configuration for double quotes, but I doubt it.

We've already started followed this guideline on

* https://github.com/thoughtbot/screensteps-ember
* https://github.com/MerckUX/human_health

iancanderson:
Haha. I was worried that making this a general JS preference would make it harder to get consensus.. Let's let this PR hang out for a bit and see if we get consensus around single quotes for JS in general, or not.

joshuaclayton:
I'm also in favor of this change

seanpdoyle:
Unfortunately, JS puts us in a slightly different situation than Ruby.

Some familiar arguments for Ruby strings:

* Use `'` when the string is static
* Use `"` when the string will use string interpolation
* Use `"` no matter what (thoughtbot's approach)

For JS, the same conventions don't make sense, since neither `'` nor `"` support interpolation.

Some hypothetical conventions:

* Use `'` no matter what. Concatenate with `+`. Interpolate with "`${es6Syntax}" (what this PR is proposing)
* Use `"` no matter what. Concatenate with `+`. Interpolate with "`${es6Syntax}" (our current guideline)
* Use ` no matter what. (I don't think this convention is followed anywhere)

If a project is using es6 features, an argument could (and probably should) be made to outlaw string concatenation for literals, though that's probably something that should go in a separate PR.

drapergeek:
:+1: I would love to see doubles everywhere for simplicity but the community is going a different way and doubles in JS buys us nothing.

jakecraige:
I'm :+1:. This is the general consensus held in the JS community at large(not just Ember) so I see no reason against it.

For us to use `"` in JavaScript has us going against the grain for not much gain.

mcmire:
Well... this is probably going to be hard to remember when switching back and forth between Ruby and JavaScript at first, but I guess if we make sure it's in Hound, then I'm okay with this change.

MattMSumner:
@iancanderson I think this is good the merge!

PR #337, opened 2015-06-24
Prefer not using prototype extensions on functions

MattMSumner:
Why:

* The non-prototype extension versions of `.property` and `.observes`
are easier to grok.
* Prototype extensions can be disabled. If you are an addon author this
means you have to write you addon without using prototype extensions.
* In my opinion, having a clear separation between features of Ember vs
native in JavaScript leads to less confusion.

This PR:

* Adds the guideline to `style/ember/README.md`
* Adds an example to `style/ember/sample.js`

### DIFF
--- a/style/ember/README.md
+++ b/style/ember/README.md
@@ -2,8 +2,8 @@ Ember
 =====
 
 * Don't put a space between the opening handlebars braces and the variable.
-* Prefer not using prototype extensions for `computed` or `observes`
-  ([sample][prototype extensions])
+* Use `Ember.computed`, `Ember.observes`, and `Ember.on` instead of the function
+  prototype extensions ([sample][prototype extensions])
 
 Ember-Data
 ----------
--- a/style/ember/README.md
+++ b/style/ember/README.md
@@ -2,6 +2,8 @@ Ember
 =====
 
 * Don't put a space between the opening handlebars braces and the variable.
+* Prefer not using prototype extensions for `computed` or `observes`
+  ([sample][prototype extensions])
 
 Ember-Data
 ----------
@@ -20,3 +22,4 @@ Testing
 
 [helpers]: sample.js#L10-L11
 [assertions]: sample.js#L13-L17
+[prototype extensions]: sample.js#L23-L25
--- a/style/ember/sample.js
+++ b/style/ember/sample.js
@@ -16,3 +16,11 @@ test("checks the box", function() {
     ok(checkBox.prop("checked"), "box is checked");
   });
 });
+
+Ember.Object.extend({
+  name: 'something',
+
+  title: Ember.computed('name', function() {
+    return this.get('name').capitalize();
+  }),
+});


MattMSumner:
We're doing this on springleaf and find it much easier to reason about objects when they declare at the start of a property that it is computed. Using `.property` at the end feels like a surprise as it's not immediately obvious the property isn't a plain function.

style/ember/README.md
```diff
@@ -2,6 +2,8 @@ Ember
 =====
 
 * Don't put a space between the opening handlebars braces and the variable.
+* Prefer not using prototype extensions for `computed` or `observes`
```
gabebw: "Prefer not using" -> "Avoid"? 

From our README:

> "Avoid" means don't do it unless you have good reason.

drapergeek:
:-1: I done this on my last project. There were no issues with it but I didn't see any advantages either. When using an observer on init it also becomes hard to read.

I think the prototype extensions provide a nice interface that is easy to read and I don't think having a guideline for this is beneficial.

BlakeWilliams:
:-1: I agree with @drapergeek on this one. I think if we did have a guideline I'd prefer to use Prototype extensions. I also think that for the most part, any method on an Ember component/controller that isn't prefixed with _ or in the `actions` hash can be used as a property via `get`.

BlakeWilliams:
Also, the way we're currently not using prototype extensions isn't *really* not using prototype extensions.

http://guides.emberjs.com/v1.10.0/configuring-ember/disabling-prototype-extensions/

style/ember/README.md
BlakeWilliams: Maybe the following?

> Prefer `Ember.computed`, `Ember.observes`, and `Ember.on` over their function prototype extensions.

MattMSumner:
So the guideline isn't not to use Prototype extensions but not to use them for functions. I think this is totally a style concern as the functionality is almost identical. It's purely an argument of which case is more readable. This:
```js
some property: function() {
   some code...
}.property('dependency'),
```

vs:

```js
some property: Ember.computed('dependency', function() {
   some code...
}),
```

My argument is the second case is more readable for the following reasons:
* Im declaring my intent right at the start. Is this a computed property? An Observer? Does this fire off of an event? Using the first example I have no context of when this function will be called without checking what has been tacked onto the end.
* I'm declaring my dependencies right at the start so I know what will make my computed property recalculate.

If we don't think this is the right style then we should change it to prefer the second example. However, we should ABSOLUTELY have a style guide which tells new emberions which method to prefer.

seanpdoyle:
I think this will eventually be a moot point, given the progress being made with ES7 decorators.

https://github.com/rwjblue/ember-computed-decorators#with-dependent-keys

```js
// old busted
foo: Ember.computed('bar', 'baz', function() {

}),

bar: function() {

}.property('baz'),

// new hotness
@computed('bar', 'baz')
foo(bar, baz) {
  // Do Stuff
}
```

Until then, I defer to @drapergeek 's feedback, since he just came off a project where they avoided `.property()`.

drapergeek:
I still stand by my opinion that we do not need a guideline here. If we do need a guide, I prefer the prototype extension.

@seanpdoyle is completely right that this will only be useful for a short time.

MattMSumner:
@BlakeWilliams and I have been doing it on the current project and prefer it dramatically. Considering that the new hotness is closer to the proposed style I feel like it's a good guideline to future proof your application.

MattMSumner:
From my understanding guidelines are for what we think generally is the correct approach but are subject to change and don't have to be followed in every case. There are a bunch of people in the Boston office looking to get into Ember and are frustrated with the lack of guidelines. It sounds like this we may decide to close this change, but I feel like having an opinion on some of these "irrelevant" choices is proactive in getting more of the company started with Ember.

MattMSumner:
From what I understand, the [new computed syntax] introduced in 1.12 doesn't support using the prototype extension form for object properties.

[new computed syntax]: http://emberjs.com/blog/2015/05/13/ember-1-12-released.html#toc_new-computed-syntax

The feedback I've received from the community slack channel is that everyone is moving away from using prototype extensions like `.property` and `.on`. Most documentation still has code in this form so I think it's important to have a guideline on which form to use.

If we don't add this guideline do we want to start using the "new hotness" in existing apps? https://github.com/rwjblue/ember-computed-decorators

BlakeWilliams:
I'm a fan of what we have here since we've used it and are comfortable recommending it.

I love es7 decorators but I don't think they're ready for wide spread use yet. I don't think we should recommend using them yet either because of their immaturity and our lack of experience using ember-computed-decorators. I'm not saying don't use them at all either, but we should be careful to use them on a project where it makes sense.

MattMSumner:
I don't think this has reached consensus and the new decorator syntax should remove the conversation entirely.

PR #338, opened 2015-06-24
Local Ember/DS functions

MattMSumner:
Why:

* Ember and Ember-Data will move to JavaScript modules allowing us to
only import functions we actually use.

This PR:

* Proposes creating local constants for functions on `Ember` and `DS`
* This means when these libraries are converted to modules, updating the
code is as simple as converting:
```js
const { computed } = Ember;
```
to:
```js
import { computed } from 'ember';
```

### DIFF
--- a/style/ember/README.md
+++ b/style/ember/README.md
@@ -3,6 +3,10 @@ Ember
 
 * Don't put a space between the opening handlebars braces and the variable.
 * Prefer components over partials.
+* Alias local variables to functions from `Ember` and `DS`
+([sample][local-Ember-DS])
+
+[local-Ember-DS]: sample.js#L23-L24
 
 Ember-Data
 ----------
--- a/style/ember/sample.js
+++ b/style/ember/sample.js
@@ -16,3 +16,14 @@ test('checks the box', function() {
     ok(checkBox.prop('checked'), 'box is checked');
   });
 });
+
+import Ember from 'ember';
+import DS from 'ember-data';
+
+const { computed } = Ember;
+const { attr, hasMany, belongsTo } = DS;
+
+export default DS.Model.extend({
+  name: attr('string'),
+  ...
+});


style/ember/README.md
```diff
@@ -2,6 +2,9 @@ Ember
 =====
 
 * Don't put a space between the opening handlebars braces and the variable.
+* Alias local versions of functions from `Ember` and `DS`. These will eventually
```
BlakeWilliams: Thoughts on the following?

> * Alias local variables to functions from `Ember` and `DS` using `const`.

seanpdoyle: Could the `const` part could be implied by our other `const` guideline?

style/ember/README.md
```diff
@@ -2,6 +2,9 @@ Ember
...
+  be moved to JavaScript modules allowing you to only import the functions you
+  require ([sample][local-Ember-DS])
```
seanpdoyle: I'm in favor of this, but does the guideline need to embed the motivation behind it?

MattMSumner: Maybe not, is there anywhere we would state the motivation behind a guideline?

seanpdoyle: `git blame`?

MattMSumner: :+1:

MattMSumner:
@BlakeWilliams @seanpdoyle @drapergeek is this good to go?

style/ember/sample.js
```diff
@@ -16,3 +16,13 @@ test("checks the box", function() {
...
 });
+
+import Ember from 'ember';
+import DS from 'ember-data';
+
+const { computed } = Ember;
+const { attr, hasMany, belongsTo } = DS;
+
+export default DS.Model.extend({
+  ...
```
BlakeWilliams: Maybe use the imported functions in the example here?

BlakeWilliams:
LGTM

style/ember/sample.js
drapergeek: You should probably import the model as well and use it since that would follow the same rules.

MattMSumner: :+1:

BlakeWilliams: Should we do this for constructors as well?

For example, importing `Ember.Object` would replace JavaScript's `Object`. There are potentially other places that they would collide, but I haven't thought of any.

drapergeek: That's a good point, I hadn't considered the `Object` scenario.

I think I'd be in favor of doing this whenever it doesn't cause issues but I don't feel strongly either way.

seanpdoyle: FWIW, I've also been warned about `const inject = Ember.computed.inject`

BlakeWilliams: Is there an `Ember.computed.inject`? Do you mean `Ember.inject.service()`?

MattMSumner: What about only doing this on functions one level deep? i.e. not `Ember.Object` or `DS.Model` and not on `Ember.computed.inject`. Not sure how best to word this...

MattMSumner: Although the guideline sort of says that.

BlakeWilliams:
```
*Alias local functions to functions from `Ember` and `DS`.
* Avoid aliasing more than 1 level deep, or constructors.
```

Maybe?

BlakeWilliams:
@MattMSumner Is this good to merge?

MattMSumner:
@BlakeWilliams thanks for the reminder! Merged.

PR #339, opened 2015-06-25
Prefer components over partials

MattMSumner:
Why:

* Partials share scope with the parent view.

This PR:

* Suggest using components. Using components will provide a consistent
scope.

### DIFF
--- a/style/ember/README.md
+++ b/style/ember/README.md
@@ -2,6 +2,7 @@ Ember
 =====
 
 * Don't put a space between the opening handlebars braces and the variable.
+* Prefer components over partials.
 
 Ember-Data
 ----------


style/ember/README.md
```diff
@@ -2,6 +2,7 @@ Ember
 =====
 
 * Don't put a space between the opening handlebars braces and the variable.
+* Don't use {{partial}} helper. Use a component with only a template.
```
BlakeWilliams: Thoughts on expanding this further?

> Don't use the `partial` or `view` handlebars helpers.
> Don't use `Ember.Select`.

drapergeek: I think don't is a bit much here. There are a lot of times that partials make more sense because you're just grouping markup. I could get behind `prefer` here.

BlakeWilliams: I assumed partial was deprecated, but it seems like it isn't. Will it be deprecated once routable components drop?

MattMSumner: I don't think there is a good reason to use `partial` over a component. The implicit context is unclear. In Rails we always say "Don't use instance variables in partials. Pass local variables to partials from view templates". I feel the same way here.

drapergeek: I don't think it will be deprecated. I think partial serves exactly the purpose I reach for it which is "this is a large section of markup that I'd like to organize. I will not be using it elsewhere I just want to make this template cleaner by removing some stuff from view".

seanpdoyle: How about:

`Prefer components over views and partials.`

MattMSumner: Since views have been depreciated is it worth adding them? If so it should certainly be a `don't` for using views?

drapergeek: I wouldn't mention views at all.

```
Prefer components over partials.
```

Since they're deprecated Ember should warn about those, we shouldn't feel the need to specify it.

BlakeWilliams: Agreed, I think that's a good guide.

MattMSumner: :+1:

JoelQ:
:+1:

drapergeek:
:+1:

PR #341, opened 2015-06-30
How do we feel about Ruby's `&method(...)`?

graysonwright:
This recently came up on a client project:

match_our_schema_with_loan_fuel_by_jakecraige_ _pull_request__252_ _fundingcircle_us-direct-borrower

I'm wondering how you all feel about Ruby's `&method(...)` syntax for defining blocks. I personally like it because it's shorter and it lets you pass methods around in a more functional-ish way.

Thoughts? The original post is at https://andrewjgrimm.wordpress.com/2011/10/03/in-ruby-method-passes-you/


jessieay:
I feel great about it; I think it makes total sense in contexts like the code in the screenshot you posted.

We actually have something about this in the guides, too:

```
Prefer &:method_name to { |item| item.method_name } for simple method calls.
```

https://github.com/thoughtbot/guides/tree/master/style/ruby#ruby

sikachu:
I think this is a different thing than the guide that @jessieay mentioned. This is about passing the collection to the method, and not calling method on the collection's member.

I think it's still too early to have style guide for this. Let's create a new TIL (as, holy shit, I think a lot of people just learned this), and maybe a research card so that we can try this out once we have a chance on several projects. Just need to use it to know how it feels, you know.

And of course, @graysonwright, you're totally welcome to use it in that project. I kinda like it, but It's just too soon to propose it as a style guide.

gabebw:
I agree with @sikachu - I've seen this used, like, once and always forget the syntax. I think it's too early to have a guide and I vote No Guideline.

mcmire:
I like this (and @sgrif is a fan of this, IIRC), but I agree with @gabebw and @sikachu here. Let's research this first.

sgrif:
I'm a fan of this, but I know a lot of people are against it. On MRI it's also worth noting that this is a *non-trivial* performance hit.

teoljungberg:
I'm torn of `&method(:whatever)`, I like it's shortness but it's also a bit magical. And something I wouldn't want to use in a client-project to hand back to the client.

I say it's too early for a guideline and vote No Guideline.

This would make a good `TIL` post though

bf4:
Any performance implications?

mike-burns:
This is a tool, like any other. You're always allowed to use tools. This doesn't seem like an appropriate thing for the guides to even mention; it would then have to mention all other methods on `Kernel`.

mcmire:
I would argue that it's a technique rather than just a tool. It's a specific application of a tool, in other words, to make syntax shorter.

Regardless of what it is, after thinking about it some more, I don't think it should be in the guides. It's not widely used, and like Sean said, there's a performance impact. I'm not discouraging anyone to use it, but I don't think it should be a guideline.

sgrif:
I don't have a source off the top of my head, but it's also worth noting that patches to add sugar for this have been rejected since some think that only having symbol to proc encourages encapsulation.

mcmire:
Should we close this?

graysonwright:
Wow, completely forgot about this. Agreed with everyone above, closing. Thanks!

PR #345, opened 2015-07-07
Adjust BAD example to use `subject`

jcwilk:
@gylaz

As per https://github.com/thoughtbot/guides/commit/ae1c6bb3a52ac5ea36cb16060e864f4c957effc7#commitcomment-12043014

Adjusting the "BAD" example of `let` vs `def` to utilize `subject`, `let`, and `context` since that at least shows the (perhaps flawed) intention of the "BAD" way to properly contrast the more complex structure against the simpler, more straightforward `def` form.

### DIFF
--- a/style/testing/avoid_let_spec.rb
+++ b/style/testing/avoid_let_spec.rb
@@ -1,4 +1,4 @@
-# BAD
+# Not recommended
 describe ReportPolicy do
   let(:report_id) { 2 }
   let(:report_policy) do
@@ -12,22 +12,18 @@ describe ReportPolicy do
     subject { report_policy.allowed? }
     
     context "when user has access to report" do
-      it "returns true" do
-        expect(subject).to eql(true)
-      end
+      it { should be true }
     end
 
     context "when user does not have access to report" do
       let(:report_id) { 3 }
       
-      it "returns false" do
-        expect(subject).to eql(false)
-      end
+      it { should be false }
     end
   end
 end
 
-# GOOD
+# Recommended
 describe ReportPolicy do
   describe "#allowed?" do
     context "when user has access to report" do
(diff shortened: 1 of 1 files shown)


style/testing/avoid_let_spec.rb
```diff
@@ -1,23 +1,27 @@
...
 
   describe "#allowed?" do
+    subject { report_policy.allowed? }
+    
     context "when user has access to report" do
       it "returns true" do
-        policy = ReportPolicy.new(user, report)
-
-        expect(policy).to be_allowed
+        expect(subject).to eql(true)
```
gylaz: Mind changing this to `eq`?

style/testing/avoid_let_spec.rb
```diff
@@ -1,23 +1,27 @@
 # BAD
```
gylaz: How about changing these to `# Not recommended` and `# Recommended`?

gylaz:
Thanks @jcwilk! A few minor nit-picks, then looks good to me.

style/testing/avoid_let_spec.rb
jcwilk: Hm, seems like "BAD" would be more consistent with the rest of the guide, but sure.

style/testing/avoid_let_spec.rb
jcwilk: sure, and i noticed you guys had the implicit subject recommended elsewhere in the guide just now so i'll change it to that too, figured you guys might not have liked it since it requires the `its` gem but it certainly is nice in these situations.

jcwilk:
@gylaz - I ended up doing `be true` since I think that's more semantic for `true`/`false`, but lmk if you prefer otherwise.

mcmire:
Looks good to me too. Merged as 29e011682de4797967eb47723768f6a9fed04501.

jcwilk:
Thanks @mcmire

PR #347, opened 2015-07-17
Avoid using non-standard DNS entries

derekprior:
We have frequently configured our sites and client sites to redirect the
`www` subdomain to the zone apex. For example, `www.thoughtbot.com`
redirects to `thoughtbot.com`. This works fine until you decide to put
the site behind a CDN.  Using a CDN like Fastly requires that you CNAME
your host to their servers. Per DNS spec, you cannot have a CNAME at
your zone apex.

We get around this by using DNSimple, which has an `ALIAS` record that
effectively simulates a CNAME at the apex. This isn't standard though,
and most providers do not offer it. Providers that do, often offer it as
an `ANAME` instead -- also non-standard. By relying on one of these
record types, we are locking ourselves in to these providers and more
importantly making it more difficult to be fault tolerant.

DNSimple went down hard last year and took all of our sites and
customers with them. Had we been using standard DNS servers it would
have been a best practice to have a backup provider listed as one of our
name servers. Backup providers can sync your DNS records from your
primary provider via zone transfer (`AXFR`). But the proprietary nature
of DNSimple's ALIAS was at least in-part responsible for this not being
possible with DNSimple. Even if we had access to our full zone record,
uploading to another provider would have meant temporarily compromising
the ALIAS feature.

DNSimple has since worked out secondary DNS offerings provided by a
limited number of providers. If those providers don't support `ALIAS`
then DNSimple pushes the IP that `ALIAS` resolves to as an `A` record.
So you're again throwing out the `ALIAS` functionality.

Most of this should "just work" now that they've put some effort behind
it, but it's still a lot of magic for something that can just be avoided
by forgoing the vanity of an apex domain. Yes, it looks prettier, but I
don't think it's worth all of this hullabaloo. I've learned to stop
worrying and love my www.

This PR comes out of a discussion on thoughtbot/suspenders/565.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -220,6 +220,11 @@ Browsers
 
 * Don't support versions of Internet Explorer before IE9.
 
+DNS
+---
+
+* Avoid using non-standard DNS entries (i.e. ALIAS, ANAME).
+
 Objective-C
 -----------
 


srt32:
For one more piece of validation, [Heroku recommends](https://devcenter.heroku.com/articles/apex-domains) the proposed approach as well.

best-practices/README.md
```diff
@@ -220,6 +220,11 @@ Browsers
 
 * Don't support versions of Internet Explorer before IE9.
 
+DNS
+---
+
+* Avoid using non-standard DNS entries (i.e. ALIAS, ANAME).
```
gylaz: Possible to word this as `Prefer ...`? I might forget what DNS standards are.

JoelQ:
*Very* nice commit message/PR description! :+1:

jferris:
Does it make sense to make the guideline mention apex domains explicitly? Our problem with apex domains is the proprietary DNS extensions, but the way almost everybody will experience it is with apex domains.

best-practices/README.md
mike-burns: Super grammatical nitpick but this feels more like an e.g. than an i.e. to me ("for example" instead of "that is").

derekprior: Nitpick appreciated and you are correct.

derekprior:
@jferris, do you have a suggestion on wording?

best-practices/README.md
derekprior: @gylaz I think I'd prefer to stick with `Avoid` here. There are lots of valid DNS entries you could use, and listing them all seems more likely to cause problems than trying to remember the two common ones you should avoid.

I think adding some wording mentioning apex domains as @jferris suggested might help some.

jferris:
How about something like:

>  Avoid using apex domains like `example.com`, which will require non-standard DNS entries like `ALIAS` or `ANAME`.

mjankowski:
Are these ALWAYS worth avoiding?  Is the risk of an issue high or severe enough that it's never worth the convenience trade-off?

Is this answer the same for every scale of app/site, or does the answer change on some spectrum? (like, maybe running your staging site on an ALIAS is fine, but running your bank's production site isn't)

derekprior:
I like it. They only *require* that when you try to use something that requires your site to be `CNAME`d like a CDN. Worth mentioning that?

> Avoid using apex domains like `example.com`, which will require non-standard DNS entries like `ALIAS` or `ANAME` when put behind a CDN such as Fastly.

derekprior:
> Are these ALWAYS worth avoiding? Is the risk of an issue high or severe enough that it's never worth the convenience trade-off?

The risk isn't necessarily from using non-standard dns itself, it's from not being able to use standard DNS features such as zone transfer to provide backup name servers. Any site that wants to maximize uptime should have backup name servers.

However, having thought about this more, given our usual deployment to Heroku, I don't think there's a great way to avoid using an `ALIAS` at some level.

Even if you agree to serve your site over `www.example.com` as the canonical host, you're going to need to route users that enter `example.com` over to `www.example.com`. In an environment where you had a dedicated VPS or load balancer or something you would set an `A` record that pointed the root to that IP (same as `www`). On Heroku, we can't reliably do this.

Heroku mentions being able to redirect using [subdomain redirection](https://devcenter.heroku.com/articles/custom-domains#configuring-dns-for-root-domains-if-your-dns-provider-does-not-support-alias-anames-cnames-for-root-domains) but that has a terrible UX for https:

> However, be aware that, using this method, a secure request to the root domain, e.g., https://example.com, will result in an error or warning being displayed to the user or the connection may be refused altogether.

*sigh*

I'll probably close this and investigate alternative suggestions for ensuring we can both use ALIAS and have backup name servers.

jferris:
> I'll probably close this and investigate alternative suggestions for ensuring we can both use ALIAS and have backup name servers.

Yeah, it's unfortunate, but I think you're right. We can't realistically prevent users from entering `example.com` directly, and there's no standard technology that will properly redirect them with https in place.

blackjid:
@derekprior have you found a better way to avoid using aliases??? anything you can recommend?

derekprior:
I recommend being comfortable with `www`, but also recognize this is a battle already lost.

blackjid:
what about this on heroku? just aliases??

> Even if you agree to serve your site over www.example.com as the canonical host, you're going to need to route users that enter example.com over to www.example.com. In an environment where you had a dedicated VPS or load balancer or something you would set an A record that pointed the root to that IP (same as www). On Heroku, we can't reliably do this.

derekprior:
Ah right, I forgot about that part. I don't have a good solution. If you make `www` your canonical host then external links will likely point there (including google, things users paste into slack, etc). You can use ANAME or ALIAS at your root to go to `www` to catch domain typers. If the ANAME/ALIAS goes away, your site will still actually work, it will just be difficult for some to get to. Not a great solution.

😢

blackjid:
thanks... I was losing hope of a clean way of doing it.... Now I lost it completely... but, ok. that's the way it is.

PR #348, opened 2015-07-21
Prevent spacing in objects for JavaScript

drapergeek:
ESlint recommends this style by default: http://eslint.org/docs/rules/object-curly-spacing

Let the discussion begin!

### DIFF
--- a/style/javascript/README.md
+++ b/style/javascript/README.md
@@ -19,6 +19,8 @@ JavaScript
 * Use `const` for declaring variables that will never be re-assigned, and `let`
   otherwise.
 * Avoid `var` to declare variables.
+* Do not include a space after the opening brace of an object or before the closing
+  brace.
 
 [template strings]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings
 [arrow functions]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions
--- a/style/javascript/sample.js
+++ b/style/javascript/sample.js
@@ -1,4 +1,4 @@
-object = { spacing: true }
+object = {spacing: false}
 
 class Cat {
   canBark() {


seanpdoyle:
I am against this.

My position is purely opinion, though, so I'm not sure how to argue in favor of the opposite.

JoelQ:
The article linked in the description only says that ESlint supports both options. Did I miss the section where they prefer no spaces over spaces?

BlakeWilliams:
They support both options, but they default to no spaces.

I'd like to choose this on a per-project basis, but if I had to vote one way or the other it would be without spaces purely because I think it's easier to argue for extra space than it is for less space.

drapergeek:
>I'd like to choose this on a per-project basis, but if I had to vote one way or the other it would be without spaces purely because I think it's easier to argue for extra space than it is for less space.

@BlakeWilliams Can you provide a reason why it'd be beneficial to choose this per project? Seems like the kind of thing that wouldn't provide any benefit either way so having consistency is really the only thing that matters.

JoelQ:
I lean towards spaces because that's how I write hashes in Ruby. Feel free to say that is a bad reason.

mcmire:
I don't like this because it means, yet again, we are diverting from the Ruby styles (single quotes vs. double quotes), which means it'll make it difficult to switch back and forth between Ruby and JavaScript.

BlakeWilliams:
@drapergeek In Ember projects I would lean towards using spaces since that seems to be convention there, but I find myself ignoring spaces in the React apps I write because I get to make the conventions there.

mike-burns:
I like this because this is how we write dictionaries in Python.

(But really: if this is what the community at large does, we should probably do it.)

drapergeek:
>I like this because this is how we write dictionaries in Python.

Elixir also requires that you omit the spaces. Both of these are minor things but if there is not a requirement or a large push one way, I think the minor things are worth considering.

>I don't like this because it means, yet again, we are diverting from the Ruby styles (single quotes vs. double quotes), which means it'll make it difficult to switch back and forth between Ruby and JavaScript.

If people are interested in this I would also like to put in a PR to solidify our Ruby style to also omitting the spaces. We don't have anything in our guides currently so unless you are using Hound, there is no requirement either way.

mcmire:
> If people are interested in this I would also like to put in a PR to solidify our Ruby style to also omitting the spaces.

Sure, go for it.

gylaz:
> If people are interested in this I would also like to put in a PR to solidify our Ruby style to also omitting the spaces. We don't have anything in our guides currently so unless you are using Hound, there is no requirement either way.

I could have sworn we had this Ruby rule in the guidelines, and thus based Hound off of it. I can't find anything though. I'm :+1: for standardizing this in Ruby too. Maybe also HAML?

MattMSumner:
I'm :+1: on this but purely based on opinion.

jkrmr:
> Elixir also requires that you omit the spaces. 

To be clear, it's (an) Elixir [style guide](https://github.com/niftyn8/elixir_style_guide)(s) that requires this, not Elixir itself:

> Whitespace might be (mostly) irrelevant to the Elixir runtime, but its proper use is the key to writing easily readable code.

Whether it's including or omitting spaces around delimiters that maximizes ease of readability is ~~subjective~~ an empirical question about social aggregates that I don't have sufficient data to form a conclusion about.

style/javascript/README.md
```diff
@@ -19,6 +19,8 @@ JavaScript
 * Use `const` for declaring variables that will never be re-assigned, and `let`
   otherwise.
 * Avoid `var` to declare variables.
+* Do not include a space after the opening brace of an object or before the closing
+  brace.
```
graysonwright: Should we link to the appropriate line in `sample.js`?

graysonwright:
In general, I'm in favor of following community guidelines. I don't think we should let style guides for other languages influence how we style our JS.

mcmire:
On second thought, if this is the community's guidelines then I say this is what we should follow, especially if we do more and more JavaScript-heavy projects. So I'm okay with this guideline.

seanpdoyle:
> the community's guidelines
>
> -- @mcmire

I don't think there are global "community guidelines", given JS's wild-west, every-module-for-itself nature.

For instance, [Ember's Styleguide](https://github.com/emberjs/ember.js/blob/c0c72266e22734e683e95f35d6b5dc7b19791c8f/STYLEGUIDE.md#objects) is in favor of spacing.

PR #350, opened 2015-08-07
Do not include spaces inside braces

drapergeek:
This is related to this PR: https://github.com/thoughtbot/guides/pull/348

We don't have a guide for this currently and I think in this case it may be best to have a rule that is the same for most languages (specifically JavaScript and Ruby), since there is no clear reason.

There is not a huge reason for either side of this debate. I think that the extra spacing doesn't provide much benefit. The other benefit to omitting the spacing is that both Python and Elixir require you to omit the braces. This is minor but I think when it comes to a decision that is purely based on preference, minor things make an impact.

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -42,7 +42,7 @@ Ruby
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
 * Order class methods above instance methods.
-* Do not include a space after the opening brace of an object or before the closing
+* Do not include a space after the opening brace of a hash or before the closing
   brace.
 
 [trailing comma example]: /style/ruby/sample.rb#L57
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -42,6 +42,8 @@ Ruby
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
 * Order class methods above instance methods.
+* Do not include a space after the opening brace of an object or before the closing
+  brace.
 
 [trailing comma example]: /style/ruby/sample.rb#L57
 [required kwargs]: /style/ruby/sample.rb#L24
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -42,7 +42,7 @@ Ruby
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
 * Order class methods above instance methods.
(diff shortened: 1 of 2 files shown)


style/ruby/README.md
```diff
@@ -42,6 +42,8 @@ Ruby
 * Prefer `protected` over `private` for non-public `attr_reader`s, `attr_writer`s,
   and `attr_accessor`s.
 * Order class methods above instance methods.
+* Do not include a space after the opening brace of a hash or before the closing
+  brace.
```
JoelQ: Would this apply to procs/blocks/lambdas as well? In the example code you apply this rule to a block even this guideline explicitly only mentions hashes.

Thoughts on:

* Changing the wording to something along the lines of "all uses of curly braces"
* Explicitly mentioning that block/procs/lambdas are also affected
* Only applying this rule to hashes and not changing the block example

drapergeek: >Changing the wording to something along the lines of "all uses of curly braces"

I can't see a compelling reason to have any differences depending on the location. I'll update the PR.

style/ruby/sample.rb
```diff
@@ -40,7 +40,7 @@ def method_with_single_method_block
   end
 
   def method_with_oneline_combined_methods_block
-    items.map { |item| "#{item.one} #{item.two}" }
+    items.map {|item| "#{item.one} #{item.two}"}
```
mcmire: It's going to take a while for me to get used to this. It looks so cramped :(

Should we keep the extra space at the end? I sort of think of it as going with the space after the `|`, not the space after the opening `{`.

drapergeek: Personally, I'm not in favor of the extra space for any reason.

jferris:
I'm not in favor of this change:

* Although we don't have a guideline, our code has generally included spaces around braces.
* Most community code seems to include the spaces.
* This conflicts with the bbatsov style guide.
* This violates the default setting in Rubocop/Hound.

gabebw:
I am against it for all of Joe's reasons, plus it looks different from many other codebases I've seen.

mike-burns:
> best to have a rule that is the same for most languages (specifically JavaScript and Ruby)

Those languages' cultures have opposing style guides, so I don't see why we should try to unite them. I'm -1 on this PR.

sikachu:
I'm not in the favor of this change as well. I share the same reasoning as @jferris

graysonwright:
Agreed with @jferris on this one - mainly because of the bbastov styleguide and the default hound config.

PR #353, opened 2015-08-26
Prefer `each_with_object` over `reduce`

bernerdschaefer:
There are a few possible ways
to build a hash from an array of items,
most commonly by `reduce` or `each_with_object`

`each_with_object` offers a few advantages over `reduce`:

  * The order of arguments is described by the method name
  * The order of arguments is familiar
    because of `each_with_index`
  * The block does not need to return the hash

Thus we prefer:

```ruby
items.each_with_object({}) do |item, hash|
  hash[item] = true
end
```

to:

```ruby
items.reduce({}) do |hash, item|
  hash[item] = true
  hash
end
```

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -19,6 +19,8 @@ Ruby
 * Prefer `select` over `find_all`.
 * Prefer `map` over `collect`.
 * Prefer `reduce` over `inject`.
+* Prefer `each_with_object` over `reduce` for building hashes from arrays.
+  [Example][each_with_object example]
 * Prefer double quotes for strings.
 * Prefer `&&` and `||` over `and` and `or`.
 * Prefer `!` over `not`.
@@ -45,4 +47,5 @@ Ruby
 
 [trailing comma example]: /style/ruby/sample.rb#L57
 [required kwargs]: /style/ruby/sample.rb#L24
-[class definition example]: /style/ruby/sample.rb#L121
+[class definition example]: /style/ruby/sample.rb#L127
+[each_with_object example]: /style/ruby/sample.rb#L104
--- a/style/ruby/sample.rb
+++ b/style/ruby/sample.rb
@@ -101,6 +101,12 @@ class SomeClass
     user.ensure_authenticated!
   end
 
+  def method_which_uses_each_with_object
+    items.each_with_object({}) do |item, hash|
+      hash[item] = true
+    end
+  end
+
   def self.class_method
     method_body
   end


teoljungberg:
I'm :-1: on this. I find `reduce` to have a clearer intent and is a more versitile tool than `each_with_object`

BlakeWilliams:
I prefer no guideline here.

jkrmr:
Preferring one to the other categorically would make sense if they were equivalent, but `inject` can accept an immutable object where `each_with_object` cannot.

```ruby
('a'..'z').inject(0) { |n, c|  c =~ /[aeiou]/ ? n += 1 : n  }
=> 5

('a'..'z').each_with_object(0) { |c, n|  c =~ /[aeiou]/ ? n += 1 : n }
=> 0

('a'..'z').each_with_object(0) { |c, n|  n+=1 if c =~ /[aeiou]/  }
=> 0
```

> Note that you can’t use immutable objects like numbers, true or false as the memo. 
[[source](http://apidock.com/rails/Enumerable/each_with_object)]

That said, for typical case of iterating with a accumulator collection, `each_with_object`, imho, is both more expressive and more succinct, since you don't need to explicitly return the accumulator in each iteration.

```ruby
('a'..'z').each_with_object([]) { |c, vowels|  vowels << c if c =~ /[aeiou]/  }
=> ["a", "e", "i", "o", "u"]

('a'..'z').inject([]) { |vowels, c|  vowels << c if c =~ /[aeiou]/ }
NoMethodError: undefined method `<<' for nil:NilClass

('a'..'z').inject([]) { |vowels, c|  vowels << c if c =~ /[aeiou]/; vowels }
=> ["a", "e", "i", "o", "u"]
```

mcmire:
I agree that we should know about `each_with_object`, and I personally prefer it when creating a hash over `reduce`, but I'm not sure I'd want a guideline here.

That said... isn't this a Hound recommendation already? I can't remember if it is or not.

jkrmr:
I think it might still be a rubocop default, but not enabled on Hound:

 https://github.com/thoughtbot/hound/commit/fec939b9d5b4

sikachu:
I vote no guideline here. Use it where you see fit, just like #341 (`&method(...)`)

mike-burns:
These are different methods that do different things.

Also, this is how I do it:

```ruby
items.inject({}) do |hash, item|
  hash.merge(item => true)
end
```

(As described here: https://robots.thoughtbot.com/iteration-as-an-anti-pattern#build-a-hash-from-an-array)

And anyway: I'm -1 on recommending mutation when doing OO.

bernerdschaefer:
As a bit more context,
I've been seeing this topic come up with relative frequency
in pull request discussions and in our Ruby discussion room.

I think by stating a preference --
whatever that preference may be --
it will require less mental effort
both to remain consistent within and across projects,
as well as to review code written by others.

mike-burns:
I'm interested in seeing such a discussion.

Like, I said, fold and `#each_with_object` are different things. So if people are misusing one then we should focus on that.

mcmire:
I don't think anyone is saying that `reduce` is misused if it's used to build a hash. It just so happens that `each_with_object` is a slightly superior approach the problem of "how do I build a hash from something else". Now whether it's superior enough to warrant a guideline... I don't personally think it is.

I think that the argument "you shouldn't change the thing you're building inside of a `reduce` or `each_with_object`" is a separate discussion, which I suppose is what you're saying. I think that would be a good best practice if others feel this is something we should note.

bernerdschaefer:
Does it change anything if I change this to: "Prefer reduce over each_with_object"?

Possible advantages:

  * It can be stated without qualification, versus the current PR which requires "...when building a hash from an array"
  * We use `reduce` for other tasks, so it should be familiar

mike-burns:
I still don't prefer that, because `#reduce` and `#each_with_object` have different use cases. Maybe something like "Don't forget that `Enumerable#each_with_object` exists" is closer to the goal?

jferris:
What about a blog post explaining the differences between `inject`/`reduce` and `each_with_object`?

bernerdschaefer:
It would seem that the only consensus reached is that this should not be a guideline.

PR #354, opened 2015-08-28
Recommend new-style Rails validations

gabebw:
Avoid:

validates :name, presence: true
validates :name, uniqueness: true

or:

validates_presence_of :name
validates_uniqueness_of :name

Using new-style validations like in the recommended example forces all of the
validations for a given column to be in one place, making it easy to scan a
model's validations.

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -19,8 +19,11 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
+* Use new-style `validates :name, presence: true` validations, and put all
+  validations for a given column together. [Example][validations].
 
 [order-associations]: /style/rails/sample.rb#L2-L4
+[validations]: /style/rails/sample.rb#L6
 [`app/views/application`]: http://asciicasts.com/episodes/269-template-inheritance
 
 Migrations
--- a/style/rails/sample.rb
+++ b/style/rails/sample.rb
@@ -2,4 +2,6 @@ class SomeClass
   belongs_to :tree
   has_many :apples
   has_many :watermelons
+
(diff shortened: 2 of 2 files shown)


mike-burns:
Eh, we don't really need a new rule about this. I can read both just fine.

tute:
I agree, I don't care enough, and it's not a meaningful enough difference to produce inertia while reading either wat.

gylaz:
I think we should have a rule for this (and I'm :+1: on this), that's the point of a style guide. I can read many different variants of code style, but I prefer to stick to one for consistency.

dgalarza:
:+1:

BlakeWilliams:
:+1:

seanpdoyle:
I'm :+1: on this, but there are times when I write `validate :name, presence: true` instead, which always takes me a few seconds / minutes to figure out why it isn't working.

With `validates_presence_of`, I don't run the risk of omitting the `s` without a `NoMethodError`

mcmire:
I don't *really* care about this, either, but I wouldn't be opposed to a style for this. 

To play devil's advocate, it's worth noting that `shoulda-matchers` was designed against the old style of specifying validations.

jakecraige:
I'm :+1: on this.

gabebw:
>  it's worth noting that shoulda-matchers was designed against the old style of specifying validations.

I think it's also worth nothing that shoulda-matchers was created before the new style existed :smile_cat:

mcmire:
Fair enough :)

jferris:
It would be possible (maybe useful?) to write a `validate` matcher that works with the hash-based `validates` method.

sikachu:
I'm :+1: on this, but just because it would guide us to do

```ruby
validates :name, presence: true, uniqueness: true
```
and not 

```ruby
validates :name, presence: true
validates :name, uniqueness: true
```

gabebw:
I am merging this.

PR #355, opened 2015-09-04
Order models attributes then relationships

MattMSumner:
Why:

* Ember data gives examples with this ordering:
https://github.com/emberjs/data#defining-your-models
* Open source ember projects appear to favor this ordering.

This PR:

* Changes the order in our sample model to attributes then
relationships.
* Clarify our style on ordering ember-data models

### DIFF
--- a/style/ember/README.md
+++ b/style/ember/README.md
@@ -14,8 +14,8 @@ Ember
 Ember-Data
 ----------
 
-* Visually separate relationships from attributes with a newline.
-  [Example][relationships]
+* Organize your models with attributes then relationships, separated with a
+  newline [Example][relationships]
 
 [relationships]: sample.js#L1-L7
 
--- a/style/ember/sample.js
+++ b/style/ember/sample.js
@@ -1,9 +1,9 @@
 App.Post = DS.Model.extend({
-  author: DS.belongsTo('user'),
-  tags: DS.hasMany('tag'),
-
   createdAt: DS.attr('date'),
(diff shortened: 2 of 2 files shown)


drapergeek:
:-1: 
I don't think we need a guide here. Depending on the model, it may be more appropriate to order them differently. It could also vary based on project.

seanpdoyle:
I am also opposed, but I think a guide would be helpful.

I favor:

1. Relationships
2. Attributes
3. Computeds

Largely because we have a similar guide for ActiveRecord models

MattMSumner:
@drapergeek can you give an example where the order being different would be more appropriate?

Styles don't need to be strictly adhered to. If there's a good reason not to follow the style then you shouldn't but that doesn't mean we should remove the guideline.

This PR came up because of [a discussion on FormKeep](https://github.com/thoughtbot/formkeep/pull/453#discussion_r38716548). Seems like we are writing these models with differing styles for no good reason beyond personal preference.

I can see an argument for going with Relationships, Attributes then Computeds to stay consistent with our rails guides. But then maybe we should revert using single quotes?

The convention in the community is Attributes, Relationships then Computeds. While we shouldn't blindly follow this, what are the advantages of bucking the trend?

BlakeWilliams:
I've been doing this for a long time now and I'm a huge fan of it. It makes it a lot easier to find what I'm looking for.

Big :+1: from me.

MattMSumner:
Closing for now.

PR #357, opened 2015-09-15
Update SCSS sample

tysongach:
I don’t think our current SCSS sample provides the full picture, and might even encourage a bit of bad behavior. So I updated to:

- Change the `@import` to indicate that partial names should use hyphens
- Remove nesting where it was unnecessary, which gave the wrong
  impression
- Use more class selectors and less element selectors
- Use actual CSS properties and values so that it renders with better code
  highlighting on GitHub
- Remove reference to Neat’s `media` helper, because Neat is not always used and
  this should focus more on plain SCSS style
- Show use of zero values to indicate that they should not have a unit
- Show that `@extend` should come before `@include`
- Show the multiple selectors should go on their own line
- Use `background-color` to indicate that we prefer it over just `background`

### DIFF
--- a/style/sass/sample.scss
+++ b/style/sass/sample.scss
@@ -1,53 +1,35 @@
-@import "bourbon";
+@import "partial-name";
 
-.featured-article {
-  @include mixin;
-  attribute: value;
+.class-one {
+  background-color: $color-variable;
+  border: 0;
 
-  @include media(value) {
-    attribute: value;
-  }
-
-  &.additional-selector {
-    attribute: value;
+  @media (min-width: $screen-variable) {
+    margin: ($spacing-variable * 2) 1rem;
   }
 
   &:hover {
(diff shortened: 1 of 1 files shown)


smharley:
:+1: big improvement

brendastorer:
Very nice. :dancer: :dancer:

style/sass/sample.scss
```diff
@@ -1,53 +1,35 @@
-@import "bourbon";
+@import "partial-name";
```
jasonramirez: What about `@import "stylesheet"`. My gut reaction to seeing "partial-name" is that its too programatic. Don't care that much though.

tysongach: I wanted to show a two-word name, to indicate the use of hyphens. Sass’s documentation uses the term “partial”, so I wanted to maintain that clarity.

PR #358, opened 2015-09-15
Avoid writing vendor prefixes

tysongach:
I’d like to encourage use of Autoprefixer. It might be a bit wordy.

Thoughts?

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -205,6 +205,10 @@ CSS
 ---
 
 * Use Sass.
+* Use [Autoprefixer][autoprefixer] to generate vendor prefixes based on the
+  project-specific browser support that is needed.
+
+[autoprefixer]: https://github.com/postcss/autoprefixer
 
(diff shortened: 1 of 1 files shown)


mike-burns:
This seems like a best practice.

tysongach:
Great point @mike-burns! Updated.

brendastorer:
Yes yes yes!!

joshuaogle:
Autoprefixer doesn't handle properties that aren't supported across browsers, like `-webkit-font-smoothing`, `-webkit-tap-highlight-color` and things like `::-webkit-scrollbar`. We rarely/never write those, but maybe "User Autoprefixer..." would be better than "Avoid ... and use ..."?

jasonramirez:
+1 to what Mr. Ogle says.

tysongach:
Updated.

joshuaogle:
Nice :musical_note:

PR #359, opened 2015-09-17
Soften our stance on IE support

tysongach:
According to the [README](https://github.com/thoughtbot/guides#guides), we should say “Don’t” when “…there's never a good reason”.

So should we not _ever_ work on a project that needs IE9 support?

I think “never” is a really strong word, especially in this case. Browser support depends on the project and the audience it has. We should analyze needs and openly discuss with the team, defaulting to IE10+ whenever we can.

cc: @thoughtbot/design

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -222,7 +222,7 @@ Sass
 Browsers
 --------
 
-* Don't support versions of Internet Explorer before IE10.
+* Avoid supporting versions of Internet Explorer before IE10.
 
 Objective-C
 -----------


conchan:
For reference, our Playbook currently says:

>We support Internet Explorer 10.0+ and the latest versions of Firefox, Chrome, and Safari. We do not support Internet Explorer 6, 7, 8, or 9. Those browsers are losing market share, they have security issues, and they are time-consuming to design for, develop for, and support.

This blurb makes me pretty happy. I'm pretty sure, 95-99% of the time, I'd veer towards a "No thanks..." if a project requires <IE10, for the same reasons I'm never excited to design/create emails—there's not a lot of new interesting challenges there, it's more like a time sink for banging on pulling-my-hair-out-in-frustration limitations. Generally, we like to do projects that are more cutting edge, so would we *want* to work on projects that NEED legacy IE support?

With that said, I'm not a Sith lord and I agree never is a strong word. I am okay with defaulting to IE10+ but leaving room to have discussions with potential/incoming clients about the pros and cons for needing IE9 and below AND the amount of extra time it may take to support it. If the project is extremely beneficial to thoughtbot and/or whoever's working on it when compared to the pains involved, then go for it?

I'm curious how many of our clients (past, current, potential, and lost) need <IE10

Also, we may have never been treated with this gem if everyone said no to doing legacy code  http://littlebigdetails.com/post/54829712822/psn-newsletter-when-images-arent-loaded-the

andrewjcohen:
I think you'd have to be crazy to pass up work because you don't want to deal with an older browser. Sometimes it has to be done, and it sucks, but money is money. We might not want to work on projects that support older browsers, but I'd rather work on anything than sit around because we were "too good" for a project that required me to fix a few bugs in IE9.

Right now IE9 and 8 actually hold more market share than 10 does from my understanding. I think we should at least soften the stance or make it a case-by-case basis. I have no problem with debugging older browsers, then again you're talking to someone who spent his early career drowning in IE6.

best-practices/README.md
```diff
@@ -218,7 +218,7 @@ Sass
 Browsers
 --------
 
-* Don't support versions of Internet Explorer before IE10.
+* Avoid supporting versions of Internet Explorer before IE10.
```
wardpenney: I like the `avoid`, but here's a suggestion:

`* Be conservative about support for versions Internet Explorer pre-IE10, erring on the side of not-supporting without a compelling business reason"

wardpenney:
For an ecommerce business, you may find yourself supporting IE8... **shudder**

best-practices/README.md
tysongach: Yep, I like the clarity in your wording. That is essentially what I was trying to get it. It just _depends_.

conchan:
IMO this change is more related to <IE9 support because we've only recently bumped it up to IE10 in the Playbook. Working around IE9 was more or less fine.

I'm not saying I'm "too good" for work. However, there is a strong emphasis here on working on projects we like. I'm not going to like it if a large portion of the work is fixing bugs for <3% of browser market share. If we're in a position where we can't pick and choose, then there's no question, but I like the idea of defaulting to IE10 (or even back down to IE9 if we find that this blurb in our Playbook is making people not hire us, though I doubt it). 

For FWIW, http://caniuse.com/usage-table

tysongach:
I dislike supporting pre-IE10, but I brought this up because we just had an in-depth conversation about supporting IE8 and IE9 on a certain _large_ project I’m working on. They were easily seeing 20,000+ sessions on some of these browsers. The whole product team had to dive into analytics, talk about the pros and cons and make informed decisions around this. That’s the key.

I was _aiming_ to not dilute our reasoning for supporting IE10 and up. I just felt that the current wording of never was too strong and acted as a sort of “shield” that we can wave and say “we _never_ support IE9!”.

joshuaogle:
The explanation of `avoid` more accurately reflects how we've been handling it (and should at this point in time). The longer wording that Ward suggested is just a more explicit version of "don't do it unless you have good reason", so I'm good with the simpler language. :+1: as is from me

conchan:
Jesus... My fault, i thought this was a PR on the Playbook... I need to stop looking at github afterhours. 🙏 

👍 to this guides update.

tysongach:
@conchan Ha! No worries! We should all be out with friends or something and not on GitHub.

Sooooooo, yeah… :runner:

best-practices/README.md
drapergeek: This seems like a lot of extra wording to spell out how we define "avoid" in the README already.

>"Avoid" means don't do it unless you have good reason.

Does the clarity provide any benefit over what we have already?

jasonramirez:
This is awesome:

> Be conservative about support for versions Internet Explorer pre-IE10, erring on the side of not-supporting unless there is a legitimate and compelling business reason

But, our guides don't generally justify the decisions we make, so I would say this belongs in the playbook? I am okay with `avoid`, but in the spirit of our guides, why not:

```
prefer supporting the last 2 major releases of a browser.
```

Something like https://support.google.com/a/answer/33864?hl=en

kaishin:
I like this change. I'm trying to not use critical technologies that are not supported in IE9 in my current project, since it's a product directed towards a senior/older audience that might be still using older software.

tysongach:
I’m going to merge this, unless anyone has any other comments. I think `avoid` captures what we are after.

PR #361, opened 2015-10-08
Adjust line-wrapping rules for Rails macros

iwz:
See slack conversation:
https://thoughtbot.slack.com/archives/ruby/p1444250120000353

This adds a new rule for Rails that overrides the default line break
rule in the main README.

This aims to make it easier to scan lists of macros and find the one
you're looking for without having to look at the following line if it's
wrapped.

See the new examples.

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -19,8 +19,15 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
+* If you break up an argument list for a macro,
+  keep the first argument on the same line as the macro,
+  and break the following arguments onto their own lines.
+  [Example 1][multi-line macro 1]
+  [Example 2][multi-line macro 2]
 
 [order-associations]: /style/rails/sample.rb#L2-L4
+[multi-line macro 1]: /style/rails/sample.rb#L4
+[multi-line macro 2]: /style/rails/sample.rb#L10
 [`app/views/application`]: http://asciicasts.com/episodes/269-template-inheritance
 
 Migrations
--- a/style/rails/sample.rb
+++ b/style/rails/sample.rb
@@ -1,5 +1,13 @@
 class SomeClass
   belongs_to :tree
   has_many :apples
+  has_many :pumpkins,
+    class_name: "Gourd",
+    foreign_key: :gourd_id
   has_many :watermelons
+
+  validates :name, presence: true
+  validates :title,
+    uniqueness: { scope: :tree_id, case_sensitive: false },
+    allow_nil: true
 end


gylaz:
I'm :-1:. It's just a method to me, and I prefer consistency. Yes, I wish Ruby forced parens everywhere.

jyurek:
I'd much prefer this style everywhere to the paren way because IMHO this look *way* better. But I'm with @gylaz in that I'd prefer consistency.

gabebw:
I prefer this style. :+1:

jessieay:
I like the seasonal examples! :jack_o_lantern: 

Rule is also nice. I haven't been doing this but I think the "easier to scan a list of macros" reasoning is sound.

jferris:
I have created a poll for this important question: https://docs.google.com/a/thoughtbot.com/forms/d/16TGUftRW0Y-Xo3wsj3hqllUsDmVG90AzkvLt6YjdbDQ/viewform?usp=send_form

derekprior:
I'm sympathetic to the consistency argument, but I think the gains here outweigh it. I'd say "DSL" rather than macro as it's more descriptive (and we don't have macros for real). But I agree. For example, routes:

```ruby
resource :store do
  resources :pumpkins, 
    only: [:index, :show, :edit, :update, :create],
    controller: :season_items

  resources :scarecrows
    only: [:index, :show, :edit, :update, :create],
    controller: :season_items
```

This let's me very easily see what the resources are and allows my brain to skip the indented implementation details until such time as its pertinent. The whitespace maintains its significance.

Same for Factory girl:

```ruby
describe "halloween" do
  it "is an opportunity for kids to get candy" do
    trick_or_treater = create :neighborhood_child,
      costume: "zombie_pirate",
      age: 8

    trick_or_treater.ring_bell

    expect(trick_or_treater).to get_candy
  end
end
```

With or without the parenthesis there, the significant whitespace allows me to more easily delineate *what* we're creating from what the attributes it has. In fact, in the past, in order to allow this type of formatting, I've tried to wrap the attributes in a hash so I can break up only the hash, but hound complains about that too.

jferris:
@derekprior the FactoryGirl example is interesting, and I think it pushes this beyond what we're discussing so far. I actually don't think of `create` as a DSL or a macro or anything.

style/rails/README.md
```diff
@@ -19,8 +19,15 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
+* If you break up an argument list for a macro,
+  keep the first argument on the same line as the macro,
```
calebthompson: I'd prefer to see non-hash arguments on one line until that gets crazy, then hash arguments on their own.

```
create :user, :admin,
  name: "Caleb Thompson",
  address: "..."

fill_form :user, :update,
  name: "Caleb"

extract_options foo: :bar, baz: :bat
```

calebthompson: So all the new examples are great, but I'd add another with multiple non-hash options together on one line.

gylaz:
I don't much see a difference between reading:
```ruby
validates :username,
  uniqueness: true,
  presence: true,
```
or
```ruby
validates(
  :username,
  uniqueness: true,
  presence: true,
)
```
To me it comes down to reading the word after `validates`, for the most important piece of information, vs reading the line right after `validates`. In physical proximity, I would argue that the second example puts the two main elements closer together.

mcmire:
I responded to the poll, but my personal style goes along with Caleb's: keep all options on one line until things get crazy, then split things up on multiple lines.

I like placing the first argument on the same line as the method, then other options on separate lines, because I see the first argument as going along with the macro name, and then the options as qualifying that macro. So when I see 

``` ruby
validates :username,
  uniqueness: true,
  presence: true
```

...I read it first as "Validates username". How? "Uniqueness and presence".

I think leaving off parentheses for macros is something that the Ruby community has been doing for a long time. So I don't see this as violating consistency. I don't think I would ever write the following, I think it would look weird:

``` ruby
validates(:username, presence: true)
```

mike-burns:
I had no idea anyone put parens around the arguments to a `has_many`. Instead of adding this rule, can we delete the rule that caused people to think that was an OK thing to do?

teoljungberg:
To be somewhat consistent with code I write in Ruby, when I write Rails, I commonly do:

```ruby
validates \
  :username,
  presence: true
```

To have it look, and feel somwhat alike i.e

```ruby
validates(
  :username,
  presence: true,
)
```

I find the two equally readable and vote for not adding this guideline.

mike-burns:
I've opened #362 to fix this.

jferris:
Are there any last votes here? Here's the current breakdown:



I think this is a clear enough majority that it can be merged on Friday if there are no major shifts.

mike-burns:
I'm in favor of #362 instead of this one. It has the same effect.

jferris:
@mike-burns I was thinking it makes sense to resolve this discussion, which is an incremental effort towards #362, and then consider removing it entirely as a separate discussion.

sikachu:
:+1:

dgalarza:
:+1:

jyurek:
With the addition of #362 I'm down with this. I kinda hate the multi-line parens style.

tute:
I'm :+1: on this.

jsom:
I'm :+1: on this style. This is the way I would write it on a fresh project, but in general I don't really care and would just remain consistent with whatever the current codebase is doing.

iwz:
Is this PR actually needed if we merge #362 first?

teoljungberg:
@iwz #362 resolves this issue

jferris:
Since #362 was merged in, I'm going to close this.

PR #362, opened 2015-10-09
Remove argument list style guide

mike-burns:
This guideline has proven to be the most unintuitive one, most often
broken and complained about in Slack and real life. It's the one that we
only get right after Hound complains other times. In short, it's not
natural.

Removing this guideline allows us to put the art back into formatting,
giving us the freedom to align arguments in the style that is most
readable and expressive.

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -39,9 +39,6 @@ Formatting
 * Don't include spaces after `(`, `[` or before `]`, `)`.
 * Don't misspell.
 * Don't vertically align tokens on consecutive lines.
-* If you break up an argument list, keep the arguments on their own lines and
-  closing parenthesis on its own line. [Example 1][multi-line arguments]
-  [Example 2][long method signature]
 * If you break up a hash, keep the elements on their own lines and closing curly
   brace on its own line.
 * Indent continued lines two spaces.
@@ -59,8 +56,6 @@ Formatting
 * Use [uppercase for SQL key words and lowercase for SQL identifiers].
 
 
-[multi-line arguments]: /style/ruby/sample.rb#L69
-[long method signature]: /style/ruby/sample.rb#L16
 [dot guideline example]: /style/ruby/sample.rb#L11
 [uppercase for SQL key words and lowercase for SQL identifiers]: http://www.postgresql.org/docs/9.2/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
 [newline explanation]: http://unix.stackexchange.com/questions/23903/should-i-end-my-text-script-files-with-a-newline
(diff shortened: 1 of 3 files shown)


style/README.md
```diff
@@ -39,9 +39,6 @@ Formatting
 * Don't include spaces after `(`, `[` or before `]`, `)`.
 * Don't misspell.
 * Don't vertically align tokens on consecutive lines.
-* If you break up an argument list, keep the arguments on their own lines and
-  closing parenthesis on its own line. [Example 1][multi-line arguments]
```
teoljungberg: Should we remove `multi-line arguments` and `long method signature` while we're at it? Or are they referenced elsewhere?

teoljungberg:
I find that all different implementations of the guideline are equally easy to read. And it doesn't matter which one we pick, because we all have different styles when it comes down to it. Let's embrace that.

style/README.md
mike-burns: Done.

teoljungberg: Sadly, how the samples are organized. A few links might have been broken since they are done by line-number. That's a shame

mike-burns: Fixed.

teoljungberg: Stellar

mcmire:
Yeah... I'm afraid to say that this is better. :clap: 

> I find that all different implementations of the guideline are equally easy to read.

I think that's the first time I've heard that :smile:

teoljungberg:
@mcmire I still have my personal preference, but so does everyone else

derekprior:
Is it worth leaving a recommendation that we don't line up arguments with indentation? This is what vim does by default and it leads to more diff noise when the method name changes as now the alignment should change as well.

I'll continue to just not do it, but I'd love to see that remain something we avoid doing as a rule.

mcmire:
> Is it worth leaving a recommendation that we don't line up arguments with indentation?

Are you talking about something like this?

``` ruby 
some_really_long_method(foo,
                                           bar,
                                           baz)
```

derekprior:
Yes. That's what I was talking about, but honestly haven thought about it
some more I really don't care. Maybe we *should* just do what vim wants to
do.

mcmire:
I think the style that vim-ruby forces you to follow is terrible, IMO (I've even [patched the plugin](e3d63ee4c9d059ffbb980a32860033f047f3871a) to remove this behavior).

sikachu:
I'm in favor of removing this. :+1:

iwz:
:+1:

iwz:
Is this ready to merge?

mike-burns:
@derekprior you voiced the concern. Cool if I merge this, and we'll tackle it again if it becomes an issue?

derekprior:
Fine by me.

PR #363, opened 2015-10-16
Prefer `--force-with-lease` over `--force`

calleerlandsson:
Using `--force-with-lease` allows one to force push without the risk of
unintentionally overwriting someone else's work.

The git-push(1) man page states:

> Usually, "git push" refuses to update a remote ref that is not an
> ancestor of the local ref used to overwrite it.
>
> This option overrides this restriction if the current value of the
> remote ref is the expected value. "git push" fails otherwise.
>
> --force-with-lease alone, without specifying the details, will protect
> all remote refs that are going to be updated by requiring their
> current value to be the same as the remote-tracking branch we have for
> them.

### DIFF
--- a/protocol/git/README.md
+++ b/protocol/git/README.md
@@ -98,7 +98,7 @@ Force push your branch. This allows GitHub to automatically close your pull
 request and mark it as merged when your commit(s) are pushed to master. It also
  makes it possible to [find the pull request] that brought in your changes.
 
-    git push --force origin <branch-name>
+    git push --force-with-lease origin <branch-name>
 
 View a list of new commits. View changed files. Merge branch into master.
 


mike-burns:
Yes!

teoljungberg:
I love `--force-with-lease`, and it fits nicely into our rebase workflow.

`<sidebar>` I've intended to write a blogpost about it for months. Maybe one day. `</sidebar`

seanpdoyle:
This sounds great, but it's very verbose.

Is there an abbreviation? Is there any way to override the `-f` flag to do this instead?

teoljungberg:
There is no abbreviation according to the man page

mike-burns:
Nor could I find one in the codebase.

mike-burns:
If it's too verbose, may I recommend:

```ini
[alias]
  pf = push --force-with-lease
```

seanpdoyle:
@mike-burns is that's the case, should there be a matching PR to the dotfiles once if this is merged?

mike-burns:
@seanpdoyle no complaints from me!

seanpdoyle:
@mike-burns https://github.com/thoughtbot/dotfiles/pull/431

calleerlandsson:
I'll merge this.

sampablokuper:
@teoljungberg [wrote](https://github.com/thoughtbot/guides/pull/363#issuecomment-148665193):
> I've intended to write a blogpost about it for months. Maybe one day.

Looks like one of your colleagues has [done this for you](https://robots.thoughtbot.com/git-push-force-with-lease) :)

PR #366, opened 2015-10-22
Prefer slim over haml for rails apps

iwz:
After much discussion at thoughtbot, we've come to realize that we
generally prefer to use either ERb or Slim over HAML for templating.

https://trello.com/c/HxrnodBf/271-slim

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -19,7 +19,7 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
-* Prefer `ERb` or [`slim`][slim-lang] over `haml` for templating
+* Prefer ERb or [Slim][slim-lang] over HAML for templating
 
 [order-associations]: /style/rails/sample.rb#L2-L4
 [`app/views/application`]: http://asciicasts.com/episodes/269-template-inheritance
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -19,7 +19,7 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
-* Prefer ERb or [Slim][slim-lang] over HAML for templating
+* Prefer ERB or [Slim][slim-lang] over HAML for templating
 
 [order-associations]: /style/rails/sample.rb#L2-L4
 [`app/views/application`]: http://asciicasts.com/episodes/269-template-inheritance
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -19,9 +19,11 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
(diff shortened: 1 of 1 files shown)


jferris:
ERb/Slim/HAML are names rather than code, so I don't think you want backticks around them.

iwz:
@jferris resolved in 0112baa

style/rails/README.md
```diff
@@ -19,9 +19,11 @@ Rails
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
+* Prefer ERb or [Slim][slim-lang] over HAML for templating
```
BlakeWilliams: Is ERB supposed to be all caps? http://ruby-doc.org/stdlib-2.2.3/libdoc/erb/rdoc/ERB.html

iwz: Certainly seems that way. Resolved in 31ad8d2

teoljungberg:
I'm against this guideline, I strongly prefer ERB over HAML. But, I think everyone should be free to make their own choice which templating language to use.

Thoughts on removing this guideline entirely and let everyone pick and choose as they please?

iwz:
> I'm against this guideline, I strongly prefer ERB over HAML.

@teoljungberg the PR says:

> Prefer ERB or Slim over HAML

Seems like you're agreeing and disagreeing at the same time?

teoljungberg:
@iwz I agree with the choice of ERb over HAML or Slim. I disagree with having a guideline for it

joshuaogle:
I'm against having a guideline as well. I slightly prefer HAML over Slim, and strongly prefer either of them to plain ERB. I think we have enough ERB/HTML hardliners though that I'm fine with not having a guideline.

iwz:
@joshuaclayton mind weighing in on the research card? https://trello.com/c/HxrnodBf/271-slim

mcmire:
I prefer not to have a guideline on this. I did not get a sense reading through that card that people who once were for Haml now prefer Slim. In fact it was pretty split:

* Draper prefers ERB over Haml or Slim
* Trevor doesn't like `|

tabfugnic:
@mcmire Very interested to hear what designers think. Anecdotally I've heard more support for ERb over anything else since it's closest to HTML. I'm torn on the guideline for this.

andrewjcohen:
I just started using slim but before that was a haml lover. After a few days I now prefer slim over haml over erb.

kylefiedler:
I'm not sure that this needs to be a guide. I've used all three and they all have pros and cons. There isn't one of them that I would say I prefer heavily over the other. This choice doesn't seem as clear as the choice between Sass and Less where I feel like there are clear benefits to using Sass. I'm happy to leave it open to the designers and developers on the project.

I do think we should say that once that decision is made by the project team we need to be consistent in that choice. There shouldn't ever be a project that has ERB and HAML and Slim in it or any combo of the three.

jasonramirez:
I prefer Slim over Haml. Haml over ERb. But, I don't feel strongly that we need a guideline for this.

gylaz:
It seems that most designers do prefer Slim over HAML, while mostly developers have expressed ERB preference over others. Since designers spend the most time in views, would it be fair to give their voice a higher precedence?

mjankowski:
I'd vote for not having a guide on this topic, other than a strong preference to:

* Use the same library within a given project
* Not totally gut a project *just* to switch to another template library without a really really really strong reason to do so

sikachu:
I vote for not having a guide on this topic as well.

iwz:
No guide wins it!

PR #368, opened 2015-11-03
Use blocks when declaring date and time attributes

seanpdoyle:
`FactoryGirl` factories can lazily instantiate values. Since `Date`,
`Time`, and `DateTime` can be affected by utilities such as `Timecop`
and `ActiveSupport` (configured at the test level), we should avoid
instantiating instances until the last possible moment.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -97,7 +97,9 @@ Rails
 * Prefer `Time.zone.parse("2014-07-04 16:05:37")` over `Time.parse("2014-07-04 16:05:37")`
 * Use `ENV.fetch` for environment variables instead of `ENV[]`so that unset
   environment variables are detected on deploy.
+* [Use blocks][date-block] when declaring date and time attributes in FactoryGirl factories.
 
+[date-block]: /best-practices/samples/ruby.rb#L10
 [fkey]: http://robots.thoughtbot.com/referential-integrity-with-foreign-keys
 [`.ruby-version`]: https://gist.github.com/fnichol/1912050
 [redirects]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.30
--- a/best-practices/samples/ruby.rb
+++ b/best-practices/samples/ruby.rb
@@ -4,3 +4,9 @@ class PostSerializer < ActiveModel::Serializer
 
   delegate :to_param, to: :object
 end
+
+FactoryGirl.define do
+  factory :event do
+    start_on { 1.week.from_now }
+  end
+end


geoffharcourt:
Is there any difference between `1.week.from.now` and `Date.current + 1.week` if time zones are involved?

sikachu:
@geoffharcourt Nope. Both are based on `Date/Time.current` so it will reflect local time zone.

I'm ±0 about this change. It's a good practice, but I don't know if it has to be added.

geoffharcourt:
We just got burned on consecutive days by specs failing over this, I think it's worth adding.

seanpdoyle:
@sikachu I would say it's a *best* practice.

@geoffharcourt and I have been chasing intermittently failing tests for the past two days because of it.

BlakeWilliams:
I'm :+1:. It's pretty easy to miss.

sikachu:
:+1:

iwz:
:+1:

mcmire:
Makes sense to me. :+1:

thorncp:
Your pain will be our gain :+1:

joshuaclayton:
Should we disallow static (non-block) assignment of attributes entirely? The only time I use static attributes is when the value being assigned is numeric or boolean. Arrays, hashes, and even strings (e.g. with `upcase!`) can be mutated, which can cause all sorts of havoc on a test suite. My gut says if we're running into these issues, others are too.

gabebw:
> Should we disallow static (non-block) assignment of attributes entirely? 

Factory Girl PR time!

mcmire:
That actually isn't a bad idea... I can't think of any downsides, other than a little more verbosity.

seanpdoyle:
@joshuaclayton :+1:  in favor of that.

geoffharcourt:
If we're going to not allow static definitions, would we be better off changing FactoryGirl so that every definition happens in a Proc/Lambda without any extra syntax?

sikachu:
@joshuaclayton I don't mind seeing it in major bump with proper deprecation. I wonder how much ripple effect this will cause though (OMG now I have to update *all* of my factories to include these braces!)

I think we want to do some benchmark test before changing it though. For small app it probably won't matter. For larger app ... I wonder how much it would affect the build time and memory footprint.

Maybe the good start is to open an issue on FG for "disallow" Date/DateTime/Time as a static value. I think that's a good start.

This PR can be merge as-is, and then let's discuss the change on that FG PR ;)

jferris:
There is a performance hit for using dynamic attributes when you could use static attributes. I tried this and benchmarked it maybe 5 or 6 years ago. If we decide to remove static attributes from factory_girl, we should probably run benchmarks again.

mike-burns:
@geoffharcourt sounds great; how would that work?

I'm +1 on this PR.

geoffharcourt:
@mike-burns rethinking my suggestion after @jferris' comment. I was thinking that we would do it through `method_missing`, but the more I think about it, the more complicated and unpleasant that sounds.

If there are FactoryGirl decisions to be made, I think we should open an issue in the FG repo. I think for now we've got broad consensus around best practices with relative dates in factories. @seanpdoyle let's merge!

teoljungberg:
PR's to guides should be open for 7 days, see https://github.com/thoughtbot/guides/blob/master/CONTRIBUTING.md#getting-feedback

PR #369, opened 2015-11-04
Document private extension guideline in styleguide

dgalarza:


sidraval:
I'm into this :+1:

style/swift/README.md
```diff
@@ -11,3 +11,4 @@ Swift
 * Let the compiler infer the type whenever possible
 * Group computed properties below stored properties
 * Use a blank line above and below computed properties
+* Group `private` methods in a `private` `extension`
```
gfontenot: how about

> group methods with different levels of access control in specific extensions.

That's horribly worded, but the gist is: we should be doing this for `public` as well.

gilesvangruisen: maybe:
> group methods into specific extensions for each level of access control

dgalarza:
Updated to make this guideline not so specific to private extensions, but to grouping by level of access control.

gfontenot:
Dig it :+1:

PR #371, opened 2015-11-10
JavaScript: Avoid referring to `arguments`

seanpdoyle:
*Avoid* using `arguments` unnecessarily.

Inspired by [petkaantonov/bluebird][wiki]'s "Optimization Killers"
wiki.

Using `arguments` without `.length` or `[i]` makes it nearly impossible
for V8 to optimize the function block.

[wiki]: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -189,13 +189,15 @@ Email
 
 JavaScript
 ----------
-* Avoid referring to `arguments`.
+* Avoid referring to `arguments`. As an alternative to `arguments`, use the
+  spread syntax with a variable that is not `arguments`. [sample][spread]
 * Use Coffeescript, ES6 with [babel], or another language that compiles to
   JavaScript
 * Include a `to_param` or `href` attribute when serializing ActiveRecord models,
   and use that when constructing URLs client side, rather than the ID. Example:
 
 [babel]: http://babeljs.io/
+[spread]: samples/javascript.js#L1-L3
 
 HTML
 ----
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -189,6 +189,7 @@ Email
 
 JavaScript
 ----------
+* Avoid referring to `arguments`.
 * Use Coffeescript, ES6 with [babel], or another language that compiles to
   JavaScript
 * Include a `to_param` or `href` attribute when serializing ActiveRecord models,
--- a/best-practices/samples/javascript.js
+++ b/best-practices/samples/javascript.js
@@ -1,0 +1,3 @@
+function foo(...args) {
+  // ...
+}


BlakeWilliams:
:+1:

joshuaclayton:
:+1:

drapergeek:
:+1:

mcmire:
Yup, makes sense. (And you don't need this anymore with the argument spread syntax.)

seanpdoyle:
@mcmire avoiding `...arguments` is what this is addressing

mcmire:
Oh, I thought this was addressing avoiding referring to `arguments`, which is a "special variable" automatically available inside any function.

AFAIK, the spread syntax actually creates a new variable that is non-special. (You could write `...arguments` or `...args` or `...whatever`, in other words.) Or I am off-base on that?

seanpdoyle:
@mcmire I'm not sure. @BlakeWilliams made that comment that inspired this PR, so maybe he know for certain.

BlakeWilliams:
[This example](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&playground=false&code=class%20Foo%20%7B%0A%20%20constructor(a%2C%20b%2C%20c)%20%7B%0A%20%20%20%20%0A%20%20%7D%0A%7D%0A%0Aclass%20Bar%20extends%20Foo%20%7B%0A%20%20constructor(a%2C%20b%2C%20c)%20%7B%0A%20%20%20%20super(...arguments)%3B%0A%20%20%7D%0A%7D) shows that it's still using `arguments` directly.

mcmire:
Ah, good to know, thanks.

It seems like the following syntax is prohibited in strict mode, as well:

``` javascript
function foo(...arguments) {
  // ...
}
```

However, you can say this:

``` javascript
function foo(...args) {
  // ...
}
```

So, do you think that this is useful to distinguish? i.e. something like "As an alternative to `arguments`, use the spread syntax with a variable that is not `arguments`."

BlakeWilliams:
I think it's worth mentioning an alternative any time we recommend against something. :+1: to `...args`

seanpdoyle:
@mcmire @BlakeWilliams I've pushed an update with a sample.

mcmire:
Sweet, looks good!

BlakeWilliams:
Looks great!

seanpdoyle:
During this [deep dive into Ember.View](https://youtu.be/7kU5lLnEtJs?t=4m24s), @rwjblue replaces `this._super(this, arguments)` with `this._super(...arguments)`.

rwjblue:
Technically, I replaced `this._super.apply(this, arguments)` with `this._super(...arguments)`.  The later transpiles to the former (as pointed out earlier in the thread above).  

Also, the [article referenced](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#what-is-safe-arguments-usage) specifically states that calling `fn.apply` is a special exemption from the deopts associated with the `arguments` object.

seanpdoyle:
@BlakeWilliams from the wiki:

> STRICTLY fn.apply(y, arguments) is ok, nothing else is, e.g. .slice. Function#apply is special.

From your [example](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&playground=false&code=class%20Foo%20%7B%0A%20%20constructor(a%2C%20b%2C%20c)%20%7B%0A%20%20%20%20%0A%20%20%7D%0A%7D%0A%0Aclass%20Bar%20extends%20Foo%20%7B%0A%20%20constructor(a%2C%20b%2C%20c)%20%7B%0A%20%20%20%20super(...arguments)%3B%0A%20%20%7D%0A%7D):

>     _get(Object.getPrototypeOf(Bar.prototype), "constructor", this).apply(this, arguments);

seanpdoyle:
Given this new information, I'm closing this PR.

PR #372, opened 2015-11-15
Add `style/javascript/.jscsrc`

seanpdoyle:
Inspired by our Ruby styleguide, which has an accompanying
`.rubocop.yml.

Adding a JSCS linter to thoughtbot/guides allows us to configure Hound
to use our company-wide styleguide:

```yml
jscs:
  enabled: true
  config_file: https://raw.githubusercontent.com/thoughtbot/guides/master/style/javascript/.jscsrc
```

Uses new [`JSCS@2.6.x`][jscs] rules.

[jscs]: https://github.com/jscs-dev/node-jscs/releases/tag/v2.6.0

### DIFF
--- a/style/javascript/.jscsrc
+++ b/style/javascript/.jscsrc
@@ -1,0 +1,80 @@
+{
+  "esnext": true,
+  "verbose": true,
+  "disallowEmptyBlocks": true,
+  "disallowKeywordsOnNewLine": ["else"],
+  "disallowMultipleLineBreaks": true,
+  "disallowMultipleVarDecl": "exceptUndefined",
+  "disallowNewlineBeforeBlockStatements": true,
+  "disallowSpaceAfterObjectKeys": true,
+  "disallowSpaceAfterPrefixUnaryOperators": true,
+  "disallowSpaceBeforePostfixUnaryOperators": true,
+  "disallowSpaceBeforeSemicolon": true,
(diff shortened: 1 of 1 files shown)


seanpdoyle:
This initial `.jscsrc` has been extracted from a current client project, and should NOT be considered authoritative.

However, its a good starting point, and will serve as a good way to track our conversations around including/excluding JSCS's various rules.

BlakeWilliams:
Thoughts on using ESLint instead?

ESLint seems to be faster, pluggable, and goes beyond just style checking. I've used it a few times in the past and it's been great. I've never used JSCS though since ESLint has the same rules I believe.

seanpdoyle:
@BlakeWilliams What do you think about maintaining both?

BlakeWilliams:
That'd be fine. I can work on the `.eslintrc`.

Just curious, but is there a time when we'd want to pick JSCS over ESlint?

gylaz:
Out-of-the-box Ember provides config file for JSHint. JSCS is a companion tool to JSHint. Would it make sense not to go against what Ember recommend out of the box (or at least support it for those that prefer JSHint/JSCS)?

teoljungberg:
Since JSCS is merged into Hound now. Should we merge this PR, @seanpdoyle ?

PR #375, opened 2015-12-04
Prefer local assets

jferris:
Externally linked assets, such as Typekit, slow down page loading and
rendering. When possible, we should use assets served from the same
domain and connection as the source HTML page, which keeps latency and
load times down.

If we have to use an external asset, we should use https URLs to avoid
leaking the referrer and other information over public channels.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -191,6 +191,13 @@ Email
 [MailView]: https://github.com/37signals/mail_view
 [ActionMailer Preview]: http://api.rubyonrails.org/v4.1.0/classes/ActionMailer/Base.html#class-ActionMailer::Base-label-Previewing+emails
 
+Web
+---
+
+* Avoid a Flash of Unstyled Text, even when no cache is available.
+* Avoid rendering delays caused by synchronous loading.
+* Use https instead of http when linking to assets.
+
 JavaScript
 ----------
 * Use Coffeescript, ES6 with [babel], or another language that compiles to


best-practices/README.md
```diff
@@ -187,6 +187,14 @@ Email
 [MailView]: https://github.com/37signals/mail_view
 [ActionMailer Preview]: http://api.rubyonrails.org/v4.1.0/classes/ActionMailer/Base.html#class-ActionMailer::Base-label-Previewing+emails
 
+Web
+---
+
+* Prefer local assets to externally linked assets.
```
jakecraige: How does this relate to cdn's? Those are external assets but generally preferred

derekprior:
Are there places where we're doing this without a good reason? I haven't seen people load images remotely for quite some time. We do reference remote scripts on occasion, but that's usually at the suggestion of whomever we're using that script to integrate with.

I do see fonts loaded remotely, but I think that's often okay - it's just that TypeKit isn't a particularly good implementation of that. Google Fonts, for instance, does a lot of things we would never bother to do locally that slims down fonts while also making cache hits more likely.

derekprior:
To be specific, here's a rundown of what using Google Fonts over serving the same font locally would offer as advantages.

> * Are you going to implement unicode-range optimization yourself? What if your designer changes fonts?
> * Are you going to come up with 30+ varieties of the same font, like Google Fonts does, to serve the perfect one to each user agent?
> * Are you going to strip the font-hinting from your font files to save an extra couple of KB?
> * What if a new font technology comes out (like WOFF2 did) and even more speed becomes possible? Are you going to implement that yourself?
> * Are you absolutely sure that there's no major benefit afforded by users having already downloaded your font on another site using Google Fonts?

That and lots of other great detail about why this matters at all, and why typekit is "slow" can be found in [Nate Berkopec's article on speeding up RubyGems](http://www.nateberkopec.com/2015/11/30/how-changing-webfonts-made-rubygems-10x-faster.html) by judiciously using web fonts.

Since this change came out of a discussion about remote font loading, I think it's worth noting that we'd be unlikely to do better than Google Fonts when delivering our own fonts.

best-practices/README.md
mike-burns: For http/2, we should prefer local assets instead of a CDN. But we're not close to http/2 yet.

kaishin:
@derekprior AFAIK services like myfonts allow you to purchase Web fonts that are already optimized for Web.

best-practices/README.md
jferris: I think if we're hosting all the static assets on one CDN, we prefer that. However, I think we'd generally prefer using "our" CDN for whatever project it is instead linking to several external CDNs.

jferris:
@derekprior there are a lot of good points there. My motivation for opening this was definitely Typekit. I think it would be worth evaluating what the best option for web fonts, whether it's buying pre-optimized fonts and serving them ourselves or linking out to a more efficient service like Google Fonts.

This seems like the best way to start this discussion, but maybe we should get some Research cards going?

derekprior:
I'd be interested in an experiment where do an experiment similar to what Nate did with one of our products or the blog and measured the results, particularly on lower-bandwidth connections. I don't think I'll have investment time to do that for a bit, though.

jasonramirez:
For what its worth, I am a huge proponent of Google Fonts. I have never personally had an issue with the service, I have never seen FOUT (great cacheing), and it is incredible easy to implement (single line of code), it is free, and there are a myriad of great fonts that work with all the user agents. 

In my personal opinion, I dislike serving fonts locally as it is far more work what I think is very little gain. I would be in favor of preferring Google Fonts over Typekit and Google Fonts over serving locally.

jferris:
Any thoughts on changing the guidelines to cover the effects instead of the cause? Something like:

* Avoid a Flash of Unstyled Text, even when no cache is available
* Avoid rendering delays because of synchronous loading

jasonramirez:
I like that. Its less opinionated about a particular service, and gets right to the core of why we are even talking about it. Small language suggestion:

Avoid rendering delays ~~beacuse of~~ _caused by_ synchronous loading

mjankowski:
That's a good change -- and delegates the discussion of whether a particular service/host/etc will actually cause issues to the team/situation in question.

jferris:
Alright, I pushed an update to use the newer guidelines as discussed.

jasonramirez:
:hamburger: from me as is.

joshuaogle:
I really like those changes. :rocket: for me

best-practices/README.md
```diff
@@ -187,6 +187,13 @@ Email
 [MailView]: https://github.com/37signals/mail_view
 [ActionMailer Preview]: http://api.rubyonrails.org/v4.1.0/classes/ActionMailer/Base.html#class-ActionMailer::Base-label-Previewing+emails
 
+Web
+---
+
+* Avoid a Flash of Unstyled Text, even when no cache is available.
+* Avoid rendering delays caused by synchronous loading.
+* Avoid linking to assets over http.
```
jakecraige: I'm not sure I understand what this means, should I avoid ``?

jferris: It means we should use `https` instead. When using external font services, trackers, and widgets, I've seen a few cases where people bleed information by using `http`.

mcmire: Maybe "Avoid linking to assets over HTTP, but use HTTPS instead"?

jferris: I pushed 9c1d7d1 to try and clarify this.

mcmire: :+1:

mjankowski:
Do it.

MattMSumner:
:+1: :shipit:

PR #377, opened 2015-12-09
Recommend `belongs_to :relation, touch: true`

seanpdoyle:
Based on @sgrif's recommendation (near 17:19) on [The Bike Shed][bikeshed-41].

Thinking about an application's caching strategy and its implications to
the data model is usually put off until some point in the future.

`touch`ing `belongs_to` relationships is a straightforward way to
pre-emptively configure applications to integrate with Rails' russian
doll caching.

Rails 5 will batch `touch`-generated DB writes.

[bikeshed-41]: http://bikeshed.fm/41

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -100,6 +100,7 @@ Rails
 * Use `ENV.fetch` for environment variables instead of `ENV[]`so that unset
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryGirl factories.
+* Use `touch: true` when declaring `belongs_to` relationships.
 
 [date-block]: /best-practices/samples/ruby.rb#L10
 [fkey]: http://robots.thoughtbot.com/referential-integrity-with-foreign-keys


gylaz:
Can you provide a bit more reasoning in the commit description so it's captured in text?

geoffharcourt:
LGTM.

jakecraige:
:+1: for more reasoning

seanpdoyle:
@gylaz updated the commit message

gylaz:
So, give `Address.belongs_to :user`, every time an address is updated, user record would get an updated `updated_at` column even though the record itself didn't change. This could be surprising potentially, but it might be because that's the way I'm used to looking at the db data. It would mean I'd need to look at the code, to know the relationships that touch this table.

seanpdoyle:
@gylaz how are you using the `updated_at` column in your applications?

I recall reading (although I'm not sure where) that `created_at` and `updated_at` shouldn't be used to represent domain concepts, and should be considered private to ActiveRecord::Base

@sgrif am I off base with my assumption?

derekprior:
I haven't done this by default, but I have experienced the pain of trying to retrofit caching behavior into established apps. In nearly 3 years here, I've yet to write an app that was constrained by write times so this seems a worthwhile up-front preparation.

I can see it being annoying that the timestamps on a row are updated when that row itself wasn't updated. On the other hand, my experience tells me that anything other than trivial tooling is going to require russian doll caching to get excellent performance in Rails. So I think this is just a new normal I might just have to get used to.

sgrif:
> @sgrif am I off base with my assumption?

A little. They're not considered private, but idiomatically the timestamp
is intended to reflect child data as well. There are certainly cases where
this might be problematic, but sometimes they can be worked around easily
(separating profile from users) or by having a third timestamp

geoffharcourt:
@derekprior my last company's app was definitely more constrained by write times than read times, but even if that's the exception to the rule, isn't the smarter touch processing in Rails 5 going to reduce the penalty for setting `touch: true` aggressively? (Rephrased from "aggressively touching".)

gylaz:
> how are you using the updated_at column in your applications?

I've used `updated_at` and `created_at` for debugging issues. But I can see situations in app UI where we want to show when something was created/updated.

I'm not against this, just wanted to raise a concern and see if it's valid.

geoffharcourt:
We've had this on our client project and it hasn't posed a problem. I'm :+1:

MattMSumner:
This LGTM. @seanpdoyle good to merge?

PR #378, opened 2015-12-09
Add guideline for capitalization in Swift

gfontenot:
When naming initialisms or acronyms, we should be following the
capitalization of the initial letter.

BAD:

```swift
let Id: String
let userId: String
```

GOOD:

```swift
let id: String
let userID: String
```

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -12,3 +12,5 @@ Swift
 * Group computed properties below stored properties
 * Use a blank line above and below computed properties
 * Group methods into specific extensions for each level of access control
+* When capitalizing acronyms or initialisms, follow the capitalization of the
+  first letter.
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -70,3 +70,18 @@ private extension MyClass {
 
   }
 }
+
+// MARK: Capitalization
+
+// Types begin with a capital letter
+struct DopeObject {
+  // if the first letter of an acronym is lowercase, the entire thing should
+  // be lowercase
+  let json: AnyObject
+
+  // if the first letter of an acronym is uppercase, the entire thing should
+  // be uppercase
+  static func decodeFromJSON(json: AnyObject) -> DopeObject {
+    return DopeObject(json: json)
+  }
+}


tonyd256:
:+1:

jakecraige:
Thoughts on applying this for everything instead of only swift? I've never known what to capitalize these, no matter which language I'm in

gfontenot:
@jakecraige I was _going_ to open this as "Capitalization rules for camel (or pascal or whatever) cased languages) but I had no idea where to put that. I think it's a simple/good enough rule to be cross-language though.

dgalarza:
:+1:

gabebw:
In Ruby, classes are named `HtmlParser`, not `HTMLParser`. So it might not apply everywhere.

jakecraige:
@gabebw After doing a bit of research, it seems ruby differs from rails on this. Ruby stdlib seems to captalize them in types, rails seems not to(sometimes it does though).

screen shot 2015-12-09 at 10 58 50
screen shot 2015-12-09 at 10 58 33

@gfontenot Should we update this rule to specify that it should be uppercased if part of a type name, no matter where it is?

bernerdschaefer:
Awesome. This is the best practice in Go as well:

> Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". Here's an example: ServeHTTP not ServeHttp.
>
> This rule also applies to "ID" when it is short for "identifier," so write "appID" instead of "appId".
>
> Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.
> -- https://github.com/golang/go/wiki/CodeReviewComments#initialisms

gfontenot:
@jakecraige I think that still follows the rules as laid out here:

BAD:

```swift
struct CoolJsonParser { }
struct JsonParsapalooza { }
```

GOOD:

```swift
struct CoolJSONParser { }
struct JSONParsapalooza { }
```

@gabebw I'd personally argue that `HtmlParser` is horribly ugly.

dgalarza:
So based off:

> GOOD:

> struct CoolJSONParser { }
> struct JSONParsapalooza { }

and this conversation - https://github.com/thoughtbot/publius-ios/pull/48#discussion-diff-47122839

Would it be `URLString` or `urlString`?

gfontenot:
@dgalarza `urlString`, `URLString` would imply a type.

This is admittedly at odds with the way Cocoa does things, but I don't think that should stop us from striving to be better.

dgalarza:
Ah makes sense. Thanks!

jferris:
> I'd personally argue that `HtmlParser` is horribly ugly.

How do you feel about [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)?

gfontenot:
@jferris

kaishin:
I'm very much in favor of this. 

@gfontenot The commit message implies a lot more than acronym capitalization, so maybe change that?

PR #379, opened 2015-12-17
Add default SCSS-Lint configuration

tysongach:
Ever started working on a project with Hound only to have it bark at things our Sass style guidelines, or your own guidelines (BEM? Or not to BEM?), . Yeah, me too.

This PR adds a “default” SCSS-Lint configuration file, which is based directly off of [SCSS-Lint’s default config](https://github.com/brigade/scss-lint/blob/master/config/default.yml), but aligns to our established, team-wide guides.

It does _not_:
- Enforce any particular class naming structure (`SelectorFormat`), e.g. BEM. This is a personal/project team decision that should be made collaboratively on a per-project basis.
- Enforce _either_ lowercase or uppercase hexadecimal colors (`HexNotation`). Again, we haven’t come to a consensus as a team on this, so it’s one to toggle on a set a style on a per-project basis.

---

One difficult thing I’ve always come across with SCSS-Lint is not just sharing config files (each project has slightly different needs), but also knowing which _version_ of SCSS-Lint a particular config aligns to. So I’ve put a comment at the top of this one here which indicates so.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -6,3 +6,6 @@ jscs:
 ruby:
   enabled: true
   config_file: style/ruby/.rubocop.yml
+scss:
+  enabled: true
+  config_file: style/sass/.scss-lint.yml
--- a/style/sass/.scss-lint.yml
+++ b/style/sass/.scss-lint.yml
@@ -1,0 +1,237 @@
+# Up-to-date with SCSS-Lint v0.43.2
+
+scss_files: "**/*.scss"
+
+linters:
+  BangFormat:
+    enabled: true
+    space_before_bang: true
+    space_after_bang: false
+
+  BemDepth:
+    enabled: false
+    max_elements: 1
+
+  BorderZero:
+    enabled: true
+    convention: zero
+
+  ChainedClasses:
+    enabled: false
+
+  ColorKeyword:
+    enabled: true
+
+  ColorVariable:
+    enabled: true
+
+  Comment:
+    enabled: true
+    style: silent
+
+  DebugStatement:
+    enabled: true
+
+  DeclarationOrder:
+    enabled: true
+
+  DisableLinterReason:
+    enabled: false
+
+  DuplicateProperty:
+    enabled: true
+
+  ElsePlacement:
+    enabled: true
+    style: same_line
+
+  EmptyLineBetweenBlocks:
+    enabled: true
+    ignore_single_line_blocks: true
+
+  EmptyRule:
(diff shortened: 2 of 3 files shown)


tysongach:
It’s possibly a good idea to just put this straight into Suspenders (also, or maybe only there?)…

joshuaogle:
@tysongach You're right, that might be better. And maybe link to it from here for reference?

wardpenney:
Thank you @tysongach! :+1:

teoljungberg:
Hound can take a URL to a config file, so it might make sense to consolidate all of our linter configurations here. And then, change the default `.hound.yml` in [thoughtbot/suspenders](/thoughtbot/suspenders/) to point to those.

I'm in favor of this!

This means that we can deprecate a few of the thoughtbot specific behaviors in Hound itself. /cc @thoughtbot/hound

style/sass/scss-lint-config.yml
```diff
@@ -0,0 +1,237 @@
+# Up-to-date with SCSS-Lint v0.43.2
```
gylaz: How likely is it that that this version will grow stale? Is the idea here to bump it when Hound supports newer versions?

tysongach: @gylaz SCSS-Lint is pretty actively developed and [sees a lot of updates](https://github.com/brigade/scss-lint/releases). Hound is a key part of all of this, but I also run SCSS-Lint locally from my command line as I develop to double check style on my current work.

Since a new release from SCSS-Lint is likely to introduce to options and linters, having this comment right in the config file for a project is a nice quick-check to see what version it's for.

kaishin:
This looks good to me.

tysongach:
Just dropping a note here to link this to Hound, which should be enabled when https://github.com/thoughtbot/guides/pull/389 gets merged.

- [ ] Link `scss-lint-config.yml` to Hound

seanpdoyle:
@tysongach #389 has been merged!

tysongach:
Thanks all, for the review and discussion around whether this should belong here or in Suspenders.

I’m going to go ahead and merge as-is, here in Guides, as we now have Hound running within this repo (per #389), so we can turn on and point Hound to this specific SCSS-Lint config.

Also, per Teo’s comment, we could just point Hound in Suspenders to this config by passing it the URL here. I made a [new issue over there](https://github.com/thoughtbot/suspenders/issues/695) to look into that.

PR #384, opened 2016-01-07
Use an ORDER BY clause for user-visible queries

jferris:
> If sorting is not chosen, the rows will be returned in an unspecified
> order. The actual order in that case will depend on the scan and join
> plan types and the order on disk, but it must not be relied on. A
> particular output ordering can only be guaranteed if the sort step is
> explicitly chosen.

http://www.postgresql.org/docs/8.3/static/queries-order.html

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -165,6 +165,9 @@ Postgres
 * Consider a [partial index] for queries on booleans.
 * Constrain most columns as [`NOT NULL`].
 * [Index foreign keys].
+* Use an `ORDER BY` clause on queries where the results will be displayed to a
+  user, as queries without one may return results in a changing, arbitrary
+  order.
 
 [`NOT NULL`]: http://www.postgresql.org/docs/9.1/static/ddl-constraints.html#AEN2444
 [combines multiple indexes]: http://www.postgresql.org/docs/9.1/static/indexes-bitmap-scans.html


mcmire:
I always suspected that there were something funky going on how Postgres ordered results by default, but I didn't actually realize that it was explicitly documented. This makes sense to me. :+1:

thorncp:
:+1: The side effect of triggering me to think about _what_ to order by sounds good as well.

calleerlandsson:
Looks good to me!

best-practices/README.md
```diff
@@ -161,6 +161,8 @@ Postgres
 * Consider a [partial index] for queries on booleans.
 * Constrain most columns as [`NOT NULL`].
 * [Index foreign keys].
+* Use an `ORDER BY` clause on queries where the results will be displayed to a
+  user, as queries without one may return results in changing, arbitrary order.
```
mike-burns: This feels like it needs an article: `... queries without one may return results in a changing, arbitrary order`.

mike-burns:
I'm in favor of this.

MattMSumner:
I like this :+1:

best-practices/README.md
jferris: You mean before the second use of "results?" How about "those results?"

mike-burns: I meant before "changing". I could take or leave "those results" -- I think just "results" is more clear.

jferris: Ah, you mean like, "in a changing, arbitrary order?"

mike-burns: Exactly.

drapergeek:
I agree completely with this.

PR #386, opened 2016-01-08
Trailing comma for object literals and arrays

MattMSumner:
Why:

* We'd like to explicitly encourage the use of a trailing comma when
  defining multi line JavaScript object literals or arrays.

This PR:

* Adds a guideline for using trailing commas and an example.

### DIFF
--- a/style/javascript/README.md
+++ b/style/javascript/README.md
@@ -19,6 +19,9 @@ JavaScript
 * Use `const` for declaring variables that will never be re-assigned, and `let`
   otherwise.
 * Avoid `var` to declare variables.
+* Use a [trailing comma] after each item in a multi-line array or object
+  literal, including the last item.
 
 [template strings]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings
 [arrow functions]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions
+[trailing comma]: /style/javascript/sample.js#L11
--- a/style/javascript/sample.js
+++ b/style/javascript/sample.js
@@ -5,3 +5,8 @@ class Cat {
     return false;
   }
 }
+
+const somePerson = {
+  name: 'Ralph',
+  company: 'thoughtbot',
+};


mcmire:
:+1:

seanpdoyle:
:clap: thanks @houndci!

mike-burns:
The JSON specification does not allow this, and Ruby chokes on this:

```
irb(main):029:0> puts j
{
"a": 1,
"b": 2,
}
=> nil
irb(main):030:0> puts j2
{
"a": 1,
"b": 2
}
=> nil
irb(main):031:0> JSON.parse(j)
JSON::ParserError: 757: unexpected token at '{
"a": 1,
"b": 2,
}
'
	from /home/mike/.gem/ruby/2.2.2/gems/json-1.8.3/lib/json/common.rb:155:in `parse'
	from /home/mike/.gem/ruby/2.2.2/gems/json-1.8.3/lib/json/common.rb:155:in `parse'
	from (irb):31
	from /home/mike/.rubies/ruby-2.2.2/bin/irb:11:in `<main>'
irb(main):032:0> JSON.parse(j2)
=> {"a"=>1, "b"=>2}
```

Trailing commas are valid ES5 & 6 but not valid JSON. Is that something we should be clear about here? Does having two separate style rules lead to interoperability problems?

BlakeWilliams:
:+1: 

> Does having two separate style rules lead to interoperability problems?

Fortunately Vim syntax yells when my muscle memory decides to add more commas, so I think tooling solves this problem well enough.

drapergeek:
Totally in favor of this :+1:

MattMSumner:
As this is a javascript style guide is it already clear that this only applies to ECMAScript? Should I make it clear that this doesn't work in JSON or is the fact that it chokes enough?

teoljungberg:
I would clarify that it does not work in JSON, I was bitten by this today.

tommydangerous:
We do this at Airbnb and like it. :+1:

jferris:
@MattMSumner looks like this is ready to merge?

mjankowski:
@MattMSumner ping

sikachu:
@MattMSumner :+1: #notareactioniknow

PR #387, opened 2016-01-08
Remove advice to use coffeescript

MattMSumner:
Why:

* More projects have been using JavaScript or JavaScript with a
  transpiler such as babel to get future features of the language.
* Leaving clients a project with CoffeeScript adds an additional
  barrier to finding talent comfortable working on their product.
* ES6 transpiling give most of the benefits we enjoy with CoffeeScript,
  such as classes and object destructuring, without the drawback of a
  different syntax from JavaScript.
* More developers are familiar with JavaScript then CoffeeScript.

This PR:

* Removes the "best practise" of using CoffeeScript.
* Removes CoffeeScript styles.
* Removes an "Example:" with no following example.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -203,10 +203,9 @@ Web
 
 JavaScript
 ----------
-* Use Coffeescript, ES6 with [babel], or another language that compiles to
-  JavaScript
+* Use the latest stable JavaScript syntax with a transpiler, such as [babel].
 * Include a `to_param` or `href` attribute when serializing ActiveRecord models,
-  and use that when constructing URLs client side, rather than the ID. Example:
+  and use that when constructing URLs client side, rather than the ID.
 
 [babel]: http://babeljs.io/
 


best-practices/README.md
```diff
@@ -189,10 +189,9 @@ Email
 
 JavaScript
 ----------
-* Use Coffeescript, ES6 with [babel], or another language that compiles to
-  JavaScript
+* Use JavaScript with [babel], or another language that compiles to JavaScript.
```
joshuaclayton: This becomes more ambiguous as it doesn't specify ES6/ES2015, which might be construed as continuing to use ES5. Should we retain explicitness around ES6/ES2015?

teoljungberg: I prefer not being explicit on versions, since tooling might change throughout the course of time.
What benefits do you think it brings @joshuaclayton ?

teoljungberg:
As a side note to this PR - we should make the changes to suspenders too in that case to use ES6 instead of coffeescript.

best-practices/README.md
joshuaclayton: @teoljungberg great point; I think my concern here was that it could be construed that we're encouraging vanilla JS/ES5, but if it seems clear that we should be using babel for transpiling whatever the latest version is, I'm good with this! :+1:

MattMSumner: I explicitly left out mentioning ES6 as babel already supports some ES7 features. If I know numbers, they will continue to go up.

MattMSumner:
PR adding babel to suspenders: https://github.com/thoughtbot/suspenders/pull/694

best-practices/README.md
mcmire: Technically speaking, CoffeeScript does compile to JavaScript.

Should we simply say "Use JavaScript with Babel"?

calebthompson:
:-1: I'm a huge fan of coffeescript, and recent changes to JS only incorporate the worse features of CS into JS. It's a net detriment.

calebthompson:
If anything, I'd rather see the guideline to use one or the other be removed and it left up to projects and teams.

geoffharcourt:
I think it's OK to say that preferring Coffeescript is no longer the recommended practice, but we are extremely likely to encounter it in projects that predate us. I'd rather keep the styles we've cultivated for Coffeescript than just toss them out.

BlakeWilliams:
:+1: since I don't believe we should use CoffeeScript on client projects going forward.

MattMSumner:
I split removing CoffeeScript styles into a second commit in case we wanted to keep them. I can bring them back so this is less of an aggressive change.

This is the result of the following research card: https://trello.com/c/EGpNLWB1/393-es6-instead-of-coffeescript

geoffharcourt:
@MattMSumner thanks for splitting that out.

I think there's very little cost to leaving the styles in place until/unless we are doing no work involving Coffeescript.

MattMSumner:
@geoffharcourt done

geoffharcourt:
Thanks @MattMSumner! LGTM.

gabebw:
I'm a big fan of ES6 over CoffeeScript. :+1:

mike-burns:
I'd like to second @calebthompson 's request: drop any suggestion entirely and leave it up to the team.

- JS moves too quickly for us to reasonably make a "we've been using this technique for multiple successful projects" suggestions.
- None of these technologies are wrong.
- Under Rails, we can combine these if that makes sense for the project.

sikachu:
I'm for removing the suggestion entirely as well, with the same reasoning as @mike-burns.

By the time we merge this we might already have ES7 so _who knows_.

geoffharcourt:
#minuswell

BlakeWilliams:
> By the time we merge this we might already have ES7 so who knows.

Babel would still support this since its goal is to support the newest versions of JavaScript.

I dislike the idea of removing the suggestion since it implies that ES5 is okay as well as CoffeeScript.

derekprior:
Why isn't it ES5 okay?

BlakeWilliams:
I look at it the same way that we don't use Ruby 1.8 or 1.9. It's usable but it doesn't have a lot of the features that make our daily lives nicer and it's not very forward thinking.

MattMSumner:
The change here leaves language open to use your preferred javascript transpiler, while suggesting babel if you have no preference. This means that if the team has a preference to use something else thats fine. In the end, as with all guidelines, this should be ignored if you have good reason.

This change doesn't mention ES6 for good reason and probably isn't future proof. That doesn't make me any less confident suggesting that new projects follow this guideline.

> JS moves too quickly for us to reasonably make a "we've been using this technique for multiple successful projects" suggestions.

This seems like a common argument for never doing anything. This is a guideline that will/should be changed. We shouldn't be afraid of making suggestions even if next week the suggestion is outdated. We've had a non-trivial number of recent client projects using ES6 which lead to this PR: https://trello.com/c/EGpNLWB1

> None of these technologies are wrong.

While not inherently "wrong", I've felt pain working with CoffeeScript that makes me hesitant to suggest it as a "good" default for new projects.

> Under Rails, we can combine these if that makes sense for the project.

This is true of any guideline.

Using ES5 is also fine, but using babel feels like using `scss` instead of `css`. I'm still clearly writing JavaScript but with more of the behavior I'd expect from a modern programming language.

best-practices/README.md
gylaz: Thoughts on wording it like `Use latest stable JavaScript syntax with a transpiler (like Babel)...`?

jferris:
If we remove the recommendation for any compiler/transpiler, will we also remove coffeescript and babel from Suspenders?

It looks like Rails still defaults to including Coffeescript: https://github.com/rails/rails/blob/master/railties/lib/rails/generators/app_base.rb#L313

mcmire:
#694 removes CoffeeScript from the Gemfile and replaces it with sprockets-es6 (which uses Babel internally).

MattMSumner:
That PR adds sprockets-es6 for the time being, the next version of sprockets will allow ES6 out of the box so perhaps just removing the guideline is the correct action?

best-practices/README.md
```diff
@@ -189,10 +189,9 @@ Email
 
 JavaScript
 ----------
-* Use Coffeescript, ES6 with [babel], or another language that compiles to
-  JavaScript
+* Use the latest stable JavaScript syntax with a transpiler, such as [babel].
```
mike-burns: Saying "with a transpiler" seems unnecessary; either the latest JS syntax is supported by all our browsers, in which case we don't need a compiler, or it isn't, in which case we do. This seems like a judgement that the team will have to make for the project based on the target JS runtimes.

mcmire: ES6 (as it stands) is [not completely supported by all browsers](https://kangax.github.io/compat-table/es6/), so you need a transpiler to use it.

If you had to use ES5, that'd be fine -- you could still use a transpiler and use ES6 syntax, which gets compiled to ES5 syntax.

In the future, when ES7 comes out, then we will still need a transpiler to use it, as not all browsers will support that, either.

So, I think we will need a transpiler either way.

MattMSumner:
I'm going ahead and merging this.

BlakeWilliams:
:shipit: :put_litter_in_its_place:

gilesvangruisen:
:+1:

kenyonj:
:+1:

PR #388, opened 2016-01-08
Add link to an explanation of `git rebase -i ...`

LkeMitchll:
I did not understand the `git rebase -i ...` interface at first. 

I didn't quite understand the process of 'marking' commits with keywords, as I'd never come across this interface before.

This link to a post by David Walsh really helped me to understand the process, and seems to match our way of working pretty precisely.

### DIFF
--- a/protocol/git/README.md
+++ b/protocol/git/README.md
@@ -47,8 +47,9 @@ Write a [good commit message]. Example format:
 
     http://project.management-system.com/ticket/123
 
-If you've created more than one commit, use a rebase to squash them into
-cohesive commits with good messages:
+If you've created more than one commit,
+[use `git rebase` interactively](https://help.github.com/articles/about-git-rebase/)
+to squash them into cohesive commits with good messages:
 
     git rebase -i origin/master
 


protocol/git/README.md
```diff
@@ -47,8 +47,9 @@ Write a [good commit message]. Example format:
 
     http://project.management-system.com/ticket/123
 
-If you've created more than one commit, use a rebase to squash them into
-cohesive commits with good messages:
+If you've created more than one commit, use a
```
mcmire: What about linking a whole phrase rather than part of a phrase? Also "use a rebase" sounds awkward, what do you think about rephrasing this to "use `git rebase`" instead?

So all together it would be:

> [use `git rebase`](http://some-link-here.com) to squash them into cohesive commits with good messages

protocol/git/README.md
```diff
@@ -47,8 +47,9 @@ Write a [good commit message]. Example format:
...
+[rebase to squash](https://davidwalsh.name/squash-commits-git)
```
BlakeWilliams: Thoughts on linking to git documentation directly instead? https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages

mcmire: Additionally, what about the [official GitHub docs](https://help.github.com/articles/about-git-rebase/)?

teoljungberg: If this is to target people who are less familiar with `git` and have not used `git rebase -i` before. I think the GitHub docs might be easier to understand

protocol/git/README.md
```diff
@@ -47,8 +47,8 @@ Write a [good commit message]. Example format:
 
     http://project.management-system.com/ticket/123
 
-If you've created more than one commit, use a rebase to squash them into
-cohesive commits with good messages:
+If you've created more than one commit, [use `git rebase` interactively](https://help.github.com/articles/about-git-rebase/)
```
teoljungberg: Should this be kept under 80 characters?

LkeMitchll:
What do you guys think about the phrase "use `git rebase` interactively"? Using simply "use `git rebase`" does not look like a link, and may be confusing. Only 'use' appears to be a link.

teoljungberg:
@LkeMitchll It looks like a link to me atleast - https://www.dropbox.com/s/4mgb7zp6g4lo6p3/Screenshot%202016-01-11%2010.21.47.png?dl=0

teoljungberg:
This LGTM, use your newfound skills to squash and merge at your discression.

PR #390, opened 2016-01-12
JavaScript: Disallow multiple `var` declaration

seanpdoyle:
The `disallowMultipleVarDecl` rule already exists in the `.jscsrc`.
Unfortunately, it is invalidly declared.

This commit corrects the declaration.

[rule]: http://jscs.info/rule/disallowMultipleVarDecl


teoljungberg:
:+1:

PR #391, opened 2016-01-19
Swift: Specify preference for structs over classes

gfontenot:
Also note that if you _are_ going to use a `class`, you should default
to marking it as `final`.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -5,6 +5,8 @@ Swift
 
 * Keep up with the Objective-C style guide above. Will highlight differences
   here.
+* Prefer `struct`s over `class`es wherever possible
+* Default to marking classes as `final`
 * Use `let` whenever possible to make immutable variables
 * Name all parameters in functions and enum cases
 * Use trailing closures
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -2,6 +2,17 @@
 
 import Foundation // or not
 
+// MARK: Types
+
+// Prefer structs over classes
+struct User {
+  let name: String
+}
+
+// When using classes, default to marking them as final
+final class MyViewController: UIViewController {
+}
+
 // MARK: Closures
 
 // Use typealias when closures are referenced in multiple places


gfontenot:
I'm going to add this info to the commit message, but the reasoning behind this change:

`struct`s are immutable value objects. This means that the compiler is able to perform better optimizations since it knows that the object can't change. It also reduces the cognitive overhead around mutation and state passing. `struct`s are also incapable of using inheritance, which encourages the use of `protocol`s and composition instead, which we prefer as a design pattern.

If you're going to use a `class`, marking it as `final` will ensure that it can't be subclassed. This lets the compiler make some better optimizations since it knows the methods being added won't ever be overridden. It also encourages the use of composition over inheritance, which we prefer as a design pattern.

gilesvangruisen:
:+1:

dgalarza:
:+1:

style/swift/README.md
```diff
@@ -5,6 +5,8 @@ Swift
 
 * Keep up with the Objective-C style guide above. Will highlight differences
   here.
+* Prefer `struct`s over `class`es wherever possible
+* If you need to use a `class`, default to marking it as `final`
```
jakecraige: The wording of this is sort of based off the previous rule, what about rewording it so it stands on it's own? Something like "Default to marking classes as `final`"?

gfontenot:
ping @thoughtbot/ios

seanpdoyle:
Is there a way to codify this guideline in a [SwiftLint](https://houndci.com/configuration#swift) rule?

style/swift/README.md
kaishin: Agreed.

dazmuda:
Do you think we should also mention:
Prefer protocols over classes, if possible.

It's more like... if it's possible and makes sense but also it makes sense and is possible more often than one would think.

tonyd256:
:+1: to this concept

gfontenot:
There don't seem to be any objections to this. I'm going to go ahead and merge. @dazmuda I'll open a PR today about preferring protocols to inheritance, which is I believe what you were going for?

PR #392, opened 2016-01-20
Specify when to use () and Void in func signatures

gfontenot:
We want to avoid ambiguity, especially with our return types. Functions
that return `()` can become hard to read when passed as an argument to a
higher order function. For example:

```swift
func f(g: () -> ()) { }
```

The parens for the return value end up getting lost in the closing
parens for the higher order function. Conversely, by using `Void`:

```swift
func f(g: () -> Void) { }
```

`Void` stands out as a specific type on its own, and doesn't get lost in
a mess of parens.

At the same time, `()` is a nice brief indication that a function takes
no arguments, and mimics the look of an empty tuple. So if you declare
these two functions:

```swift
func fa(g: () -> Void) { }
func fb(g: (a: T) -> Void) { }
```

There is a visual similarity between the two functions.

This is also how Apple prefers to style functions taking or returning
`Void`.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -16,3 +16,5 @@ Swift
 * Group methods into specific extensions for each level of access control
 * When capitalizing acronyms or initialisms, follow the capitalization of the
   first letter.
+* When using `Void` in function signatures, prefer `()` for arguments and
+  `Void` for return types.
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -95,6 +95,9 @@ struct DopeObject {
   }
 }
 
+// Use () for void arguments and Void for void return types
+let f: () -> Void = { }
+
 // MARK: Guards
 
 // One expression to evaluate and short or no return


tonyd256:
I'm good with this as a standard.

dazmuda:
I'm into this. Call me a necromancer :sparkles: cause I like returning to the voiiiiiiid. :crystal_ball:

tonyd256:
@dazmuda WHEN Github supports likes ... I'm coming back here to like that comment!

gfontenot:
Ping @thoughtbot/ios

seanpdoyle:
Is there a way to codify this rule in a [SwiftLint](https://houndci.com/configuration#swift)?

klaaspieter:
100% agree with this.

@seanpdoyle A quick look at some [implementations](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/FunctionBodyLengthRule.swift) of [existing rules](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules) suggest that their `ASTRule` protocol receives enough information to detect this. I think it's still non-trivial, but definitely possible.

kaishin:
I've been using `Void -> Void` recently but this is not bad either.

dgalarza:
:+1:

jakecraige:
> I've been using `Void -> Void` recently but this is not bad either.

I kind of like the consistency of that. `() -> ()` wouldn't be too bad either but I do like the explicitness of actually writing `Void`. I've never really liked `()` personally.

gfontenot:
@kaishin @jakecraige I agree that the consistency is key. The argument I was attempting to make is that using `()` for `Void` arguments is that it is more consistent since it mimics function declaration (`func foo()`), as well as closure syntax with multiple arguments (`(Foo, Bar) -> Baz` -> `() -> Void`)

PR #393, opened 2016-01-20
Remove callout to Objective-C styleguide in Swift

gfontenot:
We aren't following many (any?) of the guidelines from Objective-C.
There's no real reason to bring attention to it anymore. We should
instead look ahead and add guidelines for Swift without the legacy
baggage.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -3,8 +3,6 @@ Swift
 
 [Sample](sample.swift)
 
-* Keep up with the Objective-C style guide above. Will highlight differences
-  here.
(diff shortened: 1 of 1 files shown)


tonyd256:
:+1: Yup, 100% agree

jakecraige:
:+1:

gfontenot:
ping @thoughtbot/ios

dazmuda:
Removing the callout makes sense. Good to keep them around though. :+1:

kaishin:
Fair enough.

gfontenot:
No objections here. Merging.

PR #394, opened 2016-01-20
Guard against bad `guard` syntax

tonyd256:
Suggest formatting for dealing with `guard` syntax in Swift.

### DIFF
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -94,3 +94,23 @@ struct DopeObject {
     return DopeObject(json: json)
   }
 }
+
+// MARK: Guards
+
+// One expression to evaluate and short or no return
+
+guard let singleTest = somethingFailable() else { return }
+guard statementThatShouldBeTrue else { return }
+
+// If there is one long expression to guard or multiple expressions
+// move else to next line
+
+guard let oneItem = somethingFailable(),
+  let secondItem = somethingFailable2()
+  else { return }
+
+// If the return in else is long, move to next line
+
+guard let something = somethingFailable() else {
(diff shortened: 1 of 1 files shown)


gfontenot:
ping @thoughtbot/ios

gfontenot:
I think this makes sense.

jakecraige:
Should we also add this to the `README.md` so that people scanning the readme can see this rule exists?

seanpdoyle:
Is there a way to codify this guideline in a [SwiftLint](https://houndci.com/configuration#swift) rule?

kaishin:
Looking good.

dgalarza:
I agree with @jakecraige, is there some way we can link to this in the `README.md`? I feel like people may not notice this otherwise.

dazmuda:
That second case still looks so weird to me but it's the best there is! Ya!

tonyd256:
Xcode wants to indent guards like that but we could do it this way instead if people like:

```
guard let oneItem = somethingFailable(),
  let secondItem = somethingFailable2()
else { return }
```

dazmuda:
Naw man, Xcode's spacing FTW simply because manually managing spacing is a fool's errand.

tonyd256:
@gfontenot if there are no more comments on this tomorrow feel free to merge it in my stead.

gfontenot:
@jakecraige @dgalarza I think this falls under the "break long lines at 100 characters" rule. Maybe we could just add a `// MARK:` section to the sample that has style recommendations for how to break up long lines?

dgalarza:
I like that idea @gfontenot.

gfontenot:
I'll merge this then, and open a new PR to reorganize the sample.

PR #395, opened 2016-01-21
Add line length limit for Swift

gfontenot:
We should probably have _some_ guideline for line length in Swift. 80
feels too short, since we tend to work with fairly verbose APIs.
SwiftLint has settled on 100 characters, which seems like a reasonable
length.

In Xcode you can add a line length guide inside Preferences -> Text
Editing.

ping @thoughtbot/ios

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -5,6 +5,7 @@ Swift
 
 * Prefer `struct`s over `class`es wherever possible
 * Default to marking classes as `final`
+* Break long lines after 100 characters
 * Use `let` whenever possible to make immutable variables
 * Name all parameters in functions and enum cases
 * Use trailing closures


style/swift/README.md
```diff
@@ -5,6 +5,7 @@ Swift
 
 * Keep up with the Objective-C style guide above. Will highlight differences
   here.
+* Limit lines to 100 characters
```
gylaz: Curious why this is different that our general [rule](https://github.com/thoughtbot/guides/tree/master/style#formatting):

> Break long lines after 80 characters.

tonyd256: We work with a lot of older API from ObjC that have very verbose names. Sometimes this is unrealistic.

tonyd256:
I do find the hound messages annoying ... but this is probably a good rule to have

style/swift/README.md
gfontenot: @gylaz I tried to reference that in the PR message/commit message. As I noted and as Tony said, 80 characters is unrealistic most of the time because of the verbosity of the APIs we're working with.

jakecraige:
:+1:

style/swift/README.md
gylaz: Ah, missed it. Thanks for the explanation!

seanpdoyle:
Can we codify this guideline with a [SwiftLint](https://houndci.com/configuration#swift) rule?

kaishin:
@seanpdoyle Yes, as the previous comments have mentioned.

I've been using 80 as a target but I broke it occasionally. 100 is generous enough.

dgalarza:
:+1:

BlakeWilliams:
:+1: rustfmt has the same rule to deal with verbosity.

dazmuda:
100 seems good. Current project doesn't have many lines over that.

Fun fact: 100 char guide line, menlo 14, 12in macbook screen = a good two inches of project nav sidebar room if you're into that. Baby laptop approved! :+1:

gfontenot:
@seanpdoyle It's already a rule in Hound.

gfontenot:
No objections, merging.

PR #396, opened 2016-01-21
Use 2 spaces for indentation in Swift

gfontenot:
This brings Swift inline with the defaults for Ruby and other languages.
The benefit here is that while working outside of Xcode (as some of us
try to do), we can use a consistent setting for indentation instead of
having to special case Swift.

From a visual aspect, I've gotten used to the look of 2 spaces, and I
think it feels nice to move away from the 4 space indentations of
Objective-C.

A few of our projects (internal and external) are already using 2
spaces, and I think in _general_ people have liked it more.

Note that this has already been proposed, but was shot down because we
couldn't come to a real consensus. I'd like to re-poll the team and see
where we stand now.

ping @thoughtbot/ios

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -6,6 +6,7 @@ Swift
 * Prefer `struct`s over `class`es wherever possible
 * Default to marking classes as `final`
 * Break long lines after 100 characters
+* Use 2 spaces for indentation
 * Use `let` whenever possible to make immutable variables
 * Name all parameters in functions and enum cases
 * Use trailing closures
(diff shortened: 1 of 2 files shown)


jakecraige:
I'm for it :+1:

tonyd256:
:+1:

seanpdoyle:
Could we codify this guideline in a [SwiftLint](https://houndci.com/configuration#swift) Hound rule?

klaaspieter:
:thumbsup:

kaishin:
I have been using 2 spaces in all my new Swift projects since day one, so needless to say I am for this change.

dgalarza:
:+1:

gfontenot:
@seanpdoyle I'm don't know for sure, but I don't think the AST has enough info to know this...

No objections to this, merging.

PR #397, opened 2016-01-21
Update for latest Swift 2+

tonyd256:
Since this was written, `println` changed to `print`.

### DIFF
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -31,13 +31,13 @@ var maybe: Bool?
 
 // Use if-let syntax to unwrap optionals
 if let definitely = maybe {
-    println("This is \(definitely) here")
+    print("This is \(definitely) here")
 }
 
 // If the API you are using has implicit unwrapping you should still use if-let
 func someUnauditedAPI(thing: String!) {
     if let string = thing {
-        println(string)
+        print(string)
     }
 }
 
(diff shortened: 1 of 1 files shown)


style/swift/sample.swift
```diff
@@ -31,16 +31,19 @@ var maybe: Bool?
...
 // If the API you are using has implicit unwrapping you should still use if-let
 func someUnauditedAPI(thing: String!) {
     if let string = thing {
-        println(string)
+        print(string)
     }
 }
 
+// You can also use the `?` postfix operator to chain possible execution
+delegate?.doSomething()
```
jakecraige: Is this a guideline? This seems like it's just mentioning a language feature

gfontenot: Do we need to call this out? This feels like trying to teach the language, rather than specifying a style guideline.

tonyd256: This section says to use `if-let` to unwrap optionals before use. I thought it would be appropriate to mention that chaining things if also ok ...

gfontenot: It specifically says that you should use `if let` even if you're dealing with an implicitly unwrapped optional. It's not talking about using `if let` generally.

tonyd256: That's what I understood from line 32

gfontenot: Oh, I see. I'd be more in favor of removing the weirder non-guidelines that boil down to "Use language features" in a separate PR. I don't think this kind of thing is going to scale. The guideline _should_ be "Avoid force-unwrapping optionals" so maybe we just change it to that?

tonyd256: Gotcha, ya I think that's fine. I'll just remove this from this PR and it can be handled in another.

jakecraige:
:+1:

gfontenot:
:+1:

PR #398, opened 2016-01-22
Advocate using locals to render views

mcmire:
It has [long] been [argued] that using instance variables to pass values
to Rails views isn't the greatest of ideas: Rails twists instance
variables to do something they aren't supposed to do, and if you think
about it, it's unclear how views even have access to variables that live
in a completely different scope. (Hint: they're copied from the
controller to the view at runtime when the view is instantiated.)

In most other languages, views (or templates, as they're sometimes
called) are essentially functions and, as such, when you're rendering a
template and want to give that template access to certain values, you
just pass those values along as arguments.

Is there a way to do this in Rails? Yes, by calling `render` explicitly
and using the `locals` option. So your controller would go from this:

``` ruby
class PostsController < ApplicationController
  def index
    @posts = Post.all
  end
end
```

to this:

``` ruby
class PostsController < ApplicationController
  def index
    posts = Post.all
    render locals: { posts: posts }
  end
end
```

The upside to this approach is that now, views and partials act the same
in terms of how they reference variables, and you can convert a view to
a partial easily.

The downside is that you can no longer easily assert that a view was
rendered with a particular variable, but that's okay, because we don't
write controller specs anyway.

[long]: https://github.com/hashrocket/decent_exposure
[argued]: http://thepugautomatic.com/2013/05/locals/

### DIFF
--- a/style/rails/sample.rb
+++ b/style/rails/sample.rb
@@ -1,7 +1,15 @@
-class SomeClass
+class SomeModel < ActiveRecord::Base
+  belongs_to :tree
   belongs_to :tree, class_name: Plant
   has_many :apples
   has_many :watermelons
 
   validates :name, presence: true, uniqueness: true
+end
+
+class PostsController < ActionController::Base
+  def index
+    posts = Post.all
+    render locals: { posts: posts }
+  end
 end
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -21,10 +21,14 @@ Rails
 * Use `link_to` for GET requests, and `button_to` for other HTTP verbs.
 * Use new-style `validates :name, presence: true` validations, and put all
   validations for a given column together. [Example][validations].
+* Avoid setting instance variables in controllers; explicitly `render`
+  views with a `locals` option to pass values to views instead.
+  [Example][rendering-views-with-locals].
 
 [order-associations]: /style/rails/sample.rb#L2-L4
 [validations]: /style/rails/sample.rb#L6
 [`app/views/application`]: http://asciicasts.com/episodes/269-template-inheritance
+[rendering-views-with-locals]: /style/rails/sample.rb#L10
 
 Migrations
 ----------


style/rails/sample.rb
```diff
@@ -1,7 +1,15 @@
-class SomeClass
+class SomeModel < ActiveRecord::Base
+  belongs_to :tree
```
gylaz: Is this unrelated?

gylaz:
:+1: :100:

style/rails/sample.rb
mcmire: Yeah, it is. I can put it in another commit.

derekprior:
I like this in theory but I feel like it's also a bit too far afield from Rails norms. Doing this an established client project would surely bring some conflict and a developer taking over one of our greenfield projects would likely be confused by this practice. 

I don't think this solves a big enough problem to justify so drastically veering from Rails norms.

We should probably try this on a few different types of projects if we want to consider making it part of guides. I haven't seen this in our code yet.

tommydangerous:
This is interesting, I see it done in the `administrate` gem.

mcmire:
> We should probably try this on a few different types of projects if we want to consider making it part of guides.

Sure. There is a [research card](https://trello.com/c/49MHwpol) open on this where people can try it out.

jferris:
I'm going to close this while we're running the experiment. We can re-open if the experiment is successful.

mcmire:
Gotcha, just wanted to get more eyes on it.

Drenmi:
> The downside is that you can no longer easily assert that a view was
rendered with a particular variable, but that's okay, because we don't
write controller specs anyway.

For what it's worth, it looks like [`assigns` will be deprecated in Rails 5](https://github.com/rails/rails/issues/18950).

jferris:
I just noticed this in an earlier comment:

> We don't write controller specs anyway.

There isn't anything in our guides that says not to write controller specs, and I use them reasonably often. Unless we come to a consensus that controller specs are not worth supporting, I wouldn't make other changes that assume that nobody is writing them.

gylaz:
> We don't write controller specs anyway.

Is this a reason to use `render locals...`? Is there a way to still access (in tests) these variables? I feel like there must be, especially if `assigns` is deprecated in Rails 5.

jferris:
> Is this a reason to use `render locals...`? Is there a way to still access (in tests) these variables? I feel like there must be, especially if `assigns` is deprecated in Rails 5.

I don't think I've recently written a test that makes an assertion on `assigns`. I test inputs and outputs to controllers in controller specs. I usually test side effects like sending emails. When testing responses, I generally just check whether a certain template was rendered or an action redirected.

dideler:
Curious to hear if the experiment is successful. This is a change I've been personally wanting to do but have been cautious for the same reasons mentioned in https://github.com/thoughtbot/guides/pull/398#issuecomment-174331194.

PR #399, opened 2016-01-26
Recommend against `toggleProperty`

drapergeek:
Using `toggleProperty` makes tracking down interactions a lot harder. Using explicit actions for what you expect to happens make code easier to maintain and debug.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -321,6 +321,7 @@ Ember
 * Prefer dependency injection through `Ember.inject` over initializers, globals
   on window, or namespaces. ([sample][inject])
 * Prefer sub-routes over maintaining state.
+* Prefer explicit setting of boolean properties over `toggleProperty`.
 * Prefer testing your application with [QUnit][ember-test-guides].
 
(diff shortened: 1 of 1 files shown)


BlakeWilliams:
I've run into this a lot in the past, :+1:

seanpdoyle:
In favor :+1:

best-practices/README.md
```diff
@@ -322,6 +322,7 @@ Ember
 * Prefer dependency injection through `Ember.inject` over initializers, globals
   on window, or namespaces. ([sample][inject])
 * Prefer sub-routes over maintaining state.
+* Prefer explicit setting of boolean properties over `toggleProperty`.
```
mike-burns: Should this be "use" instead of "prefer"?

MattMSumner:
Agreed :+1:

best-practices/README.md
drapergeek: From my understanding of the explanations, this is a better option and you should watch out for the alternative. There are cases that using `toggleProperty` is a better option but not the normal case.

jferris:
@drapergeek seems like this is good to merge?

PR #402, opened 2016-02-09
Add best practices about Service Objects

gylaz:
We're seeing Service Objects more in the community,
and starting to use them more in our apps.
This is an attempt at getting some guidelines for consistency.

These best practices were pulled from a variety of blog posts, and
backed by experience of using them successfully on several projects.

Related blog posts:

* http://brewhouse.io/blog/2014/04/30/gourmet-service-objects.html
* http://code.tutsplus.com/tutorials/service-objects-with-rails-using-aldous--cms-23689
* https://www.netguru.co/blog/service-objects-in-rails-will-help

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -101,6 +101,10 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryGirl factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
+* Name service classes after the action they perform
+(e.g. `CreatePurchase`, `UpdateCreditCard`).
+* Expose one public method on services that does the work
+(i.e. `run`, `call`, or `perform`).
 
 [date-block]: /best-practices/samples/ruby.rb#L10
 [fkey]: http://robots.thoughtbot.com/referential-integrity-with-foreign-keys


tommydangerous:
This is a great guide addition. Here at Airbnb we do something similar.

best-practices/README.md
```diff
@@ -101,6 +101,10 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryGirl factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
+* Name service classes after the action they perform
+(e.g. `CreatePurchase`, `UpdateCreditCard`).
+* Expose one public method on services that does the work
```
bernerdschaefer: Some things it might be good to clarify:

* Should it be the *only* public method?
* Should it be a class method? Or an instance method?
* If it's an instance method, should the initializer accept arguments or state?

best-practices/README.md
```diff
@@ -101,6 +101,10 @@ Rails
...
+(i.e. `run`, `call`, or `perform`).
```
bernerdschaefer: Is "i.e." intended to convey that it should be called one of these names, and not something else? Or are these only examples?

Could we agree on a single name for this?

How do we feel about `UpdateCreditCard.update()`-style naming?

gylaz: My intent was to narrow down to a handful of method names. `call` might be the best because that is also what Ruby lambda's use, thus a service can be stubbed with a lambda. I've used `run` often in the past -- as in running a service.

best-practices/README.md
gylaz: > Should it be the only public method?

Preferably yes (not counting `initialize`), to follow the Command pattern.

> Should it be a class method? Or an instance method?

Instance method, since we prefer to program to objects. A class convenience method can be added as well, per preference.

> If it's an instance method, should the initializer accept arguments or state?

Yes, I believe the state should be passed to the initializer.

@bernerdschaefer I'll re-word to clarify some of these, if others agree.

best-practices/README.md
```diff
@@ -101,6 +101,10 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryGirl factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
+* Name service classes after the action they perform
```
calleerlandsson: Are these service classes the same as "Method Objects": https://sourcemaking.com/refactoring/replace-method-with-method-object

best-practices/README.md
calleerlandsson: I'm a big fan of `call` for the same reason @gylaz mentions.

best-practices/README.md
gylaz: Pretty much yes, just a different naming concept. Both seem to utilize the [Command Pattern](https://en.wikipedia.org/wiki/Command_pattern).

derekprior: Can someone define a "service class"? Are we talking about a command object? Or are we more broadly speaking of classes that are not rails framewrok related?

joshuaclayton: I think that distinction is key, especially given some of what I've seen on other recent client projects, where almost everything is put in `app/services`.

Unrelated to the naming/purpose of this, do we have any opinions on whether they should (or should not) allow for side-effects? Does the presence of side-effects impact what we name these?

jferris: I'm also  confused by the naming here. I've personally avoided saying "service objects" lately as nobody seems  to agree on what they mean. I used to use "service object" to describe any object that encapsulates high level business logic and is more or less decoupled from the Rails stack.

As described in this pull request, these do sound like Method Objects, which are well-described in a number of places and have been around for a long time. On the other hand, a search for "service objects" find no results on sites like c2wiki or refactoring.com, and the various blog posts describing them are generally describing different things.

Thoughts on using "method object" instead?

gylaz: > ... as nobody seems to agree on what they mean.

I think we could attempt to establish what they mean and agree on some guidelines. Could we use [Service Layer](http://martinfowler.com/eaaCatalog/serviceLayer.html) for some ideas?

> Thoughts on using "method object" instead?

Would these go under `app/methods`, or would we just describe service objects in terms of method objects? I haven't seen `app/methods` out in the wild. The purpose of this PR was focused solely on getting some conventions/consistency around service objects that have been showing up in codebases in the last couple of years.

calleerlandsson: In the projects I have been on, we've put method objects in `app/models` if they are a part of the applications business logic or `app/lib` if they're not.

jkrmr: @calleerlandsson I like that rule—I could see it encouraging more generality from objects `lib`. 

I also think there are several ways for a class to be "a part of the application's business logic/domain model." The rule of thumb I've been using is to distinguish between encapsulating and coordinating classes: i.e., between those primarily meant to encapsulate business logic (models) and those primarily meant to coordinate the behavior of models (services) (in order to unburden controllers so they can stay focused on the request/response cycle). 

On way that rule of thumb manifests is that objects of the latter category tend to have more external dependencies, which is one reason I like to group them in a separate directory (`app/services`) -- the difference between files in `models` and `services` ends up being a visual cue reminding you to minimize dependencies when working with model logic.

jferris: I don't want to derail this pull request by talking about Method Objects. I've had the same confusion about the `app/services` directory, as there hasn't been much cohesion in "services" within or between projects I've joined, so I think it's good for us to try and firm up this directory. I think your current proposal would definitely introduce some needed structure.

However, I have two concerns about the naming proposed as-is:

* The "services" name seems fairly generic and is used in so many ways that I worry it will be impossible to ever have a group understanding of what "service objects" are, similar to "presenters."
* The current changes seem to very accurately describe a Method Object, and it seems weird to have two names for the same pattern, providing a standard directory for one name.

The way you're describing them reminds me of [Transaction scripts](http://martinfowler.com/eaaCatalog/transactionScript.html). Maybe "transactions" would be a more standard name that would be harder to misinterpret?

bf4:
I also like http://multithreaded.stitchfix.com/blog/2015/06/02/anatomy-of-service-objects-in-rails/ a lot

best-practices/README.md
mike-burns: I'm not sure that this bullet point makes sense as a separate bullet point -- it does not describe a thing that is true about Rails best practices, but instead a method object best ractice.

best-practices/README.md
mike-burns: I often use `#call`, or `#perform` if I'm interacting with a library that wants that (DJ, IIRC), or `#run` if that's the one that makes sense. Not totally sure we need to specify one of these.

mike-burns:
I'm not entirely convinced that we need a guideline on how to use a method object pattern, nor that we should be so rigorous about it.

I can definitely understand us having a statement about the `app/services` directory and what conceptual ideas to put there.

gylaz:
I'm going to close this. There doesn't seem to be a strong consensus for having this in our guides. It can still be of use as reference in the future.

PR #403, opened 2016-02-09
Break up long initializers and functions

gfontenot:
Instead of letting function calls sprawl out, it's more readable to
break them up by adding linebreaks at the argument labels. As an added
benefit, it makes it easy to visually scan the implementation to see
which value is being passed to which argument.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -25,3 +25,5 @@ Swift
 * Prefer to name `IBAction` and target/action methods using a verb describing
   the action it will trigger, instead of the user action
   (e.g., `edit:` instead of `editTapped:`)
+* Break up long initializers or function calls by adding linebreaks between
+  arguments.
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -100,3 +100,13 @@ guard let oneItem = somethingFailable(),
 guard let something = somethingFailable() else {
   return someFunctionThatDoesSomethingInManyWordsOrLines()
 }
+
+// Break up long function calls and initializers
+let foo = Foo(
+  bar: Bar(),
+  baz: Baz(),
+  quz: Quz(
+    x: 42,
+    y: 365
+  )
+)


gfontenot:
ping @thoughtbot/ios

seanpdoyle:
@gfontenot is there a way to enforce this via Swift linting?

gfontenot:
@seanpdoyle I don't believe there's one built in. We'd have to add it to SwiftLint.

tonyd256:
I would probably suggest this:
```swift
let foo = Foo(bar: Bar(),
              baz: Baz(),
              quz: Quz(x: 42, y: 365))
```

but maybe that's Intel style guide getting the best of me.

Anyways, this one concerns me less than how to arrange a long function signature.

gfontenot:
I don't like that because the first argument and the last paren look completely lost to me. Giving each argument its own line seems more consistent, and giving the trailing paren its own line reduces churn if new arguments are added.

Splitting up the `Quz` initializer was purely illustrative. I'd probably do that one as a single line, but I wanted to show the indentation and style for embedded objects that are being initialized.

tonyd256:
I disagree ... but I actually care less haha so :+1: from me

style/swift/sample.swift
```diff
@@ -85,3 +85,13 @@ struct DopeObject {
...
 }
+
+// Break up long function calls and initializers
+let foo = Foo(
+  bar: Bar(),
+  baz: Baz(),
+  quz: Quz(
+    x: 42,
+    y: 365
+  )
```
jakecraige: I'm wondering if this example, while illustrative, might come across as a guideline in itself that you should split up arguments onto new lines even if they are short.

What about making `Quz` take more arguments so it matches up with the rule of breaking up arguments, or removing it completely?

sidraval:
I think I agree with tony here, but I don't feel all that strongly either.

sharplet:
> giving the trailing paren its own line reduces churn if new arguments are added.

I'm not actually sure this is true: even with the closing paren on its own line, you'll still have to edit the previous line to add a comma for the new parameter.

style/swift/sample.swift
kaishin: Always struggled with knowing what to do with the closing bracket here. I sort of lean towards keeping it in the same line as the last argument:

```
let foo = Foo(
  bar: bar,
  baz: baz)
```

klaaspieter:
I prefer the closing parent to be on a separate line because it's easier to see where something starts and ends. I think it's also common practice in Ruby and Javascript to do this:

```
someFunction({
  "key": "value",
  "key2": [
    "value1",
    "value2"    
  ]
})
```

It nicely brings the indentation back to where it started. And it's easy to see where `key2` starts and ends.

style/swift/README.md
```diff
@@ -14,3 +14,5 @@ Swift
 * Group methods into specific extensions for each level of access control
 * When capitalizing acronyms or initialisms, follow the capitalization of the
   first letter.
+* Break up long initializers or function calls by adding linebreaks between
```
kaishin: We typically start our guidelines with "Use", "Prefer", "Avoid", etc to better communicate how strict the guideline is. I have noticed that the Swift ones don't follow this pattern and I believe it'd be better if they did. For instance, it'd be:
"Prefer breaking up.." instead of "Break up...".

gfontenot: Good point. I'm planning on working on this document a bit tomorrow for investment time. I'll make a pass at the guidelines to fix the wording then.

dazmuda:
Given that we omit the first argument name when calling a function, I don't think we should push the first argument to the line after the function name.

Given that initializing looks the same but has the first argument name, I don't think we should push the first argument to the line after so the guideline will be the same for both.

gfontenot:
@sharplet 
> I'm not actually sure this is true: even with the closing paren on its own line, you'll still have to edit the previous line to add a comma for the new parameter.

Yeah, that was a bit of a straw-man argument, I apologize. What I _meant_ was that you don't have to modify the line declaring the end of the function/collection, and instead can focus on modifying the arguments.

@dazmuda 
Good points. For me, I've been leaning more and more towards explicitly naming the first argument _instead_ of adding the argument label to the function name. So `doThing(foo:bar:)` instead of `doThingWithFoo(_:bar:)`. I have this weird feeling that I saw this recommended somewhere official, but I might totally be wrong about that.

Anyway, in _that_ case, I think breaking up the arguments the way I'm proposing makes more sense for the same consistency reasons you're pointing out. But even without that caveat, I personally just find this wacky looking:

```swift
doThingWithFoo(Foo(),
  bar: Bar()
)
```

To my eye, that makes it look like there's only one argument. The first argument just kinda gets lost for me.

I think @klaaspieter articulated what I wanted to say with regards to the closing brace/bracket/paren being on a new line vs the last line better than I could have.

style/swift/sample.swift
gfontenot: I think if this is moved into a section specifically about style recommendations for breaking up long lines (as I intend to open a PR for shortly) it will be clear that this isn't a hard rule for all functions.

style/swift/sample.swift
```diff
@@ -95,6 +95,7 @@ struct DopeObject {
   }
 }
 
+<<<<<<< 591c3cde83e203ccffe8a69cc79f271da5497d92
```
dgalarza: Looks like an artifact from a merge conflict snuck in.

gfontenot: oops, will fix.

gfontenot:
@thoughtbot/ios we never came to a consensus here.

jakecraige:
@gfontenot You're correct about swift guidelines changing to not have the argument name in the function name, but have it as an argument label. That's described in the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/#argument-labels). 

I like this as is. All arguments on their own line and trailing paren on newline. This is also inline with out ruby/js guidelines so it's nice to have that consistency.

sharplet:
I've been trying to respond to this for probably 30 mins now, and I'm realising that it's super hard!

There are a bunch of different issues:

- Should the trailing parenthesis hug its argument or sit on the next line?
- Should each argument get its own line, or should the first parameter live on the same line as the function call?
- Should we use fixed indent widths or left-align things across multiple lines?
- If there are exceptions, what are they in practice?

However, putting the specifics aside for a moment, I personally feel pretty strongly that our guideline should include examples from, or at least inspired by, *real code we've written*.

Is this example inspired by a particular piece of code in a project you've worked on, @gordon? Or is it something you find yourself writing a lot? If so, I think it could be helpful to flesh this guideline out with some more concrete examples.

klaaspieter:
I've used this approach in [Ruby](https://github.com/thoughtbot/courier-web/blob/master/app/controllers/api/v1/broadcasts_controller.rb#L12-L18) and [Swift](https://github.com/thoughtbot/courier-ios/blob/master/Courier/Courier.swift#L31-L36).

tonyd256:
After watching Markable PRs come through I've found that I do not like breaking up the function parameters on the definition of the function, but do not mind it for the calling site.

jakecraige:
Responding to @sharplet 

> Should the trailing parenthesis hug its argument or sit on the next line?

New line. Part of the value in this formatting is less diff-noise, and having it on the same line would introduce that. This is also in line with our Ruby style guide.

> Should each argument get its own line, or should the first parameter live on the same line as the function call?

Get it's own line. The diff noise rule applies here too, but also as Gordon said before, "The first argument just kinda gets lost for me.", and I agree with that. And since Swift 3 everything has an argument label by default, I think this generally will look better than it did historically.

> Should we use fixed indent widths or left-align things across multiple lines?

I'm not sure I understand this question correctly, but I prefer the argument labels left aligned, with a single space between the `:` and the value. Any more than that introduce excess diff noise.

> If there are exceptions, what are they in practice?

As with all guidelines, I think this falls under Principle Zero, UYBJ.

For real code examples, I do this all the time, and we(Klass, Gordon and I), did on Markable, and I continue to on any Swift code I'm working in.

To me, this seems good enough to merge as is. If we end up having problems with it, we can definitely made adjustments or add more examples if it's confusing.

sharplet:
> > Should the trailing parenthesis hug its argument or sit on the next line?
>
> New line. Part of the value in this formatting is less diff-noise, and having it on the same line would introduce that.

When it comes to parameter lists, you end up with the same number of lines changed either way:

```diff
 let foo = Foo(
-  bar: Bar()
+  bar: Bar(),
+  baz: Baz()
 )
```

vs

```diff
 let foo = Foo(
-  bar: Bar())
+  bar: Bar(),
+  baz: Baz())
```

However having the trailing `)` on the next line does result in less punctuation change, so maybe that's what you were getting at. I don't mind either way, and I can see how there could be a small clarity improvement this way.

jakecraige:
Ah I forgot you can't have a trailing comma when calling a function in Swift. In Ruby you can so the diff is only a single line addition.

tysongach:
Is this still active? If there is no consensus, should we not have a guideline?

gfontenot:
I'm going to close this since it sounds like there's no real consensus here.

PR #407, opened 2016-02-12
Prefer protocol conformance to object inheritance

gfontenot:
Protocol conformance in Swift 2+ is a much more powerful form of
polymorphism than object inheritance and so should be the preferred
method. Among other things, using protocols means that you don't need to
care about the kind of object (struct, enum, or class), which puts way
fewer restrictions on your larger project architecture decisions.
Conversely, since object inheritance can only be achieved with classes,
they force you to make certain decisions about your application
architecture that might have adverse side effects.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -5,6 +5,7 @@ Swift
 
 * Prefer `struct`s over `class`es wherever possible
 * Default to marking classes as `final`
+* Prefer protocol conformance to class inheritance
 * Break long lines after 100 characters
 * Use 2 spaces for indentation
 * Use `let` whenever possible to make immutable variables


klaaspieter:
Prefer composition over inheritance?

👍

sidraval:
Definitely :+1: 

Maybe call out that this is a better way to share code, as well? Since it allows non-hierarchical types to share code.

gfontenot:
@klaaspieter Yeah, but I wanted to be more specific than that. I think preferring composition over inheritance is a larger guideline that isn't specific to swift.

@sidraval considered that, but didn't want to call it out specifically for that. I think it _can_ be used for code reuse, but we should probably view it as polymorphism first, code reuse second (just like with object inheritance).

mike-burns:
This in itself reads well, and further refinements can be made atop it. Any reason not to merge?

sharplet:
A minor nit: should it read "class inheritance" rather than "object inheritance"?

gfontenot:
@sharplet Sure, fixed.

Merging this now.

PR #409, opened 2016-02-13
Update rubocop.yaml to follow recent changes.

nickcharlton:
With changes introduced in [0.36.0](https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#0360-14012016):
- `Rails/DefaultScope` cop was removed.
- `Style/TrailingComma` has been split into `Style/TrailingCommaInArguments` and `Style/TrailingCommaInLiteral`.

`Style/TrailingCommaInLiteral` seemed to fit the spirit of our original rule best.

### DIFF
--- a/style/ruby/.rubocop.yml
+++ b/style/ruby/.rubocop.yml
@@ -334,7 +334,7 @@ Style/StringLiterals:
   EnforcedStyle: double_quotes
   Enabled: true
 
-Style/TrailingComma:
+Style/TrailingCommaInLiteral:
   Description: 'Checks for trailing comma in parameter lists and literals.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas'
   EnforcedStyleForMultiline: comma
--- a/style/ruby/.rubocop.yml
+++ b/style/ruby/.rubocop.yml
@@ -560,10 +560,6 @@ Rails/Date:
                   such as Date.today, Date.current etc.
   Enabled: false
 
-Rails/DefaultScope:
-  Description: 'Checks if the argument passed to default_scope is a block.'
-  Enabled: false
-
 Rails/FindBy:
   Description: 'Prefer find_by over where.first.'
   Enabled: false


teoljungberg:
This file is quite out of date, in general. And I'm unsure why it is even present.

All thoughtbot repos that get linted by Hound uses [this file](https://github.com/houndci/hound/blob/master/config/style_guides/thoughtbot/ruby.yml) which the Hound team keeps up to date when we upgrade Rubocop.

Should we just remove this file?

/cc @thoughtbot/hound

nickcharlton:
I'd be up for removing it instead. Hound people will see all of the issues way earlier than a stray file here.

On the flip side, we should have some sort of reference to there here.

teoljungberg:
I agree, linking to the `.rubocop.yml` file in Hound in the README for ruby makes alot of sense.

bernerdschaefer:
@teoljungberg I don't think this file should be removed, but should rather be kept up-to-date with what we're using for our projects. I like that we include the linter configuration files we use for our various languages:

* https://github.com/thoughtbot/guides/blob/master/style/ruby/.rubocop.yml (this file)
* https://github.com/thoughtbot/guides/blob/master/style/sass/.scss-lint.yml
* https://github.com/thoughtbot/guides/blob/master/style/javascript/.jscsrc

teoljungberg:
What is the benefit of keeping two copies of the same file? One in Hound, and one in our guides repo?

---
Teo Ljungberg

BlakeWilliams:
I agree with @bernerdschaefer. I think it would make sense for Hound to use `.rubocop.yml` from this repo if possible.

teoljungberg:
As mentioned before, we are working on adding org-wide-config for Hound. And I don't see the purpose of adding support for this specifically for thoughtbot, since it's so short lived.
We also don't want to maintain two files and repos to update whenever the configuration changes.

I agree that it makes sense for the configs to live in the guides repo once we have the org-wide-styleguide feature in place.
But until that happens, let's remove the config files from the guides repo to avoid any confusion which configuation is used.

---
Teo Ljungberg

BlakeWilliams:
I'm :-1: on removing the config files. I don't think Hound should affect where the config lives.

jferris:
I agree with @bernerdschaefer and @BlakeWilliams - it's nice to be able to submit a pull request that updates our guides and our tools in one go.

Maybe we should add a comment to `rubocop.yml` with instructions on how to merge it with Hound until the org-wide config is in place?

iainbeeston:
I don't have strong feelings about where the file should live but I do think the guides should mention our hound org-specific config (for those languages that it's configured for).

@teoljungberg Would it be possible for hound to pull org-specific config from an arbitrary URL? (for example, could we tell hound to take the org-specific config from https://raw.githubusercontent.com/thoughtbot/guides/master/style/ruby/.rubocop.yml ?

tysongach:
Any update here?

Hound now supports organization-wide configurations: https://robots.thoughtbot.com/hound-introduces-style-guides-for-organisations

nickcharlton:
I'm preeetttyyy sure that we're consistently using the organisation `rubocop.yml` now, so we should close this and work with the one we're using for real (as that's newer).

I'm going to close this as it'll probably be easier to open a fresh PR given divergence. I'd happily review any replacements.

jakecraige:
@nickcharlton I'm pretty sure our org config is set to use this repo's `.hound.yml` and thus this `.rubocop.yml` for the rules?

PR #412, opened 2016-03-02
Add Swift + Obj-C guideline about weak outlets

kaishin:


kaishin:
Ping @thoughtbot/ios

style/objective_c/README.md
```diff
@@ -11,8 +11,8 @@ Objective-C
 * Order `@class` directives alphabetically.
 * Order `@property` modifiers: memory management, atomicity, writability.
 * Leave out `@property` modifiers unless needed, `nonatomic` is the only one
-  needed in most cases except connecting views with IB in which case `weak` may
-  also be needed.
+  needed in most cases.
+* Avoid `weak` references to IBOutlets.
```
gfontenot: How about `Prefer` for wording? We're not really wanting to avoid `weak`, we'd just rather not worry about it.

kaishin: Yes. The only reason I went for that is that we get `weak` by default. Fine with either.

calleerlandsson:
What are the pros of avoiding `weak` for `IBOutlets`?

gfontenot:
@calleerlandsson No pros to avoiding `weak` but there are no `pros` for keeping it around either. The only times we use `weak` is to avoid a retain cycle, which would be the same case here. There's only one real chance of creating a retain cycle with views and we wouldn't ever do it, so there's no benefit to using `weak` and it just complicates the memory model.

Apple actually suggests using `strong` (which is the default) for `IBOutlet`s as well, they just haven't updated their own documentation/sample code. This has been true since iOS 6.

kaishin:
@calleerlandsson As an addendum to Gordon's answer, it's a lot more common to manipulates views and constraints in and out of the hierarchy than for the view to hold a reference to the VC. You want a strong reference in the former and weak one in the latter.

tonyd256:
:+1:

calleerlandsson:
I see. Thanks for explaining!

> There's only one real chance of creating a retain cycle with views and we wouldn't ever do it [...]

When does this happen so I know to avoid it?

kaishin:
> When does this happen so I know to avoid it?

@calleerlandsson  That happens if the view or outlet holds a reference to the outlet owner such as the VC. As @klaaspieter put it, if you do this you deserve a retain cycle!

kaishin:
`Prefer strong IBOutlet references` is better than `Prefer strong references to IBOutlets`. Thoughts?

calleerlandsson:
Cool. Sounds good!

PR #413, opened 2016-03-04
Add Swift guideline recommending weak references are evaluated once

sharplet:
If a weak reference is evaluated multiple times within the same scope, it's technically possible for the reference to change between `nil` and non-`nil` between references. In order to ensure that valid references remain valid for the entire scope, recommend that weak references are evaluated only once.

The style in the example (using a guard to optionally bind `self`) is just the way I currently do this. It's kind of like using a `@strongify(self)` macro in Objective-C, combined with an early return. Open to suggestions for alternative styles!

### DIFF
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -38,6 +38,13 @@ let cool = yTown(5) { foo in
   }
 }
 
+// Strongify weak references in async closures
+APIClient.getAwesomeness { [weak self] result in
+  guard let `self` = self else { return }
+  self.stopLoadingSpinner()
+  self.show(result)
+}
+
 // MARK: Optionals
 
 var maybe: Bool?
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -20,3 +20,5 @@ Swift
 * When using `Void` in function signatures, prefer `()` for arguments and
   `Void` for return types.
 * Prefer strong IBOutlet references.
+* Avoid evaluating a weak reference multiple times in the same scope.
+  Strongify first, then use the strong reference.


mike-burns:
Ping @thoughtbot/ios .

jakecraige:
I like it

gfontenot:
👍 I like the backtick trick for re-binding to `self`.

style/swift/sample.swift
```diff
@@ -36,6 +36,15 @@ let cool = yTown(5) { foo in
   }
 }
 
+// Strongify weak references in async closures
+APIClient.getAwesomeness { [weak self] result in
+  guard let `self` = self else { return }
+  if let awesome = result.value {
```
tonyd256: for this case I would probably include this in the guard

dazmuda: And I'm thinking "You know what would be so yummy here? A switch statement. Sprinkle some error handling up ins."

So maybe just some really bland uses of self like:
self.stopLoadingSpinner()
self.presentAlert()
self.updateDisplay()
self.boopBoop()

dazmuda:
Good stuff. This should def go in. 🌈

style/swift/sample.swift
sharplet: Something like this?
```swift
  switch result {
  case let .Success(awesome):
    self.awesomeStuff.append(awesome)
    self.updateDisplay()
  case let .Failure(error):
    self.show(error)
  }
```

dazmuda: Ah sorry I was saying that if we're using "result" in a snippet, we should handle errors. But since error-handling style and eval-weak style in the same snippet is probably a bit too much, maybe just don't use the result, don't _ it even though we should, and then call some really barebones methods that use self.

guard let `self` = self else { return }
self.stopLoadingSpinner()
self.presentAlert()
self.updateDisplay()

sharplet: I get it now, good thinking!

sharplet:
Thanks everyone!

PR #416, opened 2016-03-22
Use camel case over snake case for readme example

ezekg:
Since using (lower) camel case is mentioned within the readme, I figured it was a mistake to use snake case over camel case for the private method and variables example. Feel free to close this if I'm wrong, but thought I'd fix it to avoid any possible confusion. :+1:


mike-burns:
Is this correct -- is this actually the CS standard?

ezekg:
That's how it is in [other](https://github.com/polarmobile/coffeescript-style-guide#naming-conventions) [guides](https://github.com/orgsync/coffeescript-style-guide#classes). Snake case in JS/CS isn't usually allowed.

BlakeWilliams:
This sounds correct for JavaScript and according to the other guides for CoffeeScript, correct there too.

PR #419, opened 2016-04-07
Start a quick "how to ..." section

mike-burns:
This section is to address the growing amount of internal tools we have
built over time; it is an attempt at making them more discoverable in a
problem/solution format instead of a long list, plus a quick-start
guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ Guides for getting things done, programming well, and programming in style.
 
 * [Best Practices](/best-practices)
 * [Code Review](/code-review)
+* [How to](/how-to)
 * [Protocol](/protocol)
   * [Communication](/protocol/communication)
   * [Git](/protocol/git)
--- a/how-to/README.rst
+++ b/how-to/README.rst
@@ -1,0 +1,55 @@
+How do I ...
+============
+
+.. contents::
+
+... start a new Rails app?
+--------------------------
+
+Use suspenders_:
+
+.. code:: sh
+
+  $ gem install suspenders
+  $ suspenders the-name-of-your-project-here
+  $ cd the-name-of-your-project-here/
+  $ bin/setup
+  $ rake
+
+.. _suspenders: https://github.com/thoughtbot/suspenders
+
+... feature-test a Rails app's Javascript?
+------------------------------------------
+
+Use capybara-webkit_. In your ``Gemfile``:
+
+.. code:: ruby
+
+  gem "capybara-webkit"
+
+In ``spec/support/capybara_webkit.rb`` (for Rspec):
+
+.. code:: ruby
+
+  Capybara.javascript_driver = :webkit
+  
+  Capybara::Webkit.configure do |config|
+    config.block_unknown_urls
+  end
+
+When writing a spec, you must set the ``:js`` flag for that test to make use of
+capybara-webkit. For example, in ``spec/features/user_signs_in_spec.rb``:
+
+.. code:: ruby
+
+  feature "Authentication", :js do
+    scenario "A user signing in" do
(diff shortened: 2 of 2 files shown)


howto/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```ruby
+Capybara.javascript_driver = :webkit
+
+Capybara::Webkit.configure do |config|
+  config.block_unknown_urls
+end
+```
+
+When writing a spec, to make use of this you must set the `:js` flag for that
```
mxie: How about: `When writing a spec, you must set the :js flag to make use of capybara-webkit.`?

howto/README.md
```diff
@@ -0,0 +1,49 @@
+# How do I ...
+
+## ... start a Rails app?
```
mxie: How about `start a new...`?

mike-burns: That does improve clarity; done.

README.md
```diff
@@ -5,6 +5,7 @@ Guides for getting things done, programming well, and programming in style.
 
 * [Best Practices](/best-practices)
 * [Code Review](/code-review)
+* [How to](/howto)
```
mxie: `howto` -> `how-to`

Also, I think it might be useful to briefly describe this page here. It's unclear as to what we can find by following this link.

howto/README.md
```diff
@@ -0,0 +1,49 @@
+# How do I ...
+
```
jakecraige: This may be premature optimization, but what about having a table of contents at the top? 

I imagine this list might grow quickly and be hard to see what's included in it

iwz: I learned recently that `RST` files get an automatically generated table of contents. It's pretty cool!

seanpdoyle:
@mike-burns I am a big fan of this idea, and I think this is a great start.

Could future revisions to this list include non-technical processes like making a company wide announcement with Constable or sharing a link with Botsters?

jakecraige:
@seanpdoyle I like that idea, but I think that stuff might have to go in the handbook since those aren't public and this is

seanpdoyle:
@jakecraige wow, I've been reading so many Handbook PRs lately that I didn't notice where I was.

Carry on!

howto/README.md
```diff
@@ -0,0 +1,49 @@
...
+Capybara::Webkit.configure do |config|
+  config.block_unknown_urls
+end
+```
+
+When writing a spec, you must set the `:js` flag for that test to make use of
+capybara-webkit. For example, in `spec/features/user_signs_in_spec.rb`:
+
+```ruby
+feature "Authentication", js: true do
```
calleerlandsson: Since symbols are now treated like metadata keys with true values by default (https://github.com/rspec/rspec-core/pull/1006), should this be `feature "Authentication", :js do`?

calleerlandsson:
I like this idea.

howto/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## ... start a new Rails app?
+
+Use [suspenders]:
+
+```ruby
+$ gem install suspenders
+$ suspenders the-name-of-your-project-here
+$ cd the-name-of-your-project-here/
+$ rake
```
teoljungberg: Should this be replaced with `bin/setup`?

README.md
mike-burns: This would be the only page with an explanation. Is there a better name for this?

howto/README.md
mike-burns: I converted it to RST and added a TOC. It's fancy now.

mike-burns:
So ... good to merge?

seanpdoyle:
:ship:

PR #420, opened 2016-04-09
Be consistent about class naming and structure in a project

kylefiedler:
My new pet peeve is finding inconsistent naming in a project. This happens a lot in thoughtbot owned projects where many designers touch the CSS.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -40,6 +40,7 @@
 
 * Don't use ID's for style.
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
+* Be consistent about naming conventions for classes. For instance, if a project is using BEM, continue using it, and if it's not, do not introduce it.
 * Use ID and class names that are as short as possible but as long as necessary.
 * Avoid using the direct descendant selector `>`.
 * Avoid nesting more than 3 selectors deep.


style/sass/README.md
```diff
@@ -40,6 +40,7 @@
 
 * Don't use ID's for style.
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
+* Be consistent about naming conventions for classes. If the project is using BEM continue to use BEM, if the project is not using BEM don't introduce a few BEM classes.
```
kaishin: Not sure the second sentence is needed but I think that's because this the problem we're most likely to deal with? It feels a bit passive-agressive especially the fact that BEM is repeated 4 times.

kylefiedler: Do you have a better way to word it? Is the first sentence enough?

kaishin: > Be consistent about naming conventions for classes. For instance, if a project is using BEM,  continue using it, and if it's not, do not introduce it.

tysongach:
In general, I’m a big 👍  to this addition…

- I do agree with @kaishin in making that sense a bit more succinct.
- One thing I love about BEM and why I use it is that there is a rough “spec” out there; people can easily pick up where someone else left off. If someone is _not_ using BEM, it’s hard to pick up when I don’t know what cobbled together methodology they used: this is my pain point with a lot of project. The class architecture is usually described as a “sort of like BEM but different in X, Y and Z ways…
- I’ve been wondering if adding a small section to a project’s README that explains things like this would be beneficial…for example, “This project uses BEM”, “It also uses Autoprefixer, here’s the config and browsers we support”, “the stylesheets should be broken down into these directories and the purpose of each direct is…”, etc.

At the end of the day, when constructing CSS classes we aim for consistency per-project, not uniformity amongst all thoughtbot designers.

kylefiedler:
@tysongach @kaishin Yup, I like that a lot more, thanks. UPDATED!

> I’ve been wondering if adding a small section to a project’s README that explains things like this would be beneficial…for example, “This project uses BEM”, “It also uses Autoprefixer, here’s the config and browsers we support”, “the stylesheets should be broken down into these directories and the purpose of each direct is…”, etc.

I've been thinking about that as well. It would certainly make this easier. Is the README the best place for it or should it go in the manifest file in a comment or somewhere else?

> At the end of the day, when constructing CSS classes we aim for consistency per-project, not uniformity amongst all thoughtbot designers.

Right, I don't think as a group we'll be able to agree on one and even if we did I think it'd be the wrong approach. I think it should be a per-project choice.

tysongach:
I recently added this to FormKeep: https://github.com/thoughtbot/formkeep/tree/tg-the-future-is-bright#front-end-architecture

Helpful? Thoughts?

kylefiedler:
@tysongach I like that. Want to start another PR with instructions for what to do?

tysongach:
@kylefiedler Yes, and I’ll also add a PR to Suspenders to have this section added to the project readme.

tysongach:
@kylefiedler Can we get this merged in?

PR #425, opened 2016-05-27
Ensure local master is up to date before merging

tjmw:
We fetched so `origin/master` is up to date, but I think we also need to make sure our local `master` is up to date before merging, so be explicit and add this command.

An alternative could be to `git merge` instead, since we already did the `fetch` part of `git pull`.

### DIFF
--- a/protocol/git/README.md
+++ b/protocol/git/README.md
@@ -106,6 +106,7 @@ View a list of new commits. View changed files. Merge branch into master.
     git log origin/master..<branch-name>
     git diff --stat origin/master
     git checkout master
+    git pull
     git merge <branch-name> --ff-only
     git push
 


protocol/git/README.md
```diff
@@ -106,6 +106,7 @@ View a list of new commits. View changed files. Merge branch into master.
     git log origin/master..<branch-name>
     git diff --stat origin/master
     git checkout master
+    git pull
```
teoljungberg: The default `pull` action is to do a merge, whilst we use a rebase workflow.

Also, [above] we already prompt the user to fetch and update with `origin/master`. Which makes me wonder, is this step required?

[above]: https://github.com/thoughtbot/guides/pull/425/files#diff-1cc33e156f1a96c475b181ed56b14931R95

tjmw: Good questions @teoljungberg! Thanks for the feedback. I think even though `origin/master` is up to date, that it's still possible that your local `master` (which is what we're merging into) could be behind.

Good point about the rebase workflow. I think because we're merging `origin/master` into `master` it will always be a fast forward merge, unless something has gone wrong (i.e. we've got changes on our local master which haven't been pushed).

Also happy to close this PR though if we don't think it adds anything.

tjmw: Actually, thinking a bit more about it, you're right. It's not necessary. Since we rebased on `origin/master` above, we know we've got changes from `origin`, so we'll get those as part of the merge. It's funny to me though that that piece is *slightly* hidden in this workflow.

teoljungberg: If your local `master` is not up to date, you won't be allowed to push to master. So I think this is implicitly solved.

I'd be fine merging this PR in with a change to `pull --rebase` instead of `pull`

tjmw:
Yeah absolutely, we'd not be able to push if we weren't up to date. I was just thinking about making it less likely for that to happen. `pull --rebase` would work, I tend to use that by default anyway. But I guess if pulling isn't getting us anything either way maybe it's better to leave it out.

protocol/git/README.md
tjmw: Yeah absolutely, we'd not be able to push if we weren't up to date. I was just thinking about making it less likely for that to happen. `pull --rebase` would work, I tend to use that by default anyway. But I guess if pulling isn't getting us anything either way maybe it's better to leave it out.

PR #426, opened 2016-06-07
Place methods and tests for methods in same order

mcmire:
When writing tests for a class, it's nice to have the methods and the
tests for those methods listed in the same order. Why? Not because it
makes it easier to find tests (because you can simply use Ctrl-F) or
easier to know where to put new tests (because you could always put them
at the bottom). Really, it's the same argument behind alphabetizing
associations -- it just looks cleaner, and shows attention to detail.

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -44,6 +44,8 @@ Ruby
   `attr_writer`s, and `attr_accessor`s.
 * Order class methods above instance methods.
 * Prefer method invocation over instance variables.
+* When writing tests for a class, place tests for methods in the same order as
+  the methods themselves are listed in the class.
 
 [trailing comma example]: /style/ruby/sample.rb#L49
 [required kwargs]: /style/ruby/sample.rb#L16


gylaz:
I get a feeling this could break down quick. Developers move methods around in the same class for no particular reasons, and likely forget to move the associated test.

tommydangerous:
I do this. +1

joshuaclayton:
I agree with @gylaz - this seems easy to forget, and doesn't seem like any/many upsides.

For associations, it's easy to both spot issues in order, and then use a single command to sort them since they're commonly one-liners. With code spanning multiple lines, running sort in your editor of choice won't work (unless there's a new tool that would handle that).

I may actually do this - but unintentionally, and not something I'd want to have to remember.

tysongach:
Any update here?

derekprior:
I generally like when this is done, but it's difficult to enforce and has minimal benefit. It doesn't seem worth the headspace to remember this rule.

PR #427, opened 2016-06-22
First pass at an Elixir style guide

BlakeWilliams:
The sample is a little lacking since I didn't have much time to flesh it
out. Would love ideas for good example functions.

### DIFF
--- a/style/elixir/README.md
+++ b/style/elixir/README.md
@@ -1,0 +1,11 @@
+Elixir
+======
+
+[Sample](sample.ex)
+
+* Don't separate multiple declarations of functions of the same name with
+  newlines.
+* Indent pipes by two spaces when using the match operator (`=`).
+* Don't indent pipes when not using the match operator (`=`).
+* Prefer pattern matching and guards in function definitions over conditionals
+  in the function body.
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -320,6 +320,11 @@ Haskell
 * Avoid partial functions (`head`, `read`, etc).
 * Compile code with `-Wall -Werror`.
 
+Elixir
+------
+
+* Avoid macros.
+
 Ember
 -----
 
--- a/style/elixir/sample.ex
+++ b/style/elixir/sample.ex
@@ -1,0 +1,10 @@
+defmodule Sample do
+  def double(a) when is_binary(a) do
+    a
+    |> String.downcase
+    |> String.duplicate(2)
+  end
+  def double(a) when is_integer(a)  do
+    a * a
+  end
+end


style/elixir/README.md
```diff
@@ -0,0 +1,11 @@
+Elixir
+======
+
+[Sample](sample.ex)
+
+* Avoid macros.
```
drapergeek: Is this a style choice?

style/elixir/README.md
```diff
@@ -0,0 +1,11 @@
...
+* Don't separate multiple declarations of the same function with a newline.
```
drapergeek: Is this a common thing in the community? I haven't followed this pattern so I'm hesitant on giving it a thumb without seeing it in action.

style/elixir/README.md
BlakeWilliams: It is not, this probably belongs in best practices. I'll move it there.

style/elixir/README.md
BlakeWilliams: I'm not sure to be honest. It's something I've done and I threw it in here to get some feedback on it.

derekprior:
Is there a community style guide with any momentum we can adopt instead?

BlakeWilliams:
Looked it up and there's https://github.com/niftyn8/elixir_style_guide#source-code-layout which seems to be relatively consistent with what we do.

derekprior:
Reviewing that, here's what I'd take issue with:

> Use `do:` for single line if/unless statements.

I'd still rather see indention as it signals complexity that was introduced by the if to me. The one-liner hides that more than I'd like.

Everything else seemed fine to me. Not always what I'd do, but I'd be happy to abide by it.

gylaz:
Idea: Could we do something like link to the community guide, then have a section for our *Deviations*?

derekprior:
Yes. But I'm also willing to say "No deviations" and then debate each of them as they come up.

style/elixir/README.md
```diff
@@ -0,0 +1,11 @@
...
+* Indent pipes by two spaces when using the match operator (`=`).
+* Don't indent pipes when not using the match operator (`=`).
```
gfontenot: I can't stand this practice. I don't understand why it's a thing. IMO, we should prefer consistency over (uninforced) semantic whitespace and just indent pipes 2 spaces all the time.

drapergeek: I agree with the consistency part but I'd go the other way and say don't indent ever.

BlakeWilliams:
Closing because this is stale. Feel free to re-open if there's any interest.

BlakeWilliams:
Reopening because of #460. I'd love feedback or additions here.

genericlady:
@BlakeWilliams I went ahead and added to your branch. Should I push up changes or add recommendations to this thread for review?

BlakeWilliams:
I think making a PR against this PR would work well, or adding the suggestions in text here. A PR might be easier for discussion though.

drapergeek:
I think this looks great! How about we start here and we can all PR against your base file here? I have a few more I'd like to add.

genericlady:
Started a modifications branch
[PR 482 modifications to bmw-elixir](https://github.com/thoughtbot/guides/pull/482)

BlakeWilliams:
@drapergeek Sounds good, I'll go ahead and merge this one then.

PR #430, opened 2016-06-30
Swift: Force unwrap `segue.destinationViewController`

jakecraige:
We should force unwrap the `destinationViewController` on a segue when it
matches one of our identifier.

This causes us to fail as early and loud as possible when
something unexpected happens.

Alternatives considered
-----------------------

**Optional unwrap or guard let/return**:

While this is safer, it will silently not do anything when this isn't a case we
expected. If we're expecting a specific type of vc and it's not that type, we
should be alerted somehow so we can fix it.

**Guard let/fatalError**:

I don't have much of an argument against this one other than it's more verbose.
I don't think the verbosity would improve debugging much so I like the less
verbose version.

### DIFF
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -126,3 +126,15 @@ guard let oneItem = somethingFailable(),
 guard let something = somethingFailable() else {
   return someFunctionThatDoesSomethingInManyWordsOrLines()
 }
+
+// MARK: UIKit
+
+final class MyViewController {
+  override func prepareForSegue(segue: UIStoryboardSegue, sender: AnyObject?) {
+    switch segue.identifier {
+    case "mySegue"?:
+      let vc = segue.destinationViewController as! MyOtherViewController
+    default: break
+    }
+  }
+}
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -22,3 +22,4 @@ Swift
 * Prefer strong IBOutlet references.
 * Avoid evaluating a weak reference multiple times in the same scope.
   Strongify first, then use the strong reference.
+* Force unwrap `segue.destinationViewController` to the type you expect


style/swift/sample.swift
```diff
@@ -126,3 +126,15 @@ guard let oneItem = somethingFailable(),
...
   return someFunctionThatDoesSomethingInManyWordsOrLines()
 }
+
+// MARK: UIKit
+
+final class MyViewController {
+  override func prepareForSegue(segue: UIStoryboardSegue, sender: AnyObject?) {
+    switch segue.identifier {
+    case "mySegue"?:
+      let vc = segue.destinationViewController as! MyOtherViewController
```
jakecraige: HAHAHAHAHA

tonyd256: I mean, the dog says it all

style/swift/sample.swift
```diff
@@ -126,3 +126,15 @@ guard let oneItem = somethingFailable(),
...
+    default: break
```
sharplet: What do you think about moving the `break` to its own line? For consistency with the other `case`.

sharplet:
One little comment about formatting, but this looks good to me 👍

tonyd256:
I've also been using a `guard` only.

```
override func prepareForSegue(segue: UIStoryboardSegue, sender: AnyObject?) {
  guard let vc = segue.destinationViewController as? MyOtherViewController else { return }
  //...
}
```
or if I have more than one I'll use if-let
```
override func prepareForSegue(segue: UIStoryboardSegue, sender: AnyObject?) {
  if let vc = segue.destinationViewController as? MyOtherViewController {
    //...
  } else if let ... {
    // ...
  }
}
```

I kinda like these because then I don't have to check 2 things to segue (the identifier and the vc). The potential bad thing is if you have 2 different segues to the same VC that need to do different things. I haven't had this happen to me yet but if I did I would switch to checking the identifier.

Both of these patterns above will silently fail if the vc is not correct where as the on in this PR will crash. I'm on the fence on which is better and mostly upset that we have to make this choice ;)

Thoughts?

gfontenot:
I think this is probably fine for the existing tools. I think we should be looking to fix the need for this force unwrap entirely for the future though.

klaaspieter:
@tonyd256 Another problem with the `destinationViewController` approach is if your destination is actually a container controller like `UINavigationController`. Your approach will silently fail there too. It's fairly easy to forget to change this when you've embedded an existing view controller inside a container controller or if you're not aware how segues work with container controllers.

I like the force unwrapping here because it'll warn as early as possible, instead of silently ignoring a failure.

tonyd256:
@klaaspieter Yes my way has some potential avenues of silently failing and I think that there is actually an interesting discussion to be had about silent failing vs hard crash in many places beyond segues. However, using the segue identifier can also lead to silent failure if there is a typo in the storyboard or `prepareForSegue` code or if the identifier is changed in one spot but not the other on accident. At least with the casting way you have the compiler to let you know if something has changed.

Perhaps we can label every segue and force unwrap the identifier too.

Also, is there something I'm missing about:
>if you're not aware how segues work with container controllers.
Is that simply that container controllers invoke a segue with their embedded VCs?

jakecraige:
To sort of take this down a different route, I was talking with @Ben-G over the weekend about this and he shared that he doesn't use storyboard segues at all. 

He cited that segue's aren't really what make storyboards valuable, it's the layout part. He even goes further and does plain views in XIBs, rather than view controllers.

If we decided not to do something like that, this whole thing goes away, and has many benefits like full control over initialization so we can remove explicitly unwrapped optionals on the destination VC and great reuse since each view is independent of where it lies in the navigation stack.

Have we considered something like that?

sharplet:
> ... he doesn't use storyboard segues at all.
>
> ...
>
> Have we considered something like that?

✋ I have! In the back of my mind I've been considering the idea as an experiment for my next project. I would _love_ to have the opportunity to use real initialisers and the certainty they provide.

Ben-G:
@tonyd256 @jakecraige Here's a writeup I just found that describes the Xib approach: https://medium.com/ios-os-x-development/dependency-injection-in-view-controllers-9fd7d2c77e55#.314i4c41r

I like this approach since it provides benefits from visual design in interface builder while allowing initializer based dependency injection.

gfontenot:
We shouldn't get sidetracked here. If we want to talk about dropping storyboard segues that should happen in a separate PR that properly tracks that discussion.

klaaspieter:
> Also, is there something I'm missing about:

>> if you're not aware how segues work with container controllers.

> Is that simply that container controllers invoke a segue with their embedded VCs?

@tonyd256 Yes. A segue to a navigation controller has a `UINavigationController` as it's `destinationViewController`, which would silently fail with an optional cast. The force unwrap ensures the error is surfaced as soon as the code path is hit. Definitely not ideal, but since the compiler can't surface it, we should surface it at the earliest possible opportunity after that.

There is an argument that this leads to a potential crash in the user hands. I think that's preferable to presenting a screen with inconsistent (or no) state.

I see this discussion as being similar to the discussion about removing asserts from release builds. [Mike Ash](https://www.mikeash.com/pyblog/friday-qa-2013-05-03-proper-use-of-asserts.html) makes some great points against that: 

> However, I am firmly of the opinion that disabling asserts in release builds is a terrible idea. The runtime cost should be negligible, and if it's not, then you should redo your asserts to fix that. As for avoiding app termination, asserts should be written such that a failure always means that something has gone terribly wrong. It is possible that the app will continue functioning in the face of that. It's more likely that it'll crash. It's also possible that it'll keep running, but corrupt your user's data. A clean crash is vastly preferable. No code is free of bugs, and crashing early and obviously when a bug is encountered is much better, even in a release build running on a user's machine. Generating a cleaner crash log will help you debug the failures more quickly.

klaaspieter:
As per @gfontenot suggestion: #431 tracks the discussion about dropping segues all together.

jakecraige:
I'm going to close this now that we have [Perform](https://github.com/thoughtbot/Perform) as that would be my preference when dealing with segues, and it also happens to force unwrap them 👍

PR #431, opened 2016-07-06
Swift: Prefer presenting a view controller manually

klaaspieter:
In the discussion about force unwrapping (#430)
`UISegue#destinationViewController` we got sidetracked in talking about
dropping segues all together. This is to continue that discussion
without it intervening with the discussion about force unwrapping.

I feel segues can cause more issues than they're worth. The value in
storyboards is not in segues, but in the visual layout. Segues are
annoying because they pretty much guarantee having to keep state around.
If a value needs to be communicated to the destination of a segue, that
needs to happen in `prepareForSegue`. This decouples presenting the view
controller from configuring it.

By dropping segues we can manually instantiate and configure view
controllers. Initialization can be done using
`UIStoryboard#instantiateViewControllerWithIdentifier`.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -22,3 +22,4 @@ Swift
 * Prefer strong IBOutlet references.
 * Avoid evaluating a weak reference multiple times in the same scope.
   Strongify first, then use the strong reference.
+* Prefer presenting a view controller manually over segues
--- a/style/swift/sample.swift
+++ b/style/swift/sample.swift
@@ -126,3 +126,17 @@ guard let oneItem = somethingFailable(),
 guard let something = somethingFailable() else {
   return someFunctionThatDoesSomethingInManyWordsOrLines()
 }
+
+// MARK: UIKit
+
+final class MyViewController {
+  func f() {
+    guard let viewController = self.storyboard?.instantiateViewControllerWithIdentifier(
+      StoryboardIdentifiers.presentedViewController
+    ) else { return }
+    let valueToBeAwareOf = "The presented VC needs to be aware of this value"
+    viewController.property = value
+    presentViewController(viewController, animated: true, completion: .None)
+  }
+}
+


klaaspieter:
I'm running into another problem with segues right now. I have a view controller that loads some information, that should also be made available to a presented controller when it becomes available. With segue it's weird to get a reference to the presented controller outside of `prepareForSegue`. The following example shows what I mean:

```
final class InformationViewController: UIViewController {
  func viewWillAppear(animated: Bool) {
    super.viewWillAppear(animated)
    loadInformation {
      self.informationLabel.text = $0.description

      // Information is loaded here but we don't have a reference to the DetailedInformationViewController, this is a potential solution.
      (self.presentedViewController as? DetailedInformationViewController)?.information = $0
    }
  }

  func prepareForSegue(segue: UIStoryboardSegue) {
    (segue.destinationViewController as! DetailedInformationViewController).information = viewModel.information
  }
}
```

With manual initialization a reference to a view controller is available whenever needed.

klaaspieter:
ping @thoughtbot/ios

sharplet:
For the case you mention @klaaspieter it might make sense to store a weak reference to the detail controller in `prepareForSegue`. But if `loadInformation` also has the side effect of presenting the controller, perhaps the controller itself can be passed as a parameter to the completion block?

In any case, having control over view controller initialisation would definitely afford more options for solving this kind of problem, especially ones that minimise state.

tonyd256:
I don't think this is actually gaining us "control" over VC initialization. We still have to call `instantiateViewControllerWithIdentifier` which is calling the init function for us. Then we set the values we want to set after this call. I'm pretty sure this is what segues are doing internally anyways then they pass the result to the `prepareForSegue` function for us to do whatever else we need.

Also, I don't like this statement so much:
>With manual initialization a reference to a view controller is available whenever needed.

I think a clean architecture should avoid this as much as possible and try to use delegates and callbacks when needed. Also, why can't we use the `prepareForSegue` function to assign a local reference then?

sharplet:
To clarify, when I refer to regaining control over initialisation, what I have in mind is actually something like this: https://gist.github.com/JaviSoto/3bfb2c9e272a1363431fdad4d9b1d7e8

That gist demonstrates a convenience initialiser that internally delegates to the storyboard. Another way might be to use factory methods on `UIStoryboard` to instantiate and configure storyboards.

These approaches aren't perfect, and feel a little complex, but it's a tradeoff for a different kind of complexity when using `prepareForSegue`. I'm not convinced it will solve all the problems, but I'm interested in experimenting.

tonyd256:
Ah, that gist is a good example. I think doing things that will does give use more control and the ability to use dependency injection.

tonyd256:
hmm, I decided to try that out ... and well he uses ObjC for a reason. I don't think what he is doing is possible in purely Swift.

sharplet:
Yeah the main difference is that in Objective-C it's legal to change the value of `self` in an initialiser.

sharplet:
Topical: the [latest episode of Swift Talk](https://talk.objc.io/episodes/S01E05-connecting-view-controllers) is about connecting view controllers. I'll report back once I've seen more of it.

sharplet:
So in that Swift Talk episode, they work through a refactoring where they replace segues with event callbacks and code to present/dismiss/push view controllers.

They then refactor this connective code with a method object they call `App`, that does two things:

- Defines connections between view controllers in its initialiser
- Defines method that implement those connections, such as `showProfile()` and `showEpisode()`

tonyd256:
@sharplet the talk sounds interesting. How did they instantiate their VCs?

sharplet:
@tonyd256 They use `instantiateViewControllerWithIdentifier` and force-cast the result. They discussed moving that to an extension / factory method on `UIStoryboard`, but didn't cover that in the episode.

tonyd256:
Ah ok so they still can't have non-optionals for dependencies?

sharplet:
That's correct. So really it seems the only way to have non-optional dependencies would actually be to ditch storyboards and use nibs.

However, while Javi's approach still requires implicitily-unwrapped optionals, it does allow for those dependencies to be injected in the initialiser and thus for properties to be (externally, at least) readonly.

jakecraige:
This lib exists now which helps a lot with this https://github.com/thoughtbot/Perform

style/swift/README.md
```diff
@@ -22,3 +22,4 @@ Swift
 * Prefer strong IBOutlet references.
 * Avoid evaluating a weak reference multiple times in the same scope.
   Strongify first, then use the strong reference.
+* Prefer presenting a view controller manually over segues
```
sharplet: These days I'd personally be tempted to say "Prefer segues over presenting a view controller manually".

klaaspieter: I haven't used `Perform` at all so I'll defer to everyone's judgement on that 👍🏼

gfontenot:
I don't think this is something we should have a guideline over (feels much more like a case-by-case basis thing), and there doesn't seem to be a consensus anyway, so I'm going to go ahead and close.

PR #432, opened 2016-07-06
Swift: Prefer imperitive verbs for IBAction/target-action methods

jakecraige:
This was brought up by @sharplet in a PR review so I figured I'd open a guideline to discuss.

### DIFF
--- a/style/swift/README.md
+++ b/style/swift/README.md
@@ -22,3 +22,6 @@ Swift
 * Prefer strong IBOutlet references.
 * Avoid evaluating a weak reference multiple times in the same scope.
   Strongify first, then use the strong reference.
+* Prefer to name `IBAction` and target/action methods using a verb describing
+  the action it will trigger, instead of the user action
+  (e.g., `edit:` instead of `editTapped:`)


sharplet:
👍

jakecraige:
cc @thoughtbot/ios

klaaspieter:
💯 ％ 👍

mike-burns:
I feel a little confused by the description here, since "edit" and "edit tapped" are both cases where the verb ("to edit") is in the imperative mood. But everyone else (who is affected by this) seems to understand what this means.

gilesvangruisen:
@mike-burns I think in the latter case ("edit tapped"), "edit" more likely refers to "edit button", as in "the edit button was tapped".

sharplet:
@mike-burns "imperative" is a word I reached for during the original PR review to try and express what I meant. I wasn't certain it was the correct one!

Perhaps the point is more that the name should describe a _user action_ rather than a UI event.

jakecraige:
Pushed an edit attempting to clarify, let me know what y'all think.

style/swift/README.md
```diff
@@ -22,3 +22,6 @@ Swift
 * Prefer strong IBOutlet references.
 * Avoid evaluating a weak reference multiple times in the same scope.
   Strongify first, then use the strong reference.
+* Prefer to name `IBAction` and target/action methods using a verb describing
+  the action it will trigger, instead of the user action.
+  Ex: `edit:` instead of `editTapped:`.
```
sharplet: Grammar suggestion:

> ... user action (e.g., `edit:` instead of `editTapped:`).

sharplet:
I like it 👍

PR #433, opened 2016-07-07
Clarify where modules live in classes

gylaz:
I believe modules (i.e. `include/extend`) should come after constants.

Example:

```
class User < ActiveRecord::Base
  PRIMARY_ATTRIBUTE = "name"

  include Personable

  validates :name, presence: true

  ...

end
```

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -14,7 +14,8 @@ Rails
 * Order ActiveRecord associations above ActiveRecord validations.
 * Order controller contents: filters, public methods, private methods.
 * Order i18n translations alphabetically by key name.
-* Order model contents: constants, macros, public methods, private methods.
+* Order model contents: constants, modules, macros, public methods,
+  private methods.
 * Put application-wide partials in the [`app/views/application`] directory.
 * Use `def self.method`, not the `scope :method` DSL.
 * Use the default `render 'partial'` syntax over `render partial: 'partial'`.


mcmire:
That seems reasonable. 👍

style/rails/README.md
```diff
@@ -14,7 +14,8 @@ Rails
 * Order ActiveRecord associations above ActiveRecord validations.
 * Order controller contents: filters, public methods, private methods.
 * Order i18n translations alphabetically by key name.
-* Order model contents: constants, macros, public methods, private methods.
+* Order model contents: constants, modules, macros, public methods,
```
bernerdschaefer: It's clear what "modules" means here with the example, but I don't think it's clear without it. What about "included modules" or "included and extended modules"?

nickrivadeneira:
@gylaz Out of curiosity, what's the reason for this preference?

style/rails/README.md
mcmire: Agreed 👍

tabfugnic:
I'm with @nickrivadeneira on this. Maybe, it's just the way I've always done it, but putting the `includes/extends` looks more correct to me. I'm just curious what the reasoning is.

gylaz:
> Maybe, it's just the way I've always done it, but putting the `includes/extends` looks more correct to me.

@tabfugnic Putting it where?

gylaz:
@nickrivadeneira Based on our current guidelines, most of us have been putting constants first in the file. So, this felt natural. What is the proposed alternative? Putting `include/exclude` first?

tabfugnic:
Sorry, I meant to say on the top

nickrivadeneira:
@gylaz Don't have a proposed alternative - just curious if there's a benefit to either method. I've typically put `include`/`extend` first because the top felt like the "composition and inheritance" section. This is where one could look to find external classes/modules that are lending behavior to the given class. Below that would be all the internal implementation.

I could see it argued the other way, that since a class definition is actually a constant assignment, all constant assignments belong together at the top.

No strong preference either way, but I do think it makes sense to have a guideline on it.

gylaz:
> This is where one could look to find external classes/modules that are lending behavior to the given class.

This is a solid point. We might need to hold a vote to see what folks prefer.

gylaz:
This is stale, and I'm just going to close it.

PR #437, opened 2016-07-19
Trailing comma in Sass maps

tysongach:
This makes for cleaner diffs when adding to or
removing from the map.

This also aligns with [our JavaScript guideline](https://github.com/thoughtbot/guides/blob/master/style/javascript/sample.js#L11).


creuter:
I like it

whmii:
dope

PR #439, opened 2016-07-22
Add coffeescript config file + update hound config

jcmorrow:
Now that this repo is the owner configuration repo for all thoughtbot
repositories in Hound it needs to explicitly define a set of
coffeescript styles.

Changes:

 - Copy over `coffeelint.json` from Hound repo
 - Configure coffeescript in `.hound.yml` to be enabled and to point at
   `coffeelint.json`


gylaz:
LGTM.

PR #441, opened 2016-07-28
Disable UrlFormat linter

tysongach:
This linter is a bit too generalized to have enabled by default.

For example, on the thoughtbot blog we use a CDN for images, so having
an external domain with protocol makes sense, but this linter will throw
a warning.

Reference: https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#urlformat

### DIFF
--- a/style/sass/.scss-lint.yml
+++ b/style/sass/.scss-lint.yml
@@ -231,7 +231,7 @@ linters:
     enabled: true
 
   UrlFormat:
-    enabled: true
+    enabled: false
 
   UrlQuotes:
     enabled: true


whmii:
This PR is to big, can you break it up in to smaller commits

tysongach:
@whmii WILL do.
<br>
<br>
<br>
<br>
<br>
<br>
See I did there?

gerrywastaken:
@tysongach @whmii The author of scss-lint was aware of external resources and has good reasoning for suggesting that you shouldn't be including the domain.

see: https://github.com/brigade/scss-lint/issues/401

tysongach:
@gerrywastaken Oh, nice! Thanks for sharing that. I think that’s a really good point.

I may look into creating such a function for Bourbon…

PR #445, opened 2016-08-30
Update SCSS-Lint config to 0.50

tysongach:
https://github.com/brigade/scss-lint/releases/tag/v0.50.0

### DIFF
--- a/style/sass/.scss-lint.yml
+++ b/style/sass/.scss-lint.yml
@@ -1,4 +1,4 @@
-# Up-to-date with SCSS-Lint v0.49
+# Up-to-date with SCSS-Lint v0.50
 
 scss_files: "**/*.scss"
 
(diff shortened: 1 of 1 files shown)


style/sass/.scss-lint.yml
```diff
@@ -181,6 +181,11 @@ linters:
     enabled: true
     style: one_space
 
+  SpaceAfterComment:
+    enabled: true
+    style: one_space
```
whmii: Does this require a space after comments' `//`? I wonder if this would conflict with sassdocs's `///` syntax.

tysongach: Yes, and yes it does not like SassDoc’s `///` syntax. But we don’t use SassDoc on everyday projects, just Bourbon and Neat and other libraries. So you can disable it specifically when using SassDoc.

wardpenney:
👍

PR #447, opened 2016-09-12
Fix links in the README.md file

joseVFernandez:
Changes from relative links to absolute ones

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,18 +3,18 @@ Guides
 
 Guides for getting things done, programming well, and programming in style.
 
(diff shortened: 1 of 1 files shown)


iwz:
Thanks for spotting this!

Instead of absolute URLs, I wonder if relative URLs like `./` would work?

joseVFernandez:
Let me try! :)

joseVFernandez:
@iwz indeed ```./``` work! PR updated :)

iwz:
Thanks, Jose!

PR #451, opened 2016-11-01
Dis-encourage use of Sass glob imports

wardpenney:
@jaclynperrone and I just ran into this first bullet (new to me). I've historically avoided doing it for the other bullets. 

* It burries errors in development (shows generic 500 page)
* It is gem-dependant. `sass-rails` handles it, or `sass-globbing` gem
* Developer loses nice-looking history of `@import` adds

Also, in legacy codebases, converting **to** a glob import can result in specificity changes due to load-order dependency.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -60,3 +60,4 @@
 * Use [Normalize](https://github.com/necolas/normalize.css) for browser rendering consistency, rather than a reset.
 * Use HTML structure for ordering of selectors. Don't just put styles at the bottom of the Sass file.
 * Avoid having files longer than 100 lines.
+* Do not use `/*` or `**/*` glob imports.


gylaz:
What is the alternative? List out all the imports and their subdirectories?

tysongach:
Though glob importing is super nice because you don’t have to think about them, these are great points. I also did not know about bullet 1!

I think one strong point _for_ using glob imports is that I have found it will “sniff out” bad architectural decisions. If you’ve not wisely broken up partials and imports into a logical, sequential   order, importing becomes a nightmare. So since glob importing is basically alphabetical and not user defined, it forces you to think about the chain of high-level project Sass and going down to specific component-level Sass. Basically, your point on…

> …converting to a glob import can result in specificity changes due to load-order dependency.

…is a smell that the CSS is poorly organized, architecturally.

👍  I’m good with this addition, though, as I think the benefits outweigh this indirect benefit.

brendastorer:
I'm not sure I would want this to be in our guidelines, unless we feel that having consistency here is important. I have moved towards the glob import recently for the reasons @tysongach mentioned and for the overall ease of it, but have been frustrated by some of the things that @wardpenney mentioned. It feels to me that there are advantages and disadvantages to each, and maybe should just be a choice made on a per project basis. Interested to hear more designers take on this.

tysongach:
> It feels to me that there are advantages and disadvantages to each, and maybe should just be a choice made on a per project basis. Interested to hear more designers take on this.

This is a great point and made me realize that the most important part of this PR is the list of advantages and disadvantages that Ward points out. _Those_ help me, as a developer, make an informed decision. Just declaring a do/don’t leads me to ask “why?”

tysongach:
@wardpenney Any update here?

tysongach:
@wardpenney Can we get this updated and merged?

wardpenney:
@tysongach @brendastorer I agree 100% that sibling-level files should be non-dependent on load order. One way to do that is to alphabetize the import list, an arbitrary choice that reveals bad CSS when regressions create. It is also arbitrary (as is the glob) and may fail to reveal that problem. 

Do you think I should put the pros/cons into it? Or is this thread enough because it is attached to this commit and PR? For example, every declaration in this syntax guide is a declarative statement, not a reasoned argument. I only put it in here because it keeps causing problems. 

I'll close for now.

tysongach:
@wardpenney I think this is valid addition and should get merged.

Joe has an [open PR](https://github.com/thoughtbot/guides/pull/473) adding annotations to the Ruby guides. I think that does a good job at tying a guideline back to the story behind it. GitHub doesn't have a nice way for that linking, though.

wardpenney:
Well, @brendastorer mentioned that it could be a choice on a per-project basis. That works, too. This guide may not be the place for it. This is not whitespace, case, or other syntax preference. So maybe this isn't the place for it.

PR #454, opened 2016-11-25
Remove broken readability link

budmc29:
The better option would be to replace the readability link with the correct one, but I don't know which one is that.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -12,9 +12,7 @@ General
 * Don't swallow exceptions or "fail silently."
 * Don't write code that guesses at future functionality.
 * Exceptions should be exceptional.
-* [Keep the code simple].
-
-[Keep the code simple]: http://www.readability.com/~/ko2aqda2
+* Keep the code simple.
 
 Object-Oriented Design
 ----------------------


best-practices/README.md
```diff
@@ -12,9 +12,7 @@ General
 * Don't swallow exceptions or "fail silently."
 * Don't write code that guesses at future functionality.
 * Exceptions should be exceptional.
-* [Keep the code simple].
-
-[Keep the code simple]: http://www.readability.com/~/ko2aqda2
+* Keep the code simple.
```
bf4: wow, looks like those have been there forever https://github.com/thoughtbot/guides/commit/04b6e2ea8597ae46b24247b96da9b04645478fa0

iwz: @jferris any idea where this link went?

croaky: I wonder if we linked to the Rich Hickey "Sinple Made Easy" talk to describe "simplicity" in code?

https://github.com/matthiasn/talk-transcripts/blob/master/Hickey_Rich/SimpleMadeEasy.md

Cool transcript version ^

croaky:
Thanks @budmc29! Merged as 752db69

PR #455, opened 2016-12-15
Rails: avoid selecting all columns in SQL views

stevenpetryk:
In SQL views, it can be tempting to grab all of the columns from a table the view relies on:

```sql
SELECT messages.*, notifications.*
FROM ...
```

However, when the database freezes the view's definition, these columns get expanded based on the current state of the database:

```sql
SELECT
messages.id,
messages.content,
messages.user_id,
notifications.id,
notifications.device_id,
notifications...
```

Which causes the view to become very fragile, since changes in any referenced tables can break it (even if they don't change columns the view _actually_ needs). Instead, views should opt to only select columns they specifically need.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -160,6 +160,7 @@ Relational Databases
 
 * [Index foreign keys].
 * Constrain most columns as [`NOT NULL`].
+* In a SQL view, only select columns you need (i.e., avoid `SELECT table.*`).
 * Use an `ORDER BY` clause on queries where the results will be displayed to a
   user, as queries without one may return results in a changing, arbitrary
   order.


derekprior:
As a point of clarification, this isn't something Scenic does. It is done by Postgres (and other DBs) at view creation time.

thorncp:
:+1: to this guidance. Database views aren't specific to Rails, though, so I don't think it fits in that section. We don't have a general SQL heading, so perhaps it belongs under Postgres?

mike-burns:
That Postgres section could be split into "Postgres" and "Relational Database": things like "Index foreign keys", "Constraint most columns as NOT NULL", "Use an ORDER BY, ..." apply to more than just Pg.

stevenpetryk:
Wow—totally didn't notice the Postgres section. But yes, I agree will all of this. I'm thinking the DB/Postgres split should be done in another PR though.

@derekprior, yeah, I realized that shortly after I posted the PR (you may have received the pre-edited text in your inbox). Thanks for the info!

Currently on vacay, but I'll open that other PR before Friday and base this on top of it.

stevenpetryk:
Alrighty, this is now based on #461, so it'll have to be rebased after that's merged to skip 166e161.

thorncp:
@stevenpetryk #461 has been merged. Do you mind rebasing this/fixing the conflict?

thorncp:
Thanks @stevenpetryk!

PR #458, opened 2016-12-21
Remove socialist sentences

b4k3r:
In my opinion everyone should be responsible for his own code. Why someone should be responsible for someone else's job? In this way lets make one github account and use one ssh. Your code? Yey, is also my.

> Assume everyone is attractive, intelligent, and well-meaning"

 yeah, and assume that world is wonderful and safe.

> Don't take it personally. The review is of the code, not you.

So who wrote this code? Ghost, God?

### DIFF
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -11,17 +11,14 @@ Everyone
 * Ask questions; don't make demands. ("What do you think about naming this
   `:user_id`?")
 * Ask for clarification. ("I didn't understand. Can you clarify?")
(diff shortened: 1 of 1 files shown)


derekprior:
Thanks for your contribution. I've wondered about the need to include the assumption of attractiveness, but the rest of the lines you have taken issue with have worked really well for us as guidelines for code review.

gfontenot:
I've opened #459 to address the weird line about attractiveness. Thanks for pointing that out!

PR #459, opened 2016-12-21
Remove weird comment about attractiveness

gfontenot:
This is out of place and borderline inappropriate. There's really no
need to reference physical appearance here, so how about we just don't.


derekprior:
:shipit:

PR #465, opened 2017-01-04
Add ESLint Hound configuration

purinkle:
Before, Hound had its own default ESLint configuration. We are moving these configurations into specific organisations. Added a thoughtbot-specific Hound configuration for ES6 files.

https://trello.com/c/HY2hoTdq


.hound.yml
```diff
@@ -1,6 +1,9 @@
 coffeescript:
   enabled: true
   config_file: style/coffeescript/coffeelint.json
+eslint:
+  enabled: true
+  config_file: style/eslint/.eslintrc
```
teoljungberg: There already seems to be an ESLint configuration present here - https://github.com/thoughtbot/guides/blob/master/style/javascript/.eslintrc.json. Maybe that should be used instead?

purinkle: Resolved in 0b959f6

PR #466, opened 2017-01-23
Remove HTML best practices

tysongach:
These two guides were introduced in https://github.com/thoughtbot/guides/commit/2360fac0c44de00cd37607cdc72b24128adfc7a7 (carried over from the old thoughtbot development handbook).

Does anyone know why these should be considered best practices? I’m not aware of a reason why reset buttons should not be used.

I recently worked on a project which had a filtering features, and it was a situation where it could make sense to have a “Clear filters“ button that would reset the filters form.

- "Don't use a reset button for forms." feels like too strong of a statement and
casts a wide net.
- It is unclear why these are best practices.
- These guidelines are highly dependent on the situation. 
- They lack contextual information on when these best practices should
be adhered to or broken.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -212,8 +212,6 @@ JavaScript
 HTML
 ----
 
-* Don't use a reset button for forms.
-* Prefer cancel links to cancel buttons.
 * Use `<button>` tags over `<a>` tags for actions.
 
 CSS


mike-burns:
I bet this was from the time when all forms on the Web had reset buttons but browsers didn't have great undo functionality and so we kept losing data accidentally and this frustration ended up in the guides.

So, +1 to giving designers the freedom to do what's right.

> They lack contextual information on when these best practices should
be adhered to or broken.

Far too much of this repo is like that.

croaky:
Looks good to me.

nickrivadeneira:
TIL reset buttons are a thing

tysongach:
> Far too much of this repo is like that.

Yes.

jferris:
> Far too much of this repo is like that.

You can usually use Git blame to find the discussion around when a change was introduced. If that's not working, and people are generally confused about why guidelines are in place, we can change the format of the guides to includes rationale or links to discussion.

nickrivadeneira:
@jferris 👍  for rationales/links to discussion. The git blame can get a bit hairy when the organization of files changes ([example](https://github.com/thoughtbot/guides/commit/e5532a17a52d6c183ef9b13038db119c3d010ab6#diff-3241340df9b71a19cbc1ede342ad226b)).

Additionally, having rationales as a living part of the guides makes it easier to continue to reevaluate a guideline's merits over time.

PR #470, opened 2017-01-23
Drop IE10 from browser support best practice

tysongach:
Can I Use (with data from StatCounter) reports that global IE10 usage has
dropped to 0.24%: http://caniuse.com/usage-table.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -241,7 +241,7 @@ Sass
 Browsers
 --------
 
-* Avoid supporting versions of Internet Explorer before IE10.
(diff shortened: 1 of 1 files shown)


tysongach:
Ah, the Playbook also references what we support, so I will update the info in a PR to that, as well: https://github.com/thoughtbot/thoughtbot.com/blob/master/app/playbook/choose-platforms/rails-gets-web-products-to-market-quickly.md

tysongach:
PR to update the Playbook: https://github.com/thoughtbot/thoughtbot.com/pull/1137

croaky:
Merge at will

PR #472, opened 2017-01-27
Expand "security" to physical + vulnerabilities

croaky:
Based on similar sections seen in compliance questionnaires,
re-title existing sections and add new sections:

* Secure Employee Access and Communication
* Application Security: Transmitting Information
* Application Security: Storing Information
* Physical Security
* Handling Vulnerabilities: Reporting
* Handling Vulnerabilities: Reviewing, Logging and Following Up

Extract the new sections from internal handbook.

### DIFF
--- a/security/README.md
+++ b/security/README.md
@@ -1,10 +1,8 @@
-Security
-========
+# Security
 
 A guide for practicing safe web.
 
-Think
------
+## Think
 
 Security is important, and you can't practice these guidelines without
 understanding them. Make sure you understand each guideline, why it exists, and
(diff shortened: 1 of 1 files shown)


security/README.md
```diff
@@ -13,36 +11,21 @@ how to follow it.
...
+
+The following guidelines apply to how you as an individual
+secure access to your systems (laptop, accounts, etc.)
+and communication (email, etc.).
+
+### Using Passwords
 
 * Use a unique password for every account you create.
 * Use a tool like [pwgen](https://github.com/jbernard/pwgen) or
   [1password](https://agilebits.com/onepassword) to generate random passwords.
```
tysongach: While you’re in here, this is now simply https://1password.com/.

security/README.md
```diff
@@ -55,3 +38,62 @@ PGP and GnuPG
...
+
+## Handling Vulnerabilities
+
+The following guidelines apply to how we handle security incidents.
+
+### Reporting
+
+When someone finds a possible security issue in our software,
+we encourage them to report it to our <security@thoughtbot.com> email address.
+This email address is an alias for <everyone@thoughtbot.com>.
```
geoffharcourt: Is putting this here going to result in that alias getting spammed? Is it enough to write the address out in a slightly obfuscated way and then to say, "it reaches everyone at thoughtbot directly"?

security/README.md
```diff
@@ -55,3 +38,62 @@ PGP and GnuPG
...
+  traffic.
+* Use secure cookies.
+* Avoid protocol-relative URLs.
+
+### Storing Information
+
+* Don't log passwords.
+* Don't store passwords in plain text.
+* Don't hash passwords using a reversible cipher.
+* Don't hash passwords using a broken cipher, such as MD5 or SHA1.
```
geoffharcourt: Should we discuss considering the place your data is stored?

BlakeWilliams:
Should we mention encrypting your laptop with FileVault or your platform's equivalent?

geoffharcourt:
+ firmware passwords.

derekprior:
> Should we mention encrypting your laptop with FileVault or your platform's equivalent?

Yes. This should be a requirement. We all have client data and source codes on our laptops. Not encrypting that is irresponsible.

security/README.md
```diff
@@ -55,3 +38,62 @@ PGP and GnuPG
...
   secure location, such as a thumb drive.
+
+## Physical Security
+
+The following guidelines apply to how we physically secure
+our laptops and mobile devices that may contain customer or user data.
+
+* Don't leave your devices unattended in an unsecured area.
+* Install a device tracking and remote data wipe tool such as [Prey].
+
```
mike-burns: Should we add:

* Lock your device when you are away from it.

On the one hand, people don't currently do this reliably, so it might be odd to document a thing that we don't do. On the other hand, we should be doing this.

BlakeWilliams:
It's pretty intrusive, but I use https://www.obdev.at/products/littlesnitch/index.html which is pretty great and might deserve a mention if we can fit it in somewhere.

security/README.md
croaky: Not sure? Would prefer to do that in a follow-on PR if desired. This list is the same as before, I mostly just moved it.

I did add one new item to the top: "Don't log passwords." I meant things like `filter_parameter_logging` in Rails. If anyone has thoughts on that, or better wording, that'd be something to give a stink-eye.

croaky:
Think I got most of the feedback.

Didn't understand the "firmware passwords" comment and not sure how to generalize the Little Snitch program. Okay to leave those to others for follow-on PRs if you feel passionately and can better describe / summarize them than I?

geoffharcourt:
@croaky sorry about the vagueness of that. I was advocating for people having a firmware password on their laptop even if they also have disk-level encryption.

PR #473, opened 2017-01-27
Annotate Ruby style guide with references

jferris:
This adds a reference for each Ruby style guideline to the pull request (when
available) or the original commit which introduced that guideline.

This allows a reader to quickly jump to the discussion that led to a particular
guideline being introduced without having to trace through Git history.

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -3,47 +3,51 @@ Ruby
 
 [Sample](sample.rb)
 
-* Avoid conditional modifiers (lines that end with conditionals).
-* Avoid multiple assignments per line (`one, two = 1, 2`).
-* Avoid organizational comments (`# Validations`).
+* Avoid conditional modifiers (lines that end with conditionals). 36491dbb9
+* Avoid multiple assignments per line (`one, two = 1, 2`). #109
+* Avoid organizational comments (`# Validations`). #63
 * Avoid ternary operators (`boolean ? true : false`). Use multi-line `if`
-  instead to emphasize code branches.
-* Avoid explicit return statements.
-* Avoid using semicolons.
-* Avoid bang (!) method names. Prefer descriptive names.
+  instead to emphasize code branches. 36491dbb9
+* Avoid explicit return statements. #48
+* Avoid using semicolons. #108
+* Avoid bang (!) method names. Prefer descriptive names. #122
 * Don't use `self` explicitly anywhere except class methods (`def self.method`)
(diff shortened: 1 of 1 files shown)


jferris:
Based on the discussion in #466, I decided to see what it would take to track down all the original conversations involved in creating the guides. The Ruby style guide is one of the larger and older guides, so I think it serves as a good test bed.

If people like this approach, we can apply it to the rest of the guides to make history available everywhere. Introducing pull request links to future guidelines will require creating the pull request and then updating the commit, which is somewhat annoying. I'm open to other ideas for how to reference the conversation, but it seems like making it easy to trace back to a pull request is the lightest and most thorough way to ensure that you can see the thinking behind a guideline.

backus:
I was actually going to suggest you guys do this after seeing this comment https://github.com/thoughtbot/guides/pull/466#issuecomment-275693933. RuboCop annotates its changelog like this which seems very helpful. It is also pretty easy to programmatically enforce: https://github.com/bbatsov/rubocop/blob/master/spec/project_spec.rb#L117-L147

mike-burns:
You're my hero.

style/ruby/README.md
```diff
@@ -3,47 +3,51 @@ Ruby
 
 [Sample](sample.rb)
 
-* Avoid conditional modifiers (lines that end with conditionals).
-* Avoid multiple assignments per line (`one, two = 1, 2`).
-* Avoid organizational comments (`# Validations`).
+* Avoid conditional modifiers (lines that end with conditionals). 36491dbb9
```
croaky: Does the short SHA auto-link? Docs don't list it but would be nice if so.

https://help.github.com/articles/autolinked-references-and-urls/

Doesn't look like short SHA or pound-number working in "View" on this PR but pound-number listed in docs as autolinking.

jferris: Yeah, it seems like they autolink in comments, pull request messages, etc, but maybe not in Markdown documents?

Links would be nice, but I'm not sure if it's worth the readability tradeoff in the source document. Thoughts?

croaky: Yeah, maybe leave them as-is and we ask GitHub to add the auto-linking to MD docs.

geoffharcourt:
Given that we're OK on approach here, @jferris, feel like this is mergeable?

jferris:
@geoffharcourt thanks for the reminder - this is now MERGED

PR #476, opened 2017-03-02
Encourage security alert tools in CI builds

croaky:


mike-burns:
- Yes, I like this writing.
- Not totally sure this is the right section -- this section deals more with "how to handle security issues in our software that are reported to us", and less of "how to run automated tooling that reports potential issues". Perhaps a new section would be good, or maybe this can go elsewhere above?

croaky:
What do you think of b4f7675 @mike-burns?

jferris:
I agree with @mike-burns - maybe adding a new section for "Preventing Vulnerabilities" would be good? It seems valuable to have a section exclusively dedicated to handling incidents.

croaky:
I guess I was thinking of Bundler Audit as handling potential incidents. It highlights a known vulnerability in a library. May not be exploitable, though? Was thinking of CI as another reporting channel, input about actual vulnerabilities, less about prevention of future incidents? I guess because it's not running against known production code, though, maybe better described as prevention.

croaky:
Created a "Preventing Vulnerabilities and Regressions" h3 within "Application Security" h2

mike-burns:
Yup, love this change. :+1: from me.

PR #478, opened 2017-03-14
Explicit `dependent` on ActiveRecord associations.

georgebrock:
I recently added some missing `dependent: :destroy` settings to associations in a Rails project. I got the following comment on the PR:

> Aside: we (as a person/team/company) should figure out how and when and why to put `dependent: :destroy` on associations. I certainly leave them off, as do others.

I'm as guilty of this as anyone else. It has bitten me in various projects, either because orphaned data was left behind, or more recently—in these glorious days of database-level foreign key constraints—because I couldn't easily delete data from my development environment.

Let's add this to the guides, and ultimately to Hound, so that we get PR comments that make us think about the right thing to do.

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -33,8 +33,10 @@ Migrations
 
 * Set an empty string as the default constraint for non-required string and text
   fields. [Example][default example].
+* Set an explicit [`on_delete` behavior for foreign keys][add_foreign_key].
 
 [default example]: migration.rb#L6
+[add_foreign_key]: http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_foreign_key
 
 Routes
 ------
--- a/style/rails/migration.rb
+++ b/style/rails/migration.rb
@@ -4,8 +4,10 @@ class CreateClearanceUsers < ActiveRecord::Migration
       t.timestamps null: false
       t.string :email, null: false
       t.string :name, null: false, default: ''
+      t.references :company
     end
 
     add_index :users, :email
+    add_foreign_key :users, :company_id, on_delete: :restrict
   end
 end


mike-burns:
As the metacomment-leaver: yes.

Is there something that can drive this? The only reason we noticed is because we imported some bad data. Is there a kind of test we should be writing more often? What patterns should we follow to catch problems around `:destroy`?

derekprior:
I'd prefer to not specify it, opting to leave it as a property of the foreign key, which avoids many of the potential problems with `:destroy`

mike-burns:
derekprior, you mean pushing it into the DB schema?

derekprior:
Yes, the database level foreign key.

You can specify it in both locations, if desired, but that defines a callback that we know will never run. It costs us a database round trip on every delete (admittedly, highly unlikely to be of any performance concern).

If we're going to specify `on_delete: :cascade` on the FK and also add a redundant definition to the association, I think we should encourage `:delete_all`, as at least it's clear there that no callbacks will be run. I'd only use `destroy` for cases where we aren't using the FK like this because we *must* have the callbacks.

georgebrock:
A guides/Hound rule that calls out migrations creating foreign keys without specifying `on_delete` sounds even better than one that calls out associations without `dependent`.

georgebrock:
PR updated to talk about databases, not models. Thanks, @derekprior

georgebrock:
As an interesting aside, Django's changing its default behaviour soon to require an explicit `on_delete`: https://docs.djangoproject.com/en/1.10/ref/models/fields/#arguments

geoffharcourt:
+1 on using FK and `CASCADE` 

I greatly prefer to handle this in the database and to avoid having it be part of Rails and callbacks.

wardpenney:
+1 to database FK.

style/rails/README.md
```diff
@@ -33,6 +33,7 @@ Migrations
 
 * Set an empty string as the default constraint for non-required string and text
   fields. [Example][default example].
+* Set an explicit `on_delete` behaviour for foreign keys.
```
croaky: Maybe link to http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_foreign_key docs?

georgebrock:
Link added, and “behaviour” changed to the en-US “behavior” for consistency with the rest of the doc.

I'll merge this tomorrow.

PR #479, opened 2017-03-14
Prefer curly braces in Ruby

mike-burns:
The big difference between `do`/`end` and `{`/`}`, with regards to
blocks, is [precedence].

As a simplification, let's pick one and always use it. How about curly
braces (`{`/`}`) as that one?

Due to precedence, this will not work appropriately for `#it` as we
currently write it. We could use parens around `#it`'s argument:

```ruby
it("redirects to the home page") {
  get :new
  expect(response).to redirect_to(root_url)
}
```

Or we could make it a "prefer" instead of "use", allowing for the
responsible freedom that comes with Ruby's syntax, but clearly stating
our preference.

[precedence]: https://robots.thoughtbot.com/do-end-your-braces

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -29,7 +29,8 @@ Ruby
 * Prefix unused variables or parameters with underscore (`_`).
 * Use a leading underscore when defining instance variables for memoization.
 * Use `%{}` for single-line strings needing interpolation and double-quotes.
-* Use `{...}` for single-line blocks. Use `do..end` for multi-line blocks.
+* Prefer `{...}` for all blocks. Use `do..end` for the special case of passing
+  a block to a method whose arguments should not have parentheses around them.
 * Use `?` suffix for predicate methods.
 * Use `CamelCase` for classes and modules, `snake_case` for variables and
   methods, `SCREAMING_SNAKE_CASE` for constants.


gylaz:
I think using `{}` for blocks everywhere would go against idiomatic Ruby. IMO, we'd be safer to say prefer/use `do...end`.

Alternatively, we could say `Prefer {} in the these situations: ...`. (we might already have some of this?)

odlp:
I think this would put us at odds with large chunk of the Ruby community (compatibility is a virtue?). I appreciate the blog post link - I wasn't aware of the precedence aspect.

+1 to:

> Prefer {} in the these situations: ...

mike-burns:
The other option (which Greg hinted at) is to go vertical: only use `do`/`end`. This does seem consistent with our overall Ruby style: newlines around lines with blocks, newlines around `if`, no `if` modifier, etc.

purinkle:
There is always [The Weirich Rule]

> The semantic rule says use {} for blocks where the primary purpose of the block is to return a value, use do/end for blocks where the primary purpose of the block is to execute side effects.

[The Weirich Rule]: https://github.com/bbatsov/ruby-style-guide/issues/162

mike-burns:
Closing this PR because I don't actually want more rules.

PR #480, opened 2017-04-03
Trying to reword potentially de-humanizing guideline

sardaukar:
It has been brought to my attention by more than one person that saying "The review is of the code, not you" can be perceived as a way to dehumanize the process of code review and to focus on cold, hard logic ignoring the fact that there's a person on the receiving end of one's feedback.

That was not the original intent, so this attempts to change that guideline's text to instead try to implicitly bring forth the notion that there are emotions involved, and that everyone should be mindful of that. 

Not 100% sure the external link or the current text in the commit are the best way, open to suggestions.

### DIFF
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -31,7 +31,8 @@ Having Your Code Reviewed
 
 * Be grateful for the reviewer's suggestions. ("Good call. I'll make that
   change.")
-* Don't take it personally. The review is of the code, not you.
+* Be aware of [how hard it is to convey emotion online](https://www.fastcodesign.com/3036748/why-its-so-hard-to-detect-emotion-in-emails-and-texts) and how easy it is to
+  misinterpret feedback.
 * Explain why the code exists. ("It's like that because of these reasons. Would
   it be more clear if I rename this class/file/method/variable?")
 * Extract some changes and refactorings into future tickets/stories.
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -31,7 +31,7 @@ Having Your Code Reviewed
 
 * Be grateful for the reviewer's suggestions. ("Good call. I'll make that
   change.")
-* Be aware of [how hard it is to convey emotion online](https://www.fastcodesign.com/3036748/why-its-so-hard-to-detect-emotion-in-emails-and-texts) and how easy it is to
+* Be aware of [how hard it is to convey emotion online] and how easy it is to
   misinterpret feedback.
 * Explain why the code exists. ("It's like that because of these reasons. Would
   it be more clear if I rename this class/file/method/variable?")
@@ -46,6 +46,8 @@ Having Your Code Reviewed
 * Wait to merge the branch until Continuous Integration (TDDium, TravisCI, etc.)
   tells you the test suite is green in the branch.
 * Merge once you feel confident in the code and its impact on the project.
+
+[how hard it is to convey emotion online]: https://www.fastcodesign.com/3036748/why-its-so-hard-to-detect-emotion-in-emails-and-texts
 
 Reviewing Code
 --------------
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -22,8 +22,8 @@ Everyone
 * Don't use sarcasm.
 * Keep it real. If emoji, animated gifs, or humor aren't you, don't force them.
   If they are, use them with aplomb.
-* Talk synchronously (e.g. chat, screensharing, in person) if there are too many 
-  "I didn't understand" or "Alternative solution:" comments. Post a follow-up 
+* Talk synchronously (e.g. chat, screensharing, in person) if there are too many
+  "I didn't understand" or "Alternative solution:" comments. Post a follow-up
   comment summarizing the discussion.
 
 Having Your Code Reviewed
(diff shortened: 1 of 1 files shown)


kitation:
I think this change still put the onus on the person who is feeling they have been insulted or dehumanised. I think the rest of the guidelines are good enough that you could take this line out altogether, or perhaps add the new line with link to the 'Everyone' section?

nickrivadeneira:
> "The review is of the code, not you" can be perceived as a way to dehumanize the process of code review and to focus on cold, hard logic

That seems to me like the ideal. Code reviews work best when they assess the code written and not the author.

> I think this change still put the onus on the person who is feeling they have been insulted or dehumanised.

It seems ok to me to provide guidance for code authors on how to be receptive to feedback and have a positive perspective on discussion.

mxie:
> Code reviews work best when they assess the code written and not the author.

In theory, sure, but in practice, I'm not so sure. The code you write is usually representative of your knowledge of how something works and how you choose to approach the problem at task. I think it can be challenging to separate yourself from the thing that you produced. Imagine the scenario where the code you wrote had broken something terribly on production. Do you tend to feel guilty about it or are you able to completely detach yourself from what happened?

I'm not saying that feedback in general is meant to be an attack on the person, but I do think that it's important to remember that there is a human on the other end of this.

That said, I do think that @kitation's suggestion about adding a line under "Everyone" would be more appropriate.

nickrivadeneira:
> I think it can be challenging to separate yourself from the thing that you produced.

Definitely agree with that. I think the existing guideline is intended to speak directly to that challenge. Of course, "Don't take it personally" is easier said than done, but the reminder that "the review is of the code, not you" is helpful.

What's the core problem the change is intended to solve? I haven't found code reviews to be dehumanized. I can't speak to anyone else's experience, so I'm curious how others feel having followed this guideline. Past experience can be more compelling than possible perceptions of guidelines.

> this attempts to change that guideline's text to instead try to implicitly bring forth the notion that there are emotions involved, and that everyone should be mindful of that.

My vote would be for keeping the existing guideline and adding a guideline to the "Everyone" section (as suggested by others) that is explicit rather than implicit. The notion described above makes sense to me, so it could be:
```
Code reviews are between people. Be mindful of others' emotions.
```

sardaukar:
I haven't had a bad experience with code reviews, yet. But I understand how stressful it can be. We can add the suggested mention under the "Everyone" section, but it seems lacking to not mention the emotional issue being discussed to the guidelines on code reviews.

nickrivadeneira:
@sardaukar I might be misunderstanding the emotional issue - can you talk a bit more about it?

sardaukar:
It might be stressful to give out proper feedback after a long week, or a string of particularly challenging exchanges on other pull requests, or the person on the receiving end might be having a bad day or be generally anxious about feedback. This has the potential to be a bad experience for all involved, so guidelines on code review should (IMHO) mention it.

nickrivadeneira:
Makes sense (and I think is good advice for any interaction, not just code review).

For better or worse, these guides typically prescribe a solution rather than detailing the issue. I think the issue you describe could be addressed with something simple like "Be nice", "Be empathetic", or "Be mindful". This could then be annotated with a link to this PR as is done in #473. I've also found the suggested removal to be useful advice in the past, so would be sad to see that removed. All up to you though.

wildfalcon:
I believe this is an important subject, with nuance and subtlety, and it's worth getting right. We are all human with a messy mix of emotions, and we need to be aware of how to work with that. We all care about what we do, and that results in some level of emotional investment in what we produce. 

I don't believe that, even in an ideal world, we should be separating ourselves from the emotional investment we put into our work. To do so would be to spend time pretending to be someone we are not.  However, we should also be aiming to have the emotional maturity to be able to cope with someone critiquing our work. 

I'm in favor of wording that recognizes this is a sensitive subject and to raise awareness on both sides that giving and receiving feedback is a skill that takes time to develop.

kitation:
I think if you accept the premise that a person can't be separated from their work, then `The review is of the code, not you` just isn't true. For background, I just did a talk about mental health in the workplace and how the "Don't take it personally" advice is very hard to apply in my situation, and I do have experience of it being used against me to shoot down issues I was having with a reviewer.

My question then is what does this line add to the guidelines that the other lines don't cover?

code-review/README.md
```diff
@@ -31,7 +31,8 @@ Having Your Code Reviewed
 
 * Be grateful for the reviewer's suggestions. ("Good call. I'll make that
   change.")
-* Don't take it personally. The review is of the code, not you.
+* Be aware of [how hard it is to convey emotion online] and how easy it is to
+  misinterpret feedback.
```
bf4: Maybe something like?

> A common axiom is "Don't take it personally. The review is of the code, not you."
> We used to include this, but now prefer to say what we mean:
> Be aware of [how hard it is to convey emotion online] and how easy it is to  misinterpret feedback.
> If a review seems aggressive or angry or otherwise personal, consider if it is intended to be read that way and ask the person for clarification of intent, in person if possible.

bf4: Sometimes people are mean in code reviews :(

regularfry: How about "Seek clarification of personal comments. The intended meaning may not have survived being typed."?  That suggests a positive action, rather than a change in attitude or an admonition against a thing, without leaving room for blame.

drapergeek: > A common axiom is "Don't take it personally. The review is of the code, not you."
> We used to include this, but now prefer to say what we mean:
Be aware of [how hard it is to convey emotion online] and how easy it is to misinterpret feedback.

I really like this. While I understand that people may not be able to divorce themselves from their code, we have a culture of shared success, achievement and failure. Once your code is merged, it's part of the team so reviewing the code and not the person, is the goal.

Maybe add something like :
> Assume the best intention from the reviewer's comments.

sardaukar:
Adding this in - further PRs welcome!

PR #484, opened 2017-04-10
SCSS-Lint config updates

tysongach:
- Remove options from disabled linters
- Enable `ImportPath` options
  - Throw warnings for leading underscores and file extensions in imports.
- Update SpaceAfterComment linter
  - Allow at least one space, instead of only one space.
- Add `text-decoration-skip` as a known property

### DIFF
--- a/style/sass/.scss-lint.yml
+++ b/style/sass/.scss-lint.yml
@@ -125,7 +125,8 @@ linters:
 
   PropertySpelling:
     enabled: true
-    extra_properties: []
+    extra_properties:
+      - text-decoration-skip
     disabled_properties: []
 
   PropertyUnits:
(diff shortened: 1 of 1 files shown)


style/sass/.scss-lint.yml
```diff
@@ -12,7 +12,6 @@ linters:
 
   BemDepth:
     enabled: false
-    max_elements: 1
```
wardpenney: I 👍  this. When https://github.com/sasstools/sass-lint/issues/413  started showing up in Hound comments, I didn't not like it. I understand some developers prefer only one level of BEM depth in order to allow enforce shorter class names, but I believe that is a band-aid that belies a strategic problem in the pattern itself. 

I prefer ancestral (non-direct descendant, just ancestral) embedding into BEM naming. The reason: it creates a balancing feedback loop for the developer to signal that if a class is too long, maybe there's more than one pattern here.

tysongach: I hear ya…but this linter is disabled any way, I just saw no reason to define an option that isn't even used.

PR #486, opened 2017-04-19
Revert "Enable ImportPath options"

tysongach:
This reverts commit dff58c897cfdcce720c1c0649938be661cc5dba9.

The value of `false` DOES mean that import paths should not have a
leading underscore or its file extension.

### DIFF
--- a/style/sass/.scss-lint.yml
+++ b/style/sass/.scss-lint.yml
@@ -78,8 +78,8 @@ linters:
 
   ImportPath:
     enabled: true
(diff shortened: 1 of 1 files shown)


creuter:
👍 

It'd be helpful to link to the linter docs on these types of PRs in the future.

tysongach:
Documentation (added to commit message, as well): https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#importpath

PR #487, opened 2017-05-04
ruby string interpolation syntax

bfauble:
update for ruby style guide to use correct `#{}` format for string interpolation instead of `%{}`


gylaz:
The `%{}` is correct. This describes defining the string, not interpolating a variable.
```rb
name = "Bond"
%{They call me "#{name}", James "#{name}}
```

bfauble:
@gylaz ah thank you for the clarification. Thoughts on rewording to `Use `%{}` for single-line strings containing double-quotes that require interpolation.` to remove the ambiguity that I misunderstood?

gylaz:
@bfauble I think that would be good. Want to open another PR?

PR #488, opened 2017-05-05
update single line string interpolation wording

bfauble:
update single line string interpolation wording to be less ambiguous as to what it's referring to.


gylaz:
Referencing #487 for more context.

mike-burns:
How about something like:

> Prefer `"` or `%{}` for strings.

This leaves it up to the dev to decide which of the two options look best.

cpytel:
This has been sitting for a while. Can we either decide to revise before merging, merging without revision, or closing without merging?

geoffharcourt:
@bfauble I rebased, cleaned up the conflict, and merged as fd3cb02. Thank you for your contribution!

PR #490, opened 2017-05-11
Change Elixir line length rules

drapergeek:
We're consistently running into issues, especially in tests, where it is overkill to break lines at 80 characters. Internally on Elixir projects I've worked on we've had a "soft limit" at 80 and a "hard limit" at 100. Unfortunately this is hard to monitor and the "false positives" from hound lead to me ignoring Hound which is less than ideal.

### DIFF
--- a/style/elixir/README.md
+++ b/style/elixir/README.md
@@ -9,3 +9,4 @@ Elixir
 * Don't indent pipes when not using the match operator (`=`).
 * Prefer pattern matching and guards in function definitions over conditionals
   in the function body.
+* Break long lines after 100 characters.


style/elixir/README.md
```diff
@@ -9,3 +9,4 @@ Elixir
 * Don't indent pipes when not using the match operator (`=`).
 * Prefer pattern matching and guards in function definitions over conditionals
   in the function body.
+* Break long lines after 80 characters.
```
derekprior: I'm confused. I thought you were saying 80 characters was overly restrictive?

drapergeek: Updated!

BlakeWilliams:
I agree with this change. 👍

PR #491, opened 2017-05-12
Add guide for splitting translations by controller

drapergeek:
We've tried this on a few projects and it's been very well received so I'd like to solidify the decision.

https://trello.com/c/FaxFdKxv/639-split-i18n-translations-into-individual-files-based-on-the-controller

### DIFF
--- a/style/rails/README.md
+++ b/style/rails/README.md
@@ -14,6 +14,7 @@ Rails
 * Order ActiveRecord associations above ActiveRecord validations.
 * Order controller contents: filters, public methods, private methods.
 * Order i18n translations alphabetically by key name.
+* Split i18n translations into files based on the controller name.
 * Order model contents: constants, macros, public methods, private methods.
 * Put application-wide partials in the [`app/views/application`] directory.
 * Use `def self.method`, not the `scope :method` DSL.


croaky:
What does the directory structure look like for an example project? What do you do for non-controller translations?

drapergeek:
Here's the project where we started this research: https://github.com/thoughtbot/momentum/tree/master/config/locales/en

As for non-controller based, I would tend to put them under the "relevant" controller if the model/controller matches. If it falls outside that, I'd say whatever the key name starts with,  put it in a file named the same.

derekprior:
I've broken them up by namespace before and this seems like a good next step. When breaking them up, I've put "generic" things in their own files: `helpers.en.yml`, `activerecord.en.yml`, `activemodel.en.yml`, `simpleform.en.yml`, etc.

cpytel:
This has been sitting for a while. Can we decide to merge it, revise and merge, or close without merging? Thanks.

style/rails/README.md
```diff
@@ -14,6 +14,7 @@ Rails
 * Order ActiveRecord associations above ActiveRecord validations.
 * Order controller contents: filters, public methods, private methods.
 * Order i18n translations alphabetically by key name.
+* Split i18n translations into files based on the controller name.
```
rickreilly: Since you asked, I'll give this a 👎 .

I don't think this is the right way to group translations. What do you do when the same translation applies to different controllers? Repeat it? What happens if you refactor your controllers; do you now need to repeat the process for your translations? I think the question of how to group translations and how to define controller endpoints are, in general, unrelated.

geoffharcourt: This is in opposition to our recommendations about factories. Is there a reason why one approach would be better for translations vs. factories?

I used to build a separate factory per model and found being able to quickly jump between items in one file was easier to handle.

derekprior: > This is in opposition to our recommendations about factories. Is there a reason why one approach would be better for translations vs. factories?

It's a LOT easier to make sense of a factory definition -- even a long when -- then it is to find a specific fully-qualified YAML key in YAML document with hundreds to thousands of lines. You can do things in YAML files that result in valid YAML but completely obliterate earlier YAML entries, for example. That's a bummer.

It turns out, YAML is really a terrible format for this stuff :)

derekprior: > I don't think this is the right way to group translations. What do you do when the same translation applies to different controllers? Repeat it?

For the most part, yes. It's my experience that "content" translations (as opposed to model field translations, error message translations, etc) tend to be hurt from deduplication. Sometimes the same english text on two pages would be translated to two different words in other languages depending on context, for example. I've found it simpler to not worry about DRYing up translations.

derekprior:
I started trying this out on a client project that has had a history of trying to overcome the single giant YAML file issue. At first we moved to files like `views.yml` for view translations, `activerecord.yml`, `helpers.yml`, etc. Eventually `views.yml` grew to a difficult size once again. The team reacted positively to splitting them out by controller but keeping things like `activerecord.yml`, `helpers.yml` for those overarching translations.

style/rails/README.md
rickreilly: >>I don't think this is the right way to group translations. What do you do when the same translation applies to different controllers? Repeat it?

>For the most part, yes. It's my experience that "content" translations (as opposed to model field translations, error message translations, etc) tend to be hurt from deduplication. Sometimes the same english text on two pages would be translated to two different words in other languages depending on context, for example. I've found it simpler to not worry about DRYing up translations.

I agree that it's possible to go too far in DRYing up translations, and that there's a balance to be struck, as with code. I don't think we should make our translations as DRY as possible, necessarily.

My point is that translations are purely presentational, and should therefore not be coupled to controllers, in principle. I was just citing duplication as one possible symptom of what I think the underlying problem is, which is more about separation of concerns.

>It's a LOT easier to make sense of a factory definition -- even a long when -- then it is to find a specific fully-qualified YAML key in YAML document with hundreds to thousands of lines

True, but could we solve that problem better by other means?

For example:

We face a similar problem in managing deeply nested CSS rules. In that case, we typically avoid the problem by flattening it out.

Ex., `.outside { .middle { .inside: ... } }` => `.outside__middle__inside {...}`.

If we did something analogous for translations, it might be easier to search for than either deeply nested YAML files, or YAML files based on controllers.

Anyway, whether we merge this or not, it's not the end of the world. I'm just personally not sure enough of this approach to support making it our "official" policy, and wanted to voice some reservations.

tysongach:
Should we close this? It’s been open for over a year and it appears as though we don’t have a consistent opinion on the matter.

PR #494, opened 2017-05-26
Simplify guides on over-qualified CSS selectors

tysongach:
- Doing this creates overly specific selectors for no real gain beyond
  developer understanding (which can and should be solved in other
  ways).
- The higher specificity created when using element-qualified selectors
  means that a simple class selector cannot override it later in the
  cascade.
- The problem this guideline aims to solve can be fixed simply by not
  using a vague class name.

This guideline was introduced in https://github.com/thoughtbot/guides/pull/92.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -40,15 +40,11 @@
 ## Selectors
 
 * Don't use ID's for style.
+* Avoid over-qualified selectors: `h1.page-title`, `div > .page-title`
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
 * Be consistent about naming conventions for classes. For instance, if a project is using BEM, continue using it, and if it's not, do not introduce it.
 * Use ID and class names that are as short as possible but as long as necessary.
-* Avoid using the direct descendant selector `>`.
 * Avoid nesting more than 3 selectors deep.
-* Use HTML tags on vague classes that need a qualifier like `header.application` not `.main`.
-* Avoid using the HTML tag in the class name: `section.news` not `section.news-section`.
-* Avoid using HTML tags on classes for generic markup `<div>`, `<span>`: `.widgets` not `div.widgets`.
-* Avoid using HTML tags on classes with specific class names like `.featured-articles`.
 * Avoid using comma delimited selectors.
 * Avoid nesting within a media query.
 * Don't concatenate selectors using Sass's parent selector (`&`).


tysongach:
https://github.com/thoughtbot/guides/pull/494/commits/80a07ab392b63174a445fdce90e2108a268cc16d goes a step further and removes all of our guidelines around using HTML tags (type selectors) jointly with classes in selectors. 

Instead of sometimes encouraging it and other times not, this simplifies our guideline simply to “Avoid over-qualified selectors.”

style/sass/README.md
```diff
@@ -45,10 +45,9 @@
 * Use ID and class names that are as short as possible but as long as necessary.
 * Avoid using the direct descendant selector `>`.
 * Avoid nesting more than 3 selectors deep.
-* Use HTML tags on vague classes that need a qualifier like `header.application` not `.main`.
-* Avoid using the HTML tag in the class name: `section.news` not `section.news-section`.
-* Avoid using HTML tags on classes for generic markup `<div>`, `<span>`: `.widgets` not `div.widgets`.
-* Avoid using HTML tags on classes with specific class names like `.featured-articles`.
+* Avoid over-qualified selectors:
```
kylefiedler: I would prefer this just be one line and formatted similar to some of the other languages. Maybe: 
```
* Avoid using an HTML tag qualifier on classes (`h1.page-title`)
```
Or just
```
* Avoid over-qualified selectors (`h1.page-title`)
```

tysongach: I did want to stray away from “HTML tag qualifier”, because there are other ways in which a selector can be over-qualified. The “Do” and “Don't” was added to attempt to be very clear as to _what_ made that an over-qualified selector (having the `h1` in there).

kylefiedler: I sorta figured that. I would say either be specific about what we consider over qualifying or remain vague in both the statement and example. What about wrapping this and the direct descendant up together (both are over-qualifying)? 

Remove line 46 too and add
```
* Avoid over-qualified and over-specific selectors (`h1.page-title`, `div > .page-title`)
```

tysongach: TIL: Those (`>`) are called child combinators.

style/sass/README.md
```diff
@@ -40,15 +40,11 @@
 ## Selectors
 
 * Don't use ID's for style.
+* Avoid over-qualified selectors: `h1.page-title`, `div > .page-title`
```
kylefiedler: I like this. @tysongach do you like this?

tysongach: Yep; nice and concise.

style/sass/README.md
kylefiedler: WHAT!?!?! That's a terrible name and I refuse to call them that.

wardpenney:
👍  to this.

I almost never use complex selectors for the reason in @tysongach's second-bullet description: `the higher specificity created...`. Creating higher specificity causes problems when you need state modifiers, class modifiers, theming modifiers, etc.

PR #495, opened 2017-05-26
Add guide for concatenating selectors

tysongach:
By using string concatenation, the real selectors no longer exist in the
codebase's CSS/Sass, making them hard to search for. When someone does
search for a selector which is concatenated during build time, their
search might come up empty which gives the impression that it is unused
code and it could be mistakingly removed.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -51,6 +51,7 @@
 * Avoid using HTML tags on classes with specific class names like `.featured-articles`.
 * Avoid using comma delimited selectors.
 * Avoid nesting within a media query.
+* Don't concatenate selectors using Sass's parent selector (`&`).
 
 ## Organization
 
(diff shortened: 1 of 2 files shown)


creuter:
👍

style/sass/README.md
```diff
@@ -51,6 +51,26 @@
 * Avoid using HTML tags on classes with specific class names like `.featured-articles`.
 * Avoid using comma delimited selectors.
 * Avoid nesting within a media query.
+* Don't concatenate selectors using Sass's parent selector (`&`):
+    ```scss
```
kylefiedler: Should this be in the example page instead?

tysongach: Yeah, probably. As you can see, I'm playing around a bit with our guides layout 😉

wardpenney:
Big 👍  I value developer text-search a lot more than the clever gains resulting from meta-programming the selector name.

PR #497, opened 2017-06-05
Add exclusion to Rubocop's `BlockLength` rule

purinkle:
Before, we were using Rubocop's default `Metrics/BlockLength` rule. This was causing Hound to warn about long blocks in our specs when we did not want it to. Updated the `Metrics/BlockLength` rule to exclude any spec files.

### DIFF
--- a/style/ruby/.rubocop.yml
+++ b/style/ruby/.rubocop.yml
@@ -90,6 +90,13 @@ Metrics/AbcSize:
                  branches, and conditions.
   Enabled: false
 
+Metrics/BlockLength:
+  CountComments: true  # count full line comments?
+  Max: 25
+  ExcludedMethods: []
+  Exclude:
+    - "spec/**/*"
+
 Metrics/CyclomaticComplexity:
   Description: >-
                  A complexity metric that is strongly correlated to the number


nickcharlton:
I noticed the other day that @gylaz did something similar to Hound's `.rubocop.yml` file: https://github.com/houndci/hound/pull/1373/commits/dfed6ab4cd2f56d396dd5d51ff9e8627b34d906d

purinkle:
@nickcharlton, that's interesting. It did not seem to get picked up by my repo. I am going off what I saw in the [default config]

[default config]: https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L1299-L1302

nickcharlton:
It looks like Hound has a different version now (so I'm not surprised they're drifting a bit).

I wondered if you had a preference to how the rule should be setup (it seems like @gylaz's is a little simpler?)

style/ruby/.rubocop.yml
```diff
@@ -90,6 +90,11 @@ Metrics/AbcSize:
                  branches, and conditions.
   Enabled: false
 
+Metrics/BlockLength:
+  CountComments: false  # count full line comments?
```
mike-burns: I vote: `true`.

- We tend not to use comments in our code to begin with.
- This heuristic is, roughly, "is this method easy to quickly understand?".
- Comments indicate that the method is not easy to quickly understand.
- I even believe that comments should incur a 2x penalty, though I doubt this is an option.

tabfugnic: I would also agree to the last statement

style/ruby/.rubocop.yml
```diff
@@ -90,6 +90,11 @@ Metrics/AbcSize:
...
+  Max: 25
+  ExcludedMethods: ["describe"]
```
tabfugnic: Love this!

thorncp: Also include `feature` or `scenario`?

thorncp:
> It did not seem to get picked up by my repo.

The hound repo is outside of the thoughtbot org, so it has to have its own config (and thus updating it does not affect repos in the thoughtbot org).

style/ruby/.rubocop.yml
jakecraige: Do we need to include nested ones like `context` and `it` as well?

Separately, thoughts excluding the entire `spec` dir? I sometimes have helper methods and such that can get pretty long that may not be worth extracting and having this rule triggered could be annoying

style/ruby/.rubocop.yml
purinkle: @mike-burns  / @tabfugnic, resolved in dceb15e

style/ruby/.rubocop.yml
purinkle: @tabfugnic / @thorncp / @jakecraige, decided to go with the exclusion of files in dceb15e

style/ruby/.rubocop.yml
```diff
@@ -90,6 +90,13 @@ Metrics/AbcSize:
                  branches, and conditions.
   Enabled: false
 
+Metrics/BlockLength:
+  CountComments: true  # count full line comments?
+  Max: 25
+  ExcludedMethods: []
```
gylaz: This is the default in RuboCop. Are you thinking that it might be nice to explicit here?

PR #499, opened 2017-07-07
Change SCSS concatenation to interpolation

wardpenney:
Not sure what word we want to use (semantics, amirite?). I have been referring to this action as "interpolation" because one string is sort of rendered into the other's class name. 

I have referred to concatenation before as when two classes are joined into a selector to create a modifier like:

```
.parent {
  &.--fancy { };
}
```

Resulting in `parent.parent--fancy { }`.

Thoughts?

Edit: I actually *also* don't nest these, but I think it might be good to just have a separation between the two in speaking.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -47,7 +47,7 @@
 * Avoid nesting more than 3 selectors deep.
 * Avoid using comma delimited selectors.
 * Avoid nesting within a media query.
-* Don't concatenate selectors using Sass's parent selector (`&`).
+* Don't interpolate selectors using Sass's parent selector (`&`).
 
 ## Organization
 
--- a/style/sass/sample.scss
+++ b/style/sass/sample.scss
@@ -38,7 +38,7 @@ $map: (
     }
   }
 
-  // don't concatenate selectors
+  // don't interpolate selectors
   &__child {
     color: $red;
   }


tysongach:
Interesting; we should definitely make this more clear if we can. Personally, I feel like concatenation is correct here because the two strings are being joined, even if it is creating one class, not two. I understand interpolation as more of a placeholder which gets replaced at a later point.

tysongach:
So I guess the `&` is the placeholder in the interpolation scenario?

tysongach:
While we do have the `sample.scss` here in the repo, having examples next to each guide could go a long way in providing clarity around these things. 🤔

wardpenney:
Yeah good point. Interpolation does refer to the usage of `%#{}` more specifically.

It looks like the [SASS docs](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#Referencing_Parent_Selectors_____parent-selector) seem to refer to this as a compound selector, but it's unclear (search that page for `compound`, first result). 

I'm going to close this. On a few projects, this has been referred to in different ways. I think the example speaks well enough.

wardpenney:
And yeah, to your 2nd comment, the `&` acts as interpolation, but it's the SASS engine that uses it, not the developer (like interpolating a variable). So It's better to keep "interpolation" to the thing the developer does, not the framework.

PR #500, opened 2017-09-29
Start at some Android best practices

mike-burns:
I spent a week pairing with Amanda on a small, greenfield project. Here
are some things I learned.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -264,6 +264,21 @@ Objective-C
 * Prefer categories on `Foundation` classes to helper methods.
 * Prefer string constants to literals when providing keys or key paths to methods.
 
+Android
+-------
+
+* Each callback on an `Activity` should invoke methods on an instance of a
+  presenter object.
+* Unit test each presenter object.
+* Integration tests are optional and not encouraged.
+* Always prefix `R.id` with the `R`, even if Android Studio tries to remove it.
+* Divide packages by feature, not design pattern. The same package will likely
+  have one activity and one presenter.
+* Do not use `android:onClick`.
+* As you delve deeper into the system, change your naming to focus more on the
+  data and less on the user. Use the converse strategy when programming out
+  from the depths.
+
 Shell
 -----
 


mike-burns:
Topics from my notes that didn't make the list:

- Kotlin's collection has a `lastIndex` method.
- If you need the time, check to see whether Android gives you a value you can use already. `MotionEvent.eventTime` is an example.

best-practices/README.md
```diff
@@ -264,6 +264,21 @@ Objective-C
 * Prefer categories on `Foundation` classes to helper methods.
 * Prefer string constants to literals when providing keys or key paths to methods.
 
+Android
+-------
+
+* Each callback on an `Activity` should invoke methods on an instance of a
```
mandybess: I realize this isn't how the rest of the `best-practices` is organized but I think we should separate `project and style` guidelines from`architecture` guidelines as those are very different and the `architecture` one is more likely to change. I also think that helps give readers an idea of what type of "best practice" were talking about. Because for example, "unit test each presenter object" only makes sense if the user is using the MVP architecture pattern. 

For formal guidelines I happen to really love [Ribot labs' guidelines](https://github.com/ribot/android-guidelines). I think they did an excellent job and at some points maybe even get too specific. But I kinda love that. 

Maybe we update those to our standards and add some Kotlin stuff instead?

best-practices/README.md
```diff
@@ -264,6 +264,21 @@ Objective-C
...
+  presenter object.
+* Unit test each presenter object.
+* Integration tests are optional and not encouraged.
```
mandybess: I feel weird posting this publicly. I think I, Amanda, have reasons not to use integration tests and a lot of those are specific to my experiences and the fact that we do client work. I think integration tests are hard to set up and maintain especially given our tight timelines and how quickly and frequently client's change designs. But I don't think this statement is true for ALL android projects. I think at product companies they might make a great deal of sense especially when they have the time and resources for it

best-practices/README.md
```diff
@@ -264,6 +264,21 @@ Objective-C
...
+* Always prefix `R.id` with the `R`, even if Android Studio tries to remove it.
+* Divide packages by feature, not design pattern. The same package will likely
+  have one activity and one presenter.
+* Do not use `android:onClick`.
```
mandybess: maybe specify `in XML`

mandybess: I also wonder for advice like this if it's worth including a "why"? not sure if that's out of the scope of `best practices` but my first instinct when i read something like "do not do this" is to ask "why?"

mike-burns:
Well now you're getting at everyone's pet peeve with `guides`: we never say why.

Would you rather we don't have any Android best practices documented? Is it too variable to say what a best practice is? Do we stick with what the larger community is doing? Should we link our Android best practices to Ribot Labs' instead (we do something like this for Python already)?

mike-burns:
@mandybess is anything in here salvageable, or should I close this without merging?

mandybess:
@mike-burns i'd say close without merging for now. This can be a project @alexsullivan114 and I take on more seriously in the coming year 👍

PR #501, opened 2017-10-10
Fix warnings and an error in rubocop.yml

jmhooper:
Rubocop v0.50.0 moved some cops and renamed them. This means that running rubocop without this change results in a number of warnings, then fails with an error.

To fix the warnings:

- Rename `Style/AccessorMethodName` to `Naming/AccessorMethodName`
- Rename `Style/AsciiIdentifiers` to `Naming/AsciiIdentifiers`
- Rename `Style/FileName` to `Naming/FileName`
- Rename `Style/PredicateName` to `Naming/PredicateName`

To fix the error:

- Rename `Style/OpMethods` to `Naming/BinaryOperatorParameter`

This does not look like it is backwards compatible since renaming `Style/OpMethods` to `Naming/BinaryOperatorParameter` means old versions do not know about the config for that cop. But it does prevent rubocop from bumping against an error.

Also, note that the error has been fixed on rubocop's master branch. As of today that fix has not been pushed to rubygems. Once it has been added to a release, it looks like running rubocop without this change will result in a bunch of warnings, but not an error.

The fix to rubocop: https://github.com/bbatsov/rubocop/commit/e8a61bc118055be131f4d3ae88c6e7e293072818

### DIFF
--- a/style/ruby/.rubocop.yml
+++ b/style/ruby/.rubocop.yml
@@ -266,7 +266,7 @@ Style/OneLineConditional:
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#ternary-operator'
   Enabled: false
 
-Naming/BinaryOperatorParameter:
+Naming/BinaryOperatorParameterName:
   Description: 'When defining binary operators, name the argument other.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#other-arg'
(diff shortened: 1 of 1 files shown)


gylaz:
Thanks @jmhooper!

jmhooper:
No problem, thanks to you for making this available 😊

PR #509, opened 2018-02-06
Suggest using Fast JSON API for Ruby serialization

jaysonvirissimo:
https://medium.com/netflix-techblog/fast-json-api-serialization-with-ruby-on-rails-7c06578ad17f

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -376,12 +376,14 @@ Ruby JSON APIs
 
 * Review the recommended practices outlined in Heroku's [HTTP API Design Guide]
   before designing a new API.
-* Use a fast JSON parser, e.g. [`oj`][oj]
+* Use a fast JSON parser, e.g. [`oj`][oj].
+* Use [`fast_jsonapi`][fast_jsonapi] for serializing Ruby objects.
 * Write integration tests for your API endpoints. When the primary consumer of
   the API is a JavaScript client maintained within the same code base as the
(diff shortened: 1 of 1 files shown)


bf4:
@jaysonvirissimo That gem only serializes to the JSON:API format. Is that intended?

mike-burns:
So far as I know, we have not used this library.

It seems strange to recommend libraries in these documents; it might make more sense to remove the oj line instead. I've also never used the oj library ...

bf4:
@mike-burns Oj can take over `to_json` from JSON or Rails with compatible decoding and encoding at faster speeds, so it is often recommended to just drop in.

mike-burns:
Gem recommendations seem like something that should go in [Suspenders](https://github.com/thoughtbot/suspenders), not the guides.

Anyone object to someone (potentially me, but I won't have time for 3+ weeks):

- Adding `oj` to the suspenders `Gemfile.erb`, as part of a generator.
- Removing the line about Oj from our guides.

?

georgebrock:
@mike-burns That sounds like a great idea to me, and something we could mention in this repo's `CONTRIBUTING.md` too.

tysongach:
@mike-burns 👋 Would you be up for making the changes you suggest above, so we can close this up?

mike-burns:
I've opened https://github.com/thoughtbot/suspenders/pull/923 .

mike-burns:
I've opened https://github.com/thoughtbot/guides/pull/520 .

PR #511, opened 2018-03-13
Use "dependencies" correctly

VikiAnn:
Looks like the word "dependencies" here was being used to mean the opposite of what it usually means, and "collaborators" was being used instead of "dependencies." I propose "dependencies" instead of "collaborators" and "dependants" instead of the confusing use of "dependencies."

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -19,8 +19,8 @@ Object-Oriented Design
 
 * Avoid global variables.
 * Avoid long parameter lists.
-* Limit collaborators of an object (entities an object depends on).
-* Limit an object's dependencies (entities that depend on an object).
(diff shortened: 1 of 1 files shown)


jerrett:
:+1:

best-practices/README.md
```diff
@@ -19,8 +19,8 @@ Object-Oriented Design
 
 * Avoid global variables.
 * Avoid long parameter lists.
-* Limit collaborators of an object (entities an object depends on).
-* Limit an object's dependencies (entities that depend on an object).
+* Limit dependencies of an object (entities an object depends on).
+* Limit an object's dependants (entities that depend on an object).
```
iwz: ~~dependants~~ dependents

PR #512, opened 2018-04-06
Update Rails 'Best Practices' section

abstractart:
In Rails 5.1:
``` 
DEPRECATION WARNING: Passing a class to the `class_name` is deprecated and will raise an ArgumentError in Rails 5.2. It eagerloads more classes than necessary and potentially creates circular dependencies.
```


derekprior:
Thanks for bringing this to our attention. I had forgotten this was in here. What do you think about removing this "best practice" all together as it now seems adequately covered by the deprecation (and eventual error). I think in the face of that, these guides don't need to say anything on the matter?

abstractart:
I agree, it's best to just delete this "best practice".

PR #513, opened 2018-05-17
Add rules for ordering dependencies in Elixir

drapergeek:
This allows you to view the most "coupled" dependencies at the top as `use` and `import` should be used sparingly.


joshuaclayton:
@drapergeek might be unrelated to this PR, but is there a corresponding `.formatter.exs` we can provide for use with `mix format` that would help with this?

drapergeek:
@joshuaclayton that's a great call. Unfortunately there are practically no options for the `.formatter.exs` currently. You can see what files get covered and the only actual configuration option is `locals_without_parens`

PR #515, opened 2018-07-13
Update CI guidance, remove recommendation to use specific service

tysongach:
We tend to use CircleCI over Travis CI, as is [documented in our Handbook](https://github.com/thoughtbot/handbook/blob/master/developing/continuous-integration.md).

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -51,12 +51,11 @@ Ruby Gems
 * Use [Appraisal] to test the gem against multiple versions of gem dependencies
   (such as Rails in a Rails engine).
 * Use [Bundler] to manage the gem's dependencies.
-* Use [Travis CI] for Continuous Integration, indicators showing whether GitHub
-  pull requests can be merged, and to test against multiple Ruby versions.
+* Use continuous integration (CI) to show build status within the code review
+  process and to test against multiple Ruby versions.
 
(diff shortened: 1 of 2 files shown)


tysongach:
@nickcharlton Yeah, that’s a good point and I had a bit of hesitation about that, as well. I did remember that the readme describes the “Use” language as a “positive instruction”, and that sat nicely with me.

paulcsmith:
I may be an outlier but I've had worse and worse experiences setting up Circle lately and have had better and better experiences with Travis.

The new Circle 2.0 config and UI has been pretty painful to implement. I personally would prefer to keep it as Travis or maybe suggest both depending on team preference

mike-burns:
Historical: we once talked at extreme length about these two, and I don't understand where it ended: https://trello.com/c/GWgWRiBU/280-replace-travis-with-circle

paulcsmith:
@tysongach I think that makes sense. Maybe adding a suggestion would be nice, but I think the team would be the best at making a final decision. So maybe something like `Use continuous integration (CI). Some CI services we've used and recommend are Circle CI and Travis`?

jferris:
I agree that recommending continuous integration (and maybe integration with the code review process) is more important than the specific service.

tysongach:
I’m planning to merge this by the end of today unless people have objections.

PR #517, opened 2018-07-15
Remove mention of concatenated selectors

tysongach:
We recently added a guideline to not concatenate selectors, so this is no longer relevant. See https://github.com/thoughtbot/guides/commit/423f4e1516379ddd091eae49dc5dfd46eef525f4

Also, it’s not ideal that when we edit this ordering, you have the update the whole list because we’ve spelled out each placement. I intend to tackle this in a separate PR as part of a larger improvement.

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -32,10 +32,9 @@
 * Use alphabetical order for declarations.
 * Place `@extends` and `@includes` at the top of your declaration list.
 * Place media queries directly after the declaration list.
-* Place concatenated selectors second.
-* Place pseudo-classes and pseudo-elements third.
-* Place nested elements fourth.
-* Place nested classes fifth.
+* Place pseudo-classes and pseudo-elements second.
+* Place nested elements third.
+* Place nested classes fourth.
 
 ## Selectors
 


style/sass/README.md
```diff
@@ -32,10 +32,9 @@
 * Use alphabetical order for declarations.
 * Place `@extends` and `@includes` at the top of your declaration list.
 * Place media queries directly after the declaration list.
-* Place concatenated selectors second.
-* Place pseudo-states and pseudo-elements third.
-* Place nested elements fourth.
-* Place nested classes fifth.
+* Place pseudo-states and pseudo-elements second.
```
almnt: I know this PR focuses on declaration ordering, but I noticed that in [this PR](https://github.com/thoughtbot/guides/pull/519/files), you corrected `pseudo-states` to `pseudo-classes`. Is that a change that should happen for this section, as well?

tysongach: The two PRs are editing the same content, so once we merge we merge into `master` it will all be up-to-date with that change to `pseudo-classes`.

almnt:
Oh okay, I missed that. Gotcha!

PR #518, opened 2018-07-15
Remove guidance to use specific libraries/tools

tysongach:
- Authors should pick and use tools based on the needs of the project
- We are amongst an explosion in front-end tooling; suggesting this
  small list of libraries/tools/framework is shortsighted
- "Use Bourbon for a Sass library" is confusing because "Sass library"
  is not a defined "thing" that serves every project equally
- Now that CSS Grid Layout has broad browser support, Neat or any grid
  framework might not be needed
- It was strange to categorize these guidelines under "Organization"
- I have a hunch that this guideline is rooted in a time when Bourbon
  was used primarily as a prefixing library, before Autoprefixer was a
  thing and when prefixes were a core need in any project; this is no
  longer the case today

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -50,9 +50,6 @@
 
 ## Organization
 
-* Use Bourbon for a Sass library.
-* Use Neat for a grid framework.
-* Use Bitters/`base` directory for styling element selectors, global variables, global extends and global mixins.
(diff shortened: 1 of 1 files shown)


tysongach:
@creuter Good question. Does Suspenders and our preference to start Rails projects using it serve as that? That is Rails-specific, though…

creuter:
I don't think I've been on a Rails project in a couple years, but think that's a fine place to point people to. 👍

PR #524, opened 2018-10-02
Edited security readme

Frenchtoast0:
Security practice ideas.


mike-burns:
Thank you for this!

We use the guides as a place to document things that thoughtbot has tried and found success with. Since we haven't tried those competitions, it doesn't make sense to add them to our guides -- but we'll be sure to consider them to improve our skills.

PR #525, opened 2018-10-10
Clarify meaning of approval

JoelQ:
Based on a conversation that happened in Slack on what "approval" means
in the context of code review. This makes it explicit that the PR author
gets to make the final call on whether to merge or make changes. The
reviewer is there to give feedback, not to act as a gatekeeper.

Ideally, discussion and edits go on until both parties are satisfied. In
the event that this drags on too long or no agreement can be reached,
the author may act unilaterally to merge without making some of the
suggested changes.

### DIFF
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -47,7 +47,7 @@ Having Your Code Reviewed
 * Wait to merge the branch until continuous integration (TDDium, Travis CI,
   CircleCI, etc.) tells you the test suite is green in the branch.
 * Merge once you feel confident in the code and its impact on the project.
-* Final editorial control rests with the PR author.
+* Final editorial control rests with the pull request author.
 
 [how hard it is to convey emotion online]: https://www.fastcodesign.com/3036748/why-its-so-hard-to-detect-emotion-in-emails-and-texts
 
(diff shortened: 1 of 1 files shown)


JoelQ:
Personally, I never use the "Request changes" option because it feels so negative (big red x and all that). I prefer to use either "Comment" (when I make a lot of suggestions or when I want to imply that I'd like to do a second round of review once changes are addressed) or "Approve" (when I don't expect to return to this PR).

Not sure how others use the system though 😄

code-review/README.md
```diff
@@ -47,6 +47,7 @@ Having Your Code Reviewed
 * Wait to merge the branch until continuous integration (TDDium, Travis CI,
   CircleCI, etc.) tells you the test suite is green in the branch.
 * Merge once you feel confident in the code and its impact on the project.
+* Final editorial control rests with the PR author.
```
jferris: Worth expanding out "PR" to "pull request" so it's easy to understand without more context?

(I was very tempted to use "Request Changes" for this comment)

JoelQ: good point

JoelQ: Fixed in d3542b5

mike-burns:
LGTM, that's also how I use those buttons, and for history tracking: the prior discussion on this was in #279.

JoelQ:
Good find @mike-burns ! 🕵️

PR #526, opened 2018-10-12
Add guide about pair programming

benfurber:
Pairing is an essential skill for developers. This guide summarises
some of the most important things to consider before/during/after
every pairing session.

### DIFF
--- a/working-together/README.md
+++ b/working-together/README.md
@@ -9,7 +9,7 @@ A guide for working well together.
 - Agree Scope (Pairing for two hours? Until the ticket is complete? Just get past the bug?)
 - Agree Physical/Virtual Location ("Will we both be comfortable here?")
 - Agree Working Environment (Two keyboards? Text editor and other tools)
-- Agree Pairing Style (Time-based? Ping pong? To the next traffic light?)
+- Agree Pairing Style (Time-based? Ping pong?)
 
 **During:**
 
--- a/working-together/README.md
+++ b/working-together/README.md
@@ -1,0 +1,23 @@
+Working Together
+==============
+
+A guide for working well together.
+
+
+Pair Programming
+-------
+
+**Before:**
+* Agree Scope (Pairing for two hours? Until the ticket is complete? Just get pass the bug?)
+* Agree Working Environment (Two keyboards? VIM/what other tools?)
+* Agree Pairing Style (Time-based? Ping pong? To the next traffic light?)
+
+**During:**
+* Keep the chat going (Think aloud. Encourage/support)
+* Keep switching (Follow the pairing style)
+* Keep both involved ("Could we do this another way?")
+* Keep breaking (Have a break, have a…)
+* Keep checking in ("Could we google this separately?")
+
+**After:**
+* Ask for Feedback (What did we do well? What could we have done better? It’ll feel weird, do it anyway)
--- a/working-together/README.md
+++ b/working-together/README.md
@@ -16,7 +16,7 @@ A guide for working well together.
 - Keep the chat going (Think aloud. Encourage/support)
 - Keep switching (Follow the pairing style)
 - Keep both involved ("Could we do this another way?")
-- Keep breaking (Have a break, have a…)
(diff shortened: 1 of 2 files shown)


working-together/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+A guide for working well together.
+
+
+Pair Programming
+-------
+
+**Before:**
+* Agree Scope (Pairing for two hours? Until the ticket is complete? Just get pass the bug?)
+* Agree Working Environment (Two keyboards? VIM/what other tools?)
```
germsvel: Even though I love `vim`, I don't think everyone uses it in the office. What do you think about changing this to "Text editor and other tools?"

working-together/README.md
```diff
@@ -0,0 +1,23 @@
...
+* Agree Pairing Style (Time-based? Ping pong? To the next traffic light?)
+
+**During:**
+* Keep the chat going (Think aloud. Encourage/support)
+* Keep switching (Follow the pairing style)
+* Keep both involved ("Could we do this another way?")
+* Keep breaking (Have a break, have a…)
+* Keep checking in ("Could we google this separately?")
```
mike-burns: Unlike thoughtbot, Google capitalizes the first letter of their name.

But since a loud group of us are on DDG, how about something like "Could we search for a guide separately?"?

working-together/README.md
```diff
@@ -0,0 +1,23 @@
...
+==============
+
+A guide for working well together.
+
+
+Pair Programming
+-------
+
+**Before:**
+* Agree Scope (Pairing for two hours? Until the ticket is complete? Just get pass the bug?)
```
SViccari: Grammar suggestion: `Just get past the bug?`

working-together/README.md
```diff
@@ -0,0 +1,23 @@
...
+* Agree Working Environment (Two keyboards? VIM/what other tools?)
+* Agree Pairing Style (Time-based? Ping pong? To the next traffic light?)
```
SViccari: Just curious, is the "next traffic light" similar to the "ping pong" pattern?

working-together/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+**During:**
+* Keep the chat going (Think aloud. Encourage/support)
+* Keep switching (Follow the pairing style)
+* Keep both involved ("Could we do this another way?")
+* Keep breaking (Have a break, have a…)
```
SViccari: Should there be additional text after "have a...."? Possible suggestions: "Walk to the kitchen together for a snack/tea/coffee. Take a moment to stand and stretch. Go for a short walk."

working-together/README.md
```diff
@@ -0,0 +1,23 @@
+Working Together
+==============
+
+A guide for working well together.
+
+
+Pair Programming
+-------
+
+**Before:**
```
SViccari: I've run into pairing environments where it's tough to pair due to limited desk space. It feels awkward to bring it up and ask the person to move their stuff off the desk. As part of the **before**,  can we include:
- Agree on a location (Your desk, mine, or another location? Create space to accommodate two people and two keyboards.)

working-together/README.md
benfurber: As in the TDD "red, green, refactor" traffic lights... Sounds like I might just have made this up...

working-together/README.md
benfurber: Really good point!

working-together/README.md
SViccari: haha, I *thought* this might be KitKat related but I've never seen that particular commercial.

working-together/README.md
SViccari: Perhaps it's just a pairing pattern that I'm not familiar with. Or perhaps you have invented a new style :)  I like the name. It seems very similar to ping-pong pairing but this approach changes drivers when refactoring? I believe the refactoring stage is not explicitly called out in the ping-pong pattern.

How I'm imagining "to the next traffic light" pairing:
A writes a new test and sees that it fails (red)
B implements the code needed to pass the test (green)
A refactors (refactor)
B writes the next test and sees that it fails (red light)
... etc ...

PR #527, opened 2018-10-23
Add guidance to format JavaScript via Prettier

joshuaclayton:
Prettier is the standard for formatting many types of files, including
JavaScript. By standardizing through a community-recommended formatting
tool, we're able to reduce pull request chatter by simplifying
formatting decisions and ensure consistency across the codebase.

As such, we also recommend Prettier be incorporated into any build
processes on CI to ensure no code is committed that hasn't been
formatted through Prettier.

This provides no guidance on how to incorporate this into the build
process; however, one common approach is to add a script to the
repository's `package.json`:

```json
{
  "scripts": {
    "prettier-lint": "prettier -l \"**/*.js\""
  }
}
```

PR #528, opened 2018-10-25
Make JS style config consistent with Prettier

danbee:
We currently recommend double quotes and no spaces within arrays for Ruby. This updates our JS style to be consistent with that and adds an example to the `sample.js` config.

I also added a missing semicolon.

Edit: This also makes our eslint config consistent with Prettier.

### DIFF
--- a/style/javascript/.eslintrc.json
+++ b/style/javascript/.eslintrc.json
@@ -104,7 +104,7 @@
     ],
     "array-bracket-spacing": [
       2,
-      "always"
+      "never"
     ],
     "indent": [
       2,
@@ -115,7 +115,7 @@
     ],
     "quotes": [
       2,
-      "single",
+      "double",
       {
         "avoidEscape": true
       }
--- a/style/javascript/README.md
+++ b/style/javascript/README.md
@@ -9,7 +9,7 @@ JavaScript
 * Prefer [arrow functions] `=>`, over the `function` keyword except when
   defining classes or methods.
 * Use semicolons at the end of each statement.
-* Prefer single quotes.
+* Prefer double quotes.
 * Use `PascalCase` for classes, `lowerCamelCase` for variables and functions,
   `SCREAMING_SNAKE_CASE` for constants, `_singleLeadingUnderscore` for private
   variables and functions.
--- a/style/javascript/sample.js
+++ b/style/javascript/sample.js
@@ -1,4 +1,6 @@
-object = { spacing: true }
+object = { spacing: true };
+
+array = [1, 2, 3];
 
 class Cat {
   canBark() {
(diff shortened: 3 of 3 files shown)


style/javascript/.eslintrc.json
```diff
@@ -104,7 +104,7 @@
     ],
     "array-bracket-spacing": [
       2,
-      "always"
+      "never"
```
mike-burns: TBH, I want no rule on this, both in JS and in Ruby. We're smarter than a linter at knowing how to make arrays readable.

How do you feel about this?

```suggestion
      "never", { "objectsInArrays": true, "arraysInArrays": true }
```

For those reading along at home, here are the docs: https://eslint.org/docs/rules/array-bracket-spacing

thorncp:
#527 is recommending that we use Prettier. Should we continue maintaining an ESLint config and our own style preferences if we intend to use Prettier?

danbee:
@thorncp I don't think Hound supports Prettier.

style/javascript/.eslintrc.json
danbee: I definitely prefer consistent no spaces inside arrays regardless of the content. No spaces is also consistent with Prettier which seems to be what we're recommending too.

style/javascript/.eslintrc.json
```diff
@@ -115,7 +115,7 @@
     ],
     "quotes": [
       2,
-      "single",
+      "double",
```
stevehanson: It seems like the JS community mostly uses single quotes. I’d personally prefer that we try to conform to community styles as opposed to trying to be consistent across stacks (i.e. use double quotes for all languages).

stevehanson: [This article](https://bytearcher.com/articles/single-or-double-quotes-strings-javascript/) has some statistics on quote usage. It seems like single quotes are more popular, but some prominent style guides (JSLint, ESLint, Typescript) prefer double quotes. I've changed my mind to being OK with this change.

style/javascript/.eslintrc.json
mike-burns: How would you feel about

```suggestion
      "off"
```

instead? Let the dev figure out what's best.

joshuaclayton: Having worked with Prettier, elm-format, brittany, and rufo over the past few months, I'm in favor of a formatter handle this and make the decision for me. I don't have a preference of whether this is set to `"always"` or `"never"`, although if I had to pick, I'd lean towards `"never"`

tysongach:
Any update here?

cpytel:
Closing due to inactivity.

PR #532, opened 2019-03-05
Switch Hound to use stylelint instead of scss-lint

tysongach:
Based on a [successful Research experiment][research-card], this changes
the Hound configuration in this repo to use stylelint, and disables
scss-lint.

The configuration for stylelint was added in a
[previous commit][stylelint-commit] and matches the scss-lint as much as
possible, so this change should see only minor differences in what
Hound catches.

The reason that it's useful and important to change the Hound config in
this repo, is because we use an organization-wide config in our Hound
account. That means Hound looks to this repo's `.hound.yml` if a project
does not have a local `.hound.yml`. So once this is merged to `master`,
Hound will begin using stylelint in all future thoughtbot PRs, unless
a project declares otherwise.

[research-card]: https://trello.com/c/n17oCxXV/712-replace-scss-lint-with-stylelint
[stylelint-commit]: https://github.com/thoughtbot/guides/commit/ac79b2614a54bae36305a0be77a65c06a7aa910d


ericwbailey:
I'm in favor of this approach. 

I have used both scss-lint and stylelint for personal projects, and found that stylelint is far more flexible, extensible, and more actively worked on. I also think it's worth mentioning that stylelint is language agnostic, which will both future-proof our efforts and ensure that we have good standards for client projects that don't use Sass (vanilla CSS, Less, etc.).

This is awesome stuff, @tysongach!

enatario:
Is this ready to merge?

tysongach:
It is, yes. I was waiting for https://github.com/thoughtbot/suspenders/pull/974 to be merged around the same time as this so that any new Rails project would be in line with these expectations. But that is not a hard requirement. I believe @composerinteralia was planning to work on that Suspenders PR more today, so perhaps I see where that lands.

PR #533, opened 2019-03-08
The thoughtbot Guide to Application Security

mike-burns:
This guide is meant to help us all share knowledge and understanding
about what it means to develop an application more securely. It ensures
that we are all on the same page about these important security topics.

The goal here is not to be complete; all of these topics are discussed
in much more depth elsewhere. Rather, the goal is to make sure we have
an overview of what to look for when making software architecture
decisions.

### DIFF
--- a/security/application.md
+++ b/security/application.md
@@ -1,0 +1,571 @@
+# The thoughtbot Guide to Application Security
+
+## Threat modeling
+
+The task of identifying concrete attacks and understanding their relationship
+with the code is the core task of threat modeling. We can understand this from
+two perspectives:
+
+- identify what can go wrong, and
+- don't account for things that cannot go wrong.
+
+Identifying what can go wrong is what is most often [written about when
+discussing threat](https://www.owasp.org/index.php/Application_Threat_Modeling)
+modeling. There are [many threat modeling
+techniques](https://insights.sei.cmu.edu/sei_blog/2018/12/threat-modeling-12-available-methods.html),
+but the summary is:
+
+1. Create a list of what an attacker can do on your app. For a Web app, they
+might be able to spoof HTTP headers, submit malicious data, or embed a Web page
+in an `iframe`.
+2. Add to the list the weak points of the app. These will likely be places
+where you are doing something non-standard, which the frameworks don't know how
+to protect.
+3. Prioritize this list. Take into account factors such as difficulty of
+attack, likelihood of attack, ease of mitigating the attack, and severity of
+attack.
+
+Anything not in the list are things you cannot use as a reason to do something.
+Since the list is prioritized, you can use it to help it to prioritize tickets
+or split tickets.
+
+## Library updates
+
+The easiest line of defense you have as a developer is [applying security fixes
+for our
+dependencies](https://snyk.io/blog/top-ten-most-popular-docker-images-each-contain-at-least-30-vulnerabilities/)
+as they are released.
+
+On the flip perspective, when releasing a security fix for one of our projects,
+make it trivial to upgrade: don't include new features or unrelated bug fixes.
+
+There are a few ways to keep up with security fixes:
+
+- Any platform-specific tool, such as
+  [bundler-audit](https://github.com/rubysec/bundler-audit#readme).
+- [Any official CVE feed](https://cve.mitre.org/cve/data_updates.html).
+
+If you have access, the thoughtbot
+[#security](https://thoughtbot.slack.com/messages/security) channel does our
+best to keep up with security issues that we think will affect us or our
+clients.
+
+## Secure programming
+
+In an ideal world, access to the program's source code will not give an
+attacker an advantage. This is not always possible, but programming with
+a mindset of preventing an all-knowing attacker can be healthy.
+
+[Some tips](https://twitter.com/SarahJamieLewis/status/1097300029016989696)
+along the way:
+
+- Always check return values. If the procedure can raise, make sure to handle
+  that (to prevent DoS attacks). If the procedure can signal failure, make sure
+  to handle that (to prevent read-after-free-style attacks).
+- Fail fast. If the data seems odd, don't recover: fail.
+- Leave the most security-sensitive code as an [omega mess], once it works. Too
+  many bugs -- more than zero -- come out of refactoring to be worth a change
+  in the name of code beauty.
+
+[omega mess]: https://speakerdeck.com/skmetz/go-ahead-make-a-mess
+
+## User data
+
+Any data from the user is malicious until proven innocent. Examples of user
+input are data from forms, HTTP headers, text the user enters into your mobile
+app, IP address, MAC address, email headers, file paths, GraphQL queries,
+uploaded files, and stdin. And more.
+
+When possible, rely on a framework to parse user data. Don't parse HTTP headers
+by hand, use the Rails validations, pass JSON data through a schema validator,
+send addresses straight to the shipping or map API, etc.
+
+If you can't rely on a library, handle user data in two stages: verify, then
+work with it. For example, if someone uploads a file with a filename ending in
+`.jpg`, use libmagic to confirm that it is a JPEG, and then consider it less
+tainted and ready for use.
+
+### SQL injection
+
+We know about this one, so let's make sure it does not happen.
+
+Whenever you run a SQL query, don't insert user input into it. If you must
+insert user data into it, use a [bind
+variable](https://www.ibm.com/developerworks/library/se-bindvariables/index.html).
+(The details of how bind variables work depends on your object-relational
+mapping library.)
+
+### YAML
+
+[YAML is too vulnerable to
+attacks](https://trailofbits.github.io/rubysec/yaml/index.html) to consider for
+new projects.
+
+### Client-side validation
+
+All client-side validation, such as a React component that tells the user that
+their email address is not in a valid format, is for presentation. These
+checks, and more, must be duplicated on the backend. Any attacker can use curl
+to bypass your client-side validations.
+
+### Cookies
+
+Cookies are user-controlled input and, therefore, should be treated with
+suspicion. If possible, don't rely on a cookie.
+
+Cookies can be copied between browsers. Just because a request sends a cookie
+does not mean that the cookie was sent by the user's original browser. It might
+come from curl.
+
+One way to retain control over the cookie data is to sign it using a secret key
+only known by the server. Rails does this for you.
+
+## Logging
+
+Logging is a compromise between having enough data to be able to debug
+a problem and having too much personally-identifying information about a user.
+
+Make sure not to log passwords, credit card numbers, or any other information
+that you do not strictly need. Err on the side of not logging any strings, if
+possible.
+
+In Rails, use the `filter_paramters` configuration setting to remove known
+attributes from the logs.
+
+In addition, if you are logging to a service over a network connection, make
+sure the connection itself is secured using TLS.
+
+## Personally-Identifying Information (PII)
+
+As much as you can, do not touch any information you don't need. Some tricks
+for this:
+
+- Send any credit card data directly to the payment processor from the client.
+  They'll give back a token, which you can store safely.
+- You probably don't need the user's sex, gender, date of birth, middle name,
+  and so on. You might, but ask yourself first: do you?
+
+## Randomization
+
+Most modern cryptography is dependent on really big prime numbers and access to
+solid randomization. If you find yourself in a place where you need a random
+number, here are some things to keep in mind.
+
+- Don't do this yourself. If you can use Ruby's `SecureRandom` or functions
+  like `arc4random_buf(3)` and `arc4random_uniform(3)`, do that instead.
+- [Do not restrict the randomized
+  space](http://www.pcg-random.org/posts/bounded-rands.html) with a modulo or
+  floating-point multiplication bias. Instead, try generating a random number
+  in a loop, returning when the value is within the desired range.
+- Use an unpredictable seed. Do not use the current time, or the seconds since
+  boot, or `0`, or your age, or the result from calling rand seeded on
+  a predictable seed. If possible, use a random number generator that you do
+  not seed yourself, such as `arc4random(4)` or `/dev/random`.
+- Use a non-blocking random number generator. If an attacker discovers that the
+  random number generator blocks, such as Linux's `/dev/urandom`, that is
+  a potential denial of service attack vector.
+
+## Hashing
+
+A hashing function provides a one-way encoding of an object. Use this any time
+you don't actually care what the value _is_, but instead you care that you have
+it at all. The only operation you'll want to perform against a hashed object is
+an equality check.
+
+(As a side note, when people refer to dictionary data structures as "hashes",
+they're referring to the fact that a hashing function is used to turn the key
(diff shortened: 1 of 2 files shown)


stratigos:
Thanks, I think its helpful to point out that `ActiveSupport` is mitigating timing attacks via [.secure_compare](https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html#method-c-secure_compare).

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+
+On the flip perspective, when releasing a security fix for one of our projects,
+make it trivial to upgrade: don't include new features or unrelated bug fixes,
+if possible.
+
+There are a few ways to keep up with security fixes:
+
+- Any playform-specific tool, such as
+  [bundler-audit](https://github.com/rubysec/bundler-audit#readme).
+- The [#security](https://thoughtbot.slack.com/messages/security) channel.
```
jferris: The guides are public, but this channel is not. Maybe call out that it's only for thoughtbot team members?

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+- [Any official CVE feed](https://cve.mitre.org/cve/data_updates.html).
+
+## User data
+
+Any data from the user is malicious until proven innocent. Examples of user
+input are data from forms, HTTP headers, keyboard-entered input into an Android
```
jferris: > keyboard-entered input into an Android app

Isn't ANY data from a mobile app or other remote client considered user input?

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+We know about this one, so let's make sure it does not happen.
+
+Whenever you run a SQL query, don't insert user input into it. If you must
+insert user data into it, use a [bind
+variable](https://www.ibm.com/developerworks/library/se-bindvariables/index.html).
+(The details of how bind variables work depends on your ORM.)
+
+### YAML
+
+[YAML is too vulnerable to
```
jferris: :fire:

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+- logging
+   - TLS if networked
+   - be careful what you log
+- PII
+   - send CC data directly to processor
+   - don't touch it if you can avoid it
+   - don't store any more than you need
+- check your return values
+- HMAC - probably goes under hashing somewhere. mention encrypt-then-hmac in
+  the hashing vs encryption section
```
jferris:
```suggestion
  the hashing vs encryption section
- Multi-factor authentication
```

jferris:
Wow! This is a lot of excellent work. I went through some sections and added a few comments, and I'll try to dive deeper later.

security/application.md
mike-burns: That's true from the perspective of an API developer, but I meant this from the perspective an an Android developer -- don't trust the things the user types into your app. Make sure to validate them first.

How can I better phrase this?

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+The most popular symmetric algorithms you'll encounter are AES and Twofish.
+These might be useful for encrypting a file to share with a group of people or
+for encrypting your filesystem. Rumor has it that 1Password uses something like
+it AES to encrypt an entire vault; it is encrypted at rest, and only decrypted
+when you enter the passphrase.
+
+Asymmetric encryption algorithms, also known as public/private keypair
+encryption, is more well-known -- in large part for how tricky they are to get
+right. Some famous ones are SSH, TLS (previously SSL), and PGP.  These start by
+generated a pair of encryption secrets known as the public and private keys.
```
mike-burns:
```suggestion
generating a pair of encryption secrets known as the public and private keys.
```

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+public key, plus a signature from another key. If the signature is by one of
+the trusted public keys, the browser accepts the Web site's key; otherwise,
+it's a failure.
+
+For example, Firefox trusts GlobalSign. thoughtbot.com has a TLS certificate
+that was signed by GlobalSign. When you visit thoughtbot.com, it sends its
+public key plus the signature from GlobalSign. Firefox trusts GlobalSign, so it
+trusts thoughtbot.com's key.
+
+This mechanism is called a central authority.
```
mike-burns:
```suggestion
This mechanism is called a certificate authority.
```

(and elsewhere)

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+hashing algorithms (they're actually quite different, but that difference is
+negligible), except they are intentionally slow.
+
+The three most common password hashing algorithms are bcrypt, scrypt, and
+PBKDF2. Each of these require a salt, but handle it themselves: the output of
+these functions is a string that contains the salt plus the hashed value. Store
+that entire string as the hashed password.
+
+Never compare hashed passwords using the built-in string equality function;
+this sets you up for a timing attack.
```
georgebrock: Maybe move this under the `Timing attacks` header?

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+There are two kinds of encryption algorithms: symmetric and asymmetric.
+A symmetric algorithm is one where the same secret is used to encrypt it and
+decrypt it. An asymmetric algorithm is one where the string can be encrypted
+and decrypted using different secrets -- where the person encrypting cannot
+necessarily decrypt it.
+
+The most popular symmetric algorithms you'll encounter are AES and Twofish.
+These might be useful for encrypting a file to share with a group of people or
+for encrypting your filesystem. Rumor has it that 1Password uses something like
+it AES to encrypt an entire vault; it is encrypted at rest, and only decrypted
```
georgebrock:
```suggestion
AES to encrypt an entire vault; it is encrypted at rest, and only decrypted
```

georgebrock: The rumours check out: https://github.com/georgebrock/1pass/blob/master/onepassword/encryption_key.py#L4

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+when you enter the passphrase.
+
+Asymmetric encryption algorithms, also known as public/private keypair
+encryption, is more well-known -- in large part for how tricky they are to get
```
georgebrock:
```suggestion
encryption, are more well-known -- in large part for how tricky they are to get
```

(plural to agree with _algorithms_ on the previous line)

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+right. Some famous ones are SSH, TLS (previously SSL), and PGP.  These start by
+generated a pair of encryption secrets known as the public and private keys.
+Anyone with the public key can encrypt a string, but only the holder of the
+private key can decrypt.
+
+(The math here is cool. I won't go into it.)
```
georgebrock: If we're going to tease people with it, maybe a link to where they can read more?

security/application.md
```diff
@@ -0,0 +1,412 @@
...
+later will confirm that the key they have for the person matches the
+fingerprint on the paper.
+
+This mechanism is called a web of trust.
+
+### TLS
+
+Transport Layer Security is a way for a browser to trust a server. The browser
+ships with a list of trusted public keys. Each Web site serves up its own
+public key, plus a signature from another key. If the signature is by one of
```
georgebrock: Maybe move the signing section up before the sections on key exchange, so that we've introduced signing when we get to here.

security/application.md
```diff
@@ -82,25 +106,62 @@ their email address is not in a valid format, is for presentation. These
...
+sure the connection itself is secured using TLS.
+
+## Personally-Identifying Information (PII)
+
+As much as you can, do not touch any information you don't need. Some tricks
+for this:
+
+- Send any credit card data directly to the payment processor from the client.
+  They'll give back a token, which we can store safely.
+- You probably don't need the user's sex, gender, date of birth, middle name,
```
lunohodov: Aren't sex and gender synonyms? Maybe one can be removed...

mike-burns: Dictionary.com has gender as pertaining to "cultural roles" and sex pertaining to "reproductive functions". I think it's nice to keep both here as a reminder that we probably don't need to know either, for most products.

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+These might be useful for encrypting a file to share with a group of people or
+for encrypting your filesystem. 1Password uses AES to encrypt an entire vault;
+it is encrypted at rest, and only decrypted when you enter the passphrase.
+
+Asymmetric encryption algorithms, also known as public/private keypair
+encryption, are more well-known -- in large part for how tricky they are to get
+right. Some famous ones are SSH, TLS (previously SSL), and PGP.  These start by
+generating a pair of encryption secrets known as the public and private keys.
+Anyone with the public key can encrypt a string, but only the holder of the
+private key can decrypt.
```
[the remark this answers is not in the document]
louis-antonopoulos: Not sure if you want to get into this here, but it might be worth discussing how asymmetric encryption greatly increases the size of the thing you're encrypting while symmetric does not -- at least in the context of why databases don't use asymmetric encryption for encrypted fields.

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+`Strict-Transport-Security` header. The value for this header is
+`max-age=31536000; includeSubDomains; preload`.
+
+- `max-age` determines how long this domain should remain in the browser's
+  list. `31536000` is one year. Feel free to go longer.
+- `includeSubDomains` specifies that subdomains should also be considered part
+  of the HSTS list.
+- `preload` tells the browser maintainer that you are comfortable with this
+  domain being part of [the list shipped with the
+  browser](https://hstspreload.org/).
```
louis-antonopoulos: hstspreload.org says that thoughtbot.com isn't part of [the list](https://hstspreload.org/?domain=thoughtbot.com)? Am I misunderstanding how the site works?

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+
+Unrelated to HSTS but sort of a collorary of how the attack works: specify the
```
louis-antonopoulos: Typo: "collorary" -> "corollary"

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+
+## Passwords
+
+As has been mentioned, use bcrypt for storing your passwords in a database.
+
+Design the user experience to encourage your users to use a password manager:
+
+- Allow paste. Do the minimum to the password field -- and be sure to annotate
+  that it is a standard password field (`type="password"` in HTML,
+  `android:inputType="password"` in Android) -- so that the user's password
```
louis-antonopoulos: In Swift 4, it's `passwordTextField.isSecureTextEntry = true`.

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+    > will have to change them in the near future. When those changes do occur,
+    > they often select a secret that is similar to their old memorized secret
+    > by applying a set of common transformations such as increasing a number
+    > in the password. This practice provides a false sense of security if any
+    > of the previous secrets has been compromised since attackers can apply
+    > these same common transformations.
+- Allow passwords to be more complex. Whether you want to enforce light
+  password complexity rules or not ([NIST discourages password complexity
+  rules](https://pages.nist.gov/800-63-FAQ/#q-b6)), allow for passwords longer
+  or otherwise more complex than you expect. When possible, treat it as bytes
```
louis-antonopoulos: Suggestion: "treat passwords as bytes" for consistency; all previous mentions have used the plural "passwords" vs. the singular "it".

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+  that are immediately hashed and stored.
+
+A good password has a couple of properties:
```
louis-antonopoulos: Suggestion: "a few" instead of "a couple" since there are more than two properties.

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+an authentication secret.
+
+### Timing attacks
+
+An attacker can learn a lot from _how long_ it takes to be denied access. If
+it's instant, that means the input didn't even pass validation; if it's kinda
+long, that means that the input got past validation and computed a hash but one
+of the first few characters of the hash were incorrect; a longer delay means
+that most of the hash was right. Knowing how much of the hash was right allows
+the attacker to narrow the attack space.
```
louis-antonopoulos: This is brilliant! I hadn't thought about this before.

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+for constant-time comparisons.
+Worst case: pad the string to a fixed length then make sure your loop goes
+through every character even after you know the answer.
+
+## Multi-factor authentication (2FA)
+
+Given an email and password, you can authenticate as a user any time you wish.
+If someone were to mistakenly use the same password for multiple services, it
+is as strong as the least secure of those services: if the password were
+leaked, the password for all of those accounts are leaked at once.
```
louis-antonopoulos: Suggestion: "if the password were leaked, the password for all of those accounts would be leaked"

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+
+We can mitigate these kinds of attacks by requiring a second security factor --
+for example, a second password. We can go further by defining passwords of
+different nature:
```
louis-antonopoulos: Suggest converting "we" to "you" for consistency in both sentences. Also suggest "passwords of a different nature".

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+
+- Something you know, such as a string of letters.
+- Something you are, such as biometrics.
```
louis-antonopoulos: Given the previous comment, "Biometrics are useful for identity but incorrect to use as
an authentication secret", is it correct to suggest them here?

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+- Something you have, such as a phone.
+
+We can use a HOTP or TOTP algorithm to send and verify short codes out of band
+to something the user has, such as via email, SMS, or an external program.
```
louis-antonopoulos: There have been recent suggestions against SMS 2FA; it might be worth acknowledging them here. At the same time, SMS 2FA is better than no 2FA at all.

Example: https://www.makeuseof.com/tag/two-factor-authentication-sms-apps/

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+
+### OTP
+
+The HMAC-based One-time Password (HOTP) algorithm, RFC 4226, is a somewhat
```
louis-antonopoulos: Does "time" need to be capitalized?

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+straightforward function. The details can be found in the RFC but in summary it
```
louis-antonopoulos: If you have the link to the RFC, that'd be helpful.

security/application.md
```diff
@@ -0,0 +1,555 @@
...
+works like this: the client and server communicate a shared secret (typically
+via QR code). Whenever you need a one-time password, the shared secret is
+combined with an incrementing number, hashed, and then six digits are pulled
+out. Those six digits are the one-time password.
+
+Where can we get an incrementing number that both the client and server know
+about? We can use the number of minutes since the epoch. This gives us the
+Time-based One-time Password (TOTP) algorithm, RFC 6238.
+
+Most languages have a library for handling OTP. Ruby's is called rotp. As
```
louis-antonopoulos: Suggestion: `rotp` in single quotes.

security/application.md
mike-burns: Going with "text the user enters into your mobile app".

PR #534, opened 2019-03-21
Add guideline to avoid JSONB columns

jferris:
These columns are frequently abused to store has-one relationships when
somebody feels some aversion to adding several columns or a new table.

The actual use case for JSONB columns is slim and specific to storing
raw JSON efficiently. With a known schema, it's better to use separate
columns.

JSONB doesn't support the full range of Postgres types and constraints
(for example, there's no date type in JSON), it's more complicated to
index and query JSONB, it doesn't integrate as well with ORMs, and it's
harder to understand the schema when JSONB is involved.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -171,6 +171,8 @@ Postgres
 * Avoid multicolumn indexes. Postgres [combines multiple indexes] efficiently.
   Optimize later with a [compound index] if needed.
 * Consider a [partial index] for queries on booleans.
+* Avoid JSONB columns unless you have a strong reason to store an entire JSON
+  document from an external source.
 
 [combines multiple indexes]: http://www.postgresql.org/docs/9.1/static/indexes-bitmap-scans.html
 [compound index]: http://www.postgresql.org/docs/9.2/static/indexes-bitmap-scans.html


mike-burns:
Alternatively, how do you feel about this phrasing:

> Avoid storing data in a schemaless database column type.

jferris:
> Alternatively, how do you feel about this phrasing:

I tried to find a way to work in the JSONB example well but struggled. I think it's important to mention JSONB specifically and also to mention why it exists at all. It's not a bad column type, but it's a bad column type for every use case I've seen.

mike-burns:
Fair, yup. LGTM.

stepheneb:
Can you think of any uses where using a jsonb column type makes sense other than storing externally originating JSON documents?

jferris:
> Can you think of any uses where using a jsonb column type makes sense other than storing externally originating JSON documents?

No, I can't think of one. The only reason I've used one is to store the entire JSON document along with the columns I extracted from it. That can be useful later for debugging, if there was a bug in the extraction, or if additional fields could be useful later. The jsonb column type is more efficient and easier to migrate than text.

gylaz:
What about a situation where a front-end team needs to persist some JS object on the backend, and all they care about is sending in a, potentially deeply nested, JSON and getting back that JSON when they request it back. (No reason, or expectations, to query things inside that object). Would that warrant just throwing that JSON into a `jsonb` type? Or would the argument be that the back-end team should properly setup the necessary multiple tables (for nested structures) and de-struct and construct the JSON (which is obviously more work)?

mike-burns:
@gylaz that seems like a case where a TEXT or even BLOB would suffice, too.

6temes:
> Can you think of any uses where using a jsonb column type makes sense other than storing externally originating JSON documents?

For instance, when receiving data from an external service that you don't control. A good pattern, IMO, is to save the data inmediatly as JSON and acknowledge the reception. Then, process it. That way, if an error happens when processing (even if it's because of a bug on the senders' side) no information is lost and all the old events can be reprocessed.

PR #535, opened 2019-03-22
Change references of FactoryGirl to FactoryBot

jwkicklighter:
While going through these guides, we noticed that there were some places in the Readme that still referred to the old name `FactoryGirl` instead of the new `FactoryBot`.


georgebrock:
Rebased and merged as 09ac212f1c3bec7864e6f7dc0abac172c04d2c7f, with the addition of the example @WaKeMaTTa spotted.

Thanks, @jwkicklighter. Thanks, @WaKeMaTTa.

PR #538, opened 2019-04-02
Help people feel included and respected

mike-burns:
We cannot make someone feel anything -- their emotions and reactions are
on them. Additionally, following these guidelines does not guarantee
feelings of respect or inclusion.

The point is to help the situation.


enatario:
Also A+ branch name

PR #543, opened 2019-04-12
Update application.md

vcavallo:
Small grammatical change

### DIFF
--- a/security/application.md
+++ b/security/application.md
@@ -26,7 +26,7 @@ attack, likelihood of attack, ease of mitigating the attack, and severity of
 attack.
 
 Anything not in the list are things you cannot use as a reason to do something.
-Since the list is prioritized, you can use it to help it to prioritize tickets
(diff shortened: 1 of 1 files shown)


security/application.md
```diff
@@ -26,7 +26,7 @@ attack, likelihood of attack, ease of mitigating the attack, and severity of
 attack.
 
 Anything not in the list are things you cannot use as a reason to do something.
-Since the list is prioritized, you can use it to help it to prioritize tickets
+Since the list is prioritized, you can use it to help to prioritize tickets
```
tysongach: Thoughts on removing `to`, as well? So it would become “…you can use it to help prioritize tickets…”

vcavallo:
yep I originally did that too but then thought maybe that was more of a speech style choice than grammar and I didn’t want to stomp it. but either way is fine by me.

mike-burns:
Merged, thank you!

@tysongach removed the `to` in another commit.

PR #544, opened 2019-04-15
SHA-2 is NIST-approved

jkingdon:
The page being linked to is worded in a slightly confusing way but
https://csrc.nist.gov/Projects/Hash-Functions/NIST-Policy-on-Hash-Functions
is more explicit - SHA-2 is an approved hash algorithm.

If there is something other than the linked pages about why
SHA-2 might be considered dangerous, we could link it. But SHA-3
is on the new side (which is its own security risk), so I think
the real message here is "avoid SHA-1" rather than taking a
position on SHA-2 versus SHA-3.

### DIFF
--- a/security/application.md
+++ b/security/application.md
@@ -184,7 +184,7 @@ one-direction hash. When someone finds a way to generate the same hash for two
 different inputs, the hashing function is considered insecure. The American
 National Institute of Standards and Technology (NIST) maintains [a list of
 approved hash algorithms](https://csrc.nist.gov/Projects/Hash-Functions); as of
-this writing they recommend a SHA-3 algorithm.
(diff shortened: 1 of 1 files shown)


jkingdon:
Oh and as for who I am, I'm a devoted listener to the Bike Shed podcast and was really happy to see an episode about security and seeing this guide published publically. I am not a security engineer but am a security-interested application developer, and there is a real dearth of information which is accurate, up to date, well written, and accessible to application developers (OWASP is what people usually point to, but it tends to fall down on at least one of those criteria). Hence my enthusiasm about you publishing this guide.

mike-burns:
Good find, thank you! Merged as abebfeeae64d6c1442f12638fd0f8c7fe4e32a84.

PR #545, opened 2019-04-19
Update inclusive meeting guideline "Shine Theory"

entcheva:
Update "Shine Theory" guideline to include instructions for applying
shine theory.


enatario:
Woooohooo! Thanks @entcheva !

PR #547, opened 2019-04-30
Remove link to scss-lint config

tysongach:
We moved away from scss-lint and are using stylelint instead.

The scss-lint config this links to was removed in
https://github.com/thoughtbot/guides/commit/91ebdf6a3c96a1610a2d8f59ffd6ddd4f78cdd1f

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -1,9 +1,9 @@
 # Sass
 
 - [Sample](sample.scss)
-- [Default SCSS-Lint configuration](.scss-lint.yml)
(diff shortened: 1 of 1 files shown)


tysongach:
@setphen I don’t believe so! That’s a good point and perhaps I shouldn’t remove this text and link, but rather change it from scss-lint to styelint.

tysongach:
Updated based on above, in 3ba09b0e9675d1cb425a27a974fe9f52d8f903d8

PR #550, opened 2019-05-10
Scaffold a guide for GraphQL

lancejjohnson:
thoughtbot seems to be doing more work with GraphQL for client projects.
We would like to be to discuss and document what we discover as best
practices, recommended solutions, and issues to avoid for the benefit of
thoughtbot as we go. This commit provides an initial scaffolding for a
guide to GraphQL. Future commits can begin to fill in details.

### DIFF
--- a/graphql/README.md
+++ b/graphql/README.md
@@ -1,0 +1,61 @@
+# GraphQL
+
+A guide for GraphQL, a place to document questions, answers, best practices,
+smells, solutions to smells, etc. as we discover how to do GraphQL well.
+
+## Resources
+
+### Learning
+
+A curated list of resources for learning GraphQL.
+
+- **[Official GraphQL Learning Site](https://graphql.org/learn/)**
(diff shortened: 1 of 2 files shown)


graphql/README.md
```diff
@@ -0,0 +1,62 @@
+# GraphQL at thoughtbot
```
jferris: I would remove the "at thoughtbot" instances. The guide is under the thoughtbot org, and I don't believe these approaches are specific to thoughtbot's needs.

lancejjohnson: 👍

mike-burns:
I enjoyed:

https://github.com/Shopify/graphql-design-tutorial/blob/master/TUTORIAL.md
https://graphql-ruby.org/mutations/mutation_errors#errors-as-data

Also it took Wenger to point this out to me: GraphQL Relay is not the relay library but instead a useful thing to know about GraphQL.

mike-burns:
But also, this is great, thank you.

PR #551, opened 2019-05-10
Restructure CSS guides

tysongach:
I'd like to propose a restructuring of our guides, specifically our
CSS guides (at least for now).

Some of the pain points I have with the existing guides are:

- The current format (mostly bulleted lists) encourages brevity in our
  guides, which can be good and hopefully makes them easy to grasp, but
  potentially leads to a lack of educating and knowledge sharing within
  our guides.
- When reading the guides, it's difficult to understand the motivation
  or intent behind a guideline. If you're lucky, and with a little leg
  work, you can go through the commit history and find that someone left
  a detailed commit message explaining the "why" behind a specific
  guideline. But that takes extra work, and sometimes leads to not a
  succinct commit message, but instead a lengthy Pull Request with
  lots of back-and-forth. It can be hard to find the original intent.
- There's a lack of code examples that help to explain a guideline,
  next to the guideline. Sometimes we have an inline code snippet, but
  this limits how much we can expand on the code, like showing multiple
  code snippets to explain a guideline. We have a single `sample.scss`
  file to reference, but it's disconnected from the explanation and
  reasoning.
- The current list format, does not allow for deep-linking to specific
  guidelines. When reviewing PRs or talking about our guides, it can
  be useful to provide someone a URL to visit. We can't do that now for
  a specific guideline; only for the heading in which the guideline
  is within.

The changes in this commit hope to tackle some of these pain points by:

- Breaking up the bulleted lists into more logical groupings with
  headings. This allows us to add content that explains the guideline
  in more detail.
- Adding example code blocks, which can improve comprehension of the
  guideline and answer the question "what does this guideline mean, can
  you show me?"
- Adding content around the motivation behind a guideline. This may help
  answer questions like "why do we have this guideline?" and "how is
  this helpful?"
- Adding content on linting and/or other tooling.
- Adding content on related resources, like links to articles or code
  editor settings and tips.
- Wrapping expounding content in `<details>` (which creates a hide/show
  or accordion interaction that is collapsed by default), hopefully
  maintaining the brevity, while still allowing.

**Please note that this only restructures a portion of the CSS guidelines.
I want to get feedback and iterate before spending effort doing that.**

### DIFF
--- a/style/sass/README.md
+++ b/style/sass/README.md
@@ -9,7 +9,7 @@
 ## Formatting
 
 * Use the SCSS syntax.
-* Use hyphens when naming mixins, extends, classes, functions & variables: `span-columns` not `span_columns` or `spanColumns`.
+* Use hyphens when naming mixins, extends, functions & variables: `span-columns` not `span_columns` or `spanColumns`.
 * Use one space between property and value: `width: 20px` not `width:20px`.
 * Use a blank line above a selector that has styles.
 * Prefer hex color codes `#fff` or `#FFF`.
@@ -27,29 +27,178 @@
 * Use a `%` unit for the amount/weight when using Sass's color functions: `darken($color, 20%)`, not `darken($color, 20)`
 * Use a trailing comma after each item in a map, including the last item.
 
-## Order
-
-* Use alphabetical order for declarations.
-* Place `@extends` and `@includes` at the top of your declaration list.
-* Place media queries directly after the declaration list.
-* Place pseudo-classes and pseudo-elements second.
-* Place nested elements third.
-* Place nested classes fourth.
-
 ## Selectors
 
-* Don't use ID's for style.
-* Avoid over-qualified selectors: `h1.page-title`, `div > .page-title`
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
-* Be consistent about naming conventions for classes. For instance, if a project is using BEM, continue using it, and if it's not, do not introduce it.
 * Use ID and class names that are as short as possible but as long as necessary.
 * Avoid nesting more than 3 selectors deep.
 * Avoid using comma delimited selectors.
 * Avoid nesting within a media query.
-* Don't concatenate selectors using Sass's parent selector (`&`).
 
 ## Organization
 
 * Use a `base` directory for styling element selectors, global variables, global extends and global mixins.
 * Use HTML structure for ordering of selectors. Don't just put styles at the bottom of the Sass file.
 * Avoid having files longer than 100 lines.
+
+---
+
+## Selectors
+
+### Selector specificity
+
+- Don't use ID selectors.
+- Avoid over-qualified selectors, e.g. `h1.page-title`.
+
+<details>
+
+#### Code examples
+
+`h1.page-title` carries a specificity of 2, but can be reduced to 1 by removing
+the `h1` type selector:
+
+```diff
+-h1.page-title
++.page-title {
+   // …
+ }
+```
+
+#### Motivation
+
+Using an ID in a selector increases its specificity, making it more difficult
+to work with alongside class selectors. Furthermore, because IDs must be unique
+within an HTML document, using them as CSS selectors limits reusability.
+
+#### Resources
+
+- Learn about [how specificity is calculated][specificity-calculation].
+
+[specificity-calculation]: https://www.w3.org/TR/selectors-3/#specificity
+
+</details>
+
+### Selector naming
+
+- Use lowercase characters and hyphens (sometimes referred to as hyphen-case,
+  dash-case, or kebab-case) when naming selectors.
+- Be consistent about naming conventions. For instance, if a project is using
+  BEM, continue using it, and if it's not, don't introduce it.
+- Don't uses Sass parent selectors (`&`) to concatenate selector names.
+
+<details>
+
+#### Code examples
+
+Use lowercase characters and hyphens in selector names:
+
+```scss
+.class-name {
+  // …
+}
+```
+
+Don't concatenate selector names:
+
+```scss
+.class {
+  &__child-class {
+    // …
+  }
+}
+```
+
+#### Motivation
+
+Concatenating selector names makes it more difficult to search and find
+selectors in the codebase.
+
+</details>
+
+## General syntax and formatting
+
+### Declarations block ordering
+
+- Order declarations alphabetically.
+- Order items within the declaration block in the following order:
+    1. Sass at-rules, e.g. `@include`
+    1. CSS properties
+    1. Media queries
+    1. Pseudo-classes
+    1. Pseudo-elements
+    1. Nested elements
+
+<details>
+
+#### Code examples
+
+Alphabetize declarations:
+
+```scss
+.class {
+  display: block;
+  text-align: center;
+  width: 100%;
+}
+```
+
+Alphabetize prefixed properties as if the prefix doesn't exist:
+
+```scss
+.class {
+  font-family: system-ui;
+  -webkit-font-smoothing: antialiased;
+  font-weight: $weight-variable;
+}
+```
+
+Comprehensive example of ordering items within a declaration block:
+
+```scss
+.class {
+  @include size(10px);
+
+  display: block;
+  margin: $spacing-variable;
+
+  @media (min-width: $screen-variable) {
+    padding: $spacing-variable;
+  }
+
+  &:focus {
+    border-color: $color-variable;
+  }
+
+  &::before {
+    content: "";
+  }
+
+  .nested-element {
+    margin: $spacing-variable;
+  }
+}
+```
+
+#### Motivation
+
+Alphabetizing is automatable and is commonly a feature built into
+code editors (see Resources below).
+
+#### Linting
+
+Alphabetical declaration ordering can be linted using stylelint with the
+[stylelint-order][stylelint-order] plugin and its
+`order/properties-alphabetical-order` rule.
+
+[stylelint-order]: https://github.com/hudochenkov/stylelint-order
+
+#### Resources
+
+- Atom users can use the [Sort Lines package][sort-lines], which provides
+  commands and keybindings for alphabetical sorting.
+- Sublime Text users can use the `Edit > Sort Lines` menu item, or press
+  <kbd>F5</kbd> to sort lines alphabetically.
+
+[sort-lines]: https://github.com/atom/sort-lines
+
+</details>


style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
+---
+
+## Selectors
+
+### Selector specificity
+
+- Don't use ID selectors.
+- Avoid over-qualified selectors, e.g. `h1.page-title`.
+
+<details>
```
seanpdoyle: Using this element was going to be my VERY FIRST suggestion ❤️.

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
+
+#### Code examples
+
+`h1.page-title` carries a specificity of 2, but can be reduced to 1 by removing
+the `h1` type selector:
+
+```diff
+-h1.page-title
++.page-title {
```
seanpdoyle: In other guideline repositories that I've seen, a lot of the code samples have comments like:

```rb
# bad
the_code_with_the_violation

# bad
other_code_with_a_violation

# good
code_without_the_violation
```

Is the idea here that the added portion of the diff is the "good" example, and the removed portion is the "bad" example?

It's worth noting that I'm opposed to labelling "good" and "bad". If you know of less judgmental labels, let's use those!

tysongach: Yeah, I've seen this a lot, too ([Dropbox](https://github.com/dropbox/css-style-guide) and [Airbnb](https://github.com/airbnb/swift) being some examples).

Your point about the labels “good” and “bad” are exactly why I didn't go down that path, but at the same time, I do like the clarity it may bring. 

> Is the idea here that the added portion of the diff is the "good" example, and the removed portion is the "bad" example? If so, have we explained that?

Yes, and no, that is not explained right now.

Definitely open to suggestions here…

style/sass/README.md
vendelal: Just throwing this out here for discussion: could we name this element something other than "Details?" perhaps something like "More" or "Tell me why" or "The Why" or anything better than those suggestions 😄

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
+
+#### Motivation
+
+Using an ID in a selector increases its specificity, making it more difficult
+to work with alongside class selectors. Furthermore, because IDs must be unique
+within an HTML document, using them as CSS selectors limits reusability.
+
+#### Resources
+
+- Learn about [how specificity is calculated][specificity-calculation].
```
vendelal: I really appreciate sharing these resources. If these were available for each section that would be so cool!

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
+
+[specificity-calculation]: https://www.w3.org/TR/selectors-3/#specificity
+
+</details>
+
+### Selector naming
+
+- Use lowercase characters and hyphens (sometimes referred to as hyphen-case,
+  dash-case, or kebab-case) when naming selectors.
```
vendelal: 💞 _kebab-case_ 💞

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
+</details>
+
+## General syntax and formatting
+
+### Declarations block ordering
+
+- Order declarations alphabetically.
+- Order items within the declaration block in the following order:
+    1. Sass at-rules, e.g. `@include`
+    1. CSS properties
```
vendelal: should these numbers be 1, 2, 3, 4... instead of all 1s?

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
+
+Alphabetical declaration ordering can be linted using stylelint with the
+[stylelint-order][stylelint-order] plugin and its
+`order/properties-alphabetical-order` rule.
+
+[stylelint-order]: https://github.com/hudochenkov/stylelint-order
+
+#### Resources
+
+- Atom users can use the [Sort Lines package][sort-lines], which provides
```
vendelal: Again, awesome to include this stuff! It has made my life much easier and it's great that someone else could find these resources right here.

style/sass/README.md
vendelal: I also like this idea, and I think it would be especially helpful if the "bad" example was something we might see a lot or that could appear fine at first glace. What if the labels depended on the context of the example? For instance, in this one the labels could be `# too specific` and `# just specific enough` or `# harder to override` and `# specific but overrideable` or... I think you get my point 😅

style/sass/README.md
seanpdoyle: This makes use of the newer [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element. I think GitHub polyfills support.

I think we can specify "More", "Tell me why", or "The Why" in a [`<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary) element:

<details>
  <summary>Click on me to expand!</summary>

```html
<details>
  <summary>Click on me to expand!</summary>
</details>
```
</details>

style/sass/README.md
tysongach: Fun little Markdown tip: these get rendered as a proper ordered list, 1, 2, 3, 4.

Raw Markdown:

```
1. item
1. item
1. item
```

Rendered Markdown:

1. item
1. item
1. item

I like to use all `1`s in the Markdown, because it makes shuffling or moving anything in the list much easier, as you don't need to then go through and re-order the numbers.

danbee: But isn't the point of Markdown to create a human readable text file too?

style/sass/README.md
```diff
@@ -9,7 +9,7 @@
 ## Formatting
 
 * Use the SCSS syntax.
-* Use hyphens when naming mixins, extends, classes, functions & variables: `span-columns` not `span_columns` or `spanColumns`.
+* Use hyphens when naming mixins, extends, functions & variables: `span-columns` not `span_columns` or `spanColumns`.
 * Use one space between property and value: `width: 20px` not `width:20px`.
 * Use a blank line above a selector that has styles.
 * Prefer hex color codes `#fff` or `#FFF`.
```
ericwbailey: What do you think about using the longhand hex value? I know it's nitpicky, but I like the consistency of always having one treatment for the declaration type.

style/sass/README.md
tysongach: Perhaps…

It is [billed as](https://daringfireball.net/projects/markdown/) an “…easy-to-read, easy-to-write plain text format.”

Though, I feel like the consumption of this content happens more often in the rendered state. In HTML, each item within an ordered list are all represented as `<li>`, which is comparable to using all `1`s in Markdown.

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
-* Place nested elements third.
-* Place nested classes fourth.
-
 ## Selectors
 
-* Don't use ID's for style.
-* Avoid over-qualified selectors: `h1.page-title`, `div > .page-title`
 * Use meaningful names: `$visual-grid-color` not `$color` or `$vslgrd-clr`.
-* Be consistent about naming conventions for classes. For instance, if a project is using BEM, continue using it, and if it's not, do not introduce it.
 * Use ID and class names that are as short as possible but as long as necessary.
```
ericwbailey: Remove ID from here since we're encouraging people not to use IDs as selectors?

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
 * Avoid nesting more than 3 selectors deep.
 * Avoid using comma delimited selectors.
```
ericwbailey: Curious about this one. Is this a "if you're using comma delimited selector list you should should probably write a mixin" kind of thing? Or is it pushing for encapsulating style within a component, and not trying to be too "efficient" with a vanilla CSS-style authoring approach.

I'm thinking situations like `:hover`/`:focus`, where you would want to combine.

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
 * Avoid nesting within a media query.
```
ericwbailey: Maybe mention prefer including media queries within a component? It's one of my favorite Sass features.

style/sass/README.md
```diff
@@ -27,29 +27,178 @@
...
-* Don't concatenate selectors using Sass's parent selector (`&`).
 
 ## Organization
 
 * Use a `base` directory for styling element selectors, global variables, global extends and global mixins.
 * Use HTML structure for ordering of selectors. Don't just put styles at the bottom of the Sass file.
```
ericwbailey: I'm a little unclear what "Use HTML structure means."

style/sass/README.md
ericwbailey: I like "Tell me why." You're not clicking if you're using this on mobile!

style/sass/README.md
ericwbailey: Casting a vote for using 1s. We use it elsewhere, and it's a pretty common pattern for a lot of Markdown documentation.

style/sass/README.md
tysongach: I don’t have a strong preference and am fine changing to longhand, but I think that’s out of scope of the PR. 

My intention with this PR was to focus on a new format and structure of how we organize the guides, but without changing any of the specific guidelines.

ericwbailey: Gotcha.

tysongach:
@ericwbailey All of your points are valid, but my intention with this PR was to focus on a new format and structure of how we organize the guides, without changing any of the specific guidelines.

style/sass/README.md
vendelal: TIL! thanks for explaining :)

tysongach:
@ericwbailey In fact, I think a lot of your points above are _exactly_ why I opened this PR and why I think we should restructure things. My answer to your questions are mostly “I don’t know and it’s hard to know why this guideline exists”, followed quickly by a “head-nod.” 😄

ericwbailey:
Ah, forest for the trees, sorry. I'll resolve and approve and make a note to myself to followup with a PR.

tysongach:
> and make a note to myself to followup with a PR.

@ericwbailey Definitely - if you’d like to open separate PRs for these suggestions, I think it would be super helpful. It feels like a handful of guides and simply go away, and so removing them before doing more re-organizing could be beneficial.

style/sass/README.md
tysongach: I’m going to sticky with all `1`s.

style/sass/README.md
tysongach: I definitely intend to keep adding more!

style/sass/README.md
tysongach: I just reviewed this and tried out some of the suggestions, but I think I like the default “Details”…

“Tell me why” seems to describe the “Motivation” content that is expanded, but we’re also nesting “Resources” “Linting”, and “Code examples” which don’t fit the descriptor “tell me why.” Details, to me, seems to be more inclusive to all the things we’re nesting within the `<details>` element.

cpytel:
Merging this as it's been ready to go for some time.

Based on the comments, it sounds like there is interest in revising some of the actual content.

PR #556, opened 2019-07-25
Prefer < over >

eebs:
From https://www.rubytapas.com/2019/07/25/dont-use-greater-than/ written
by Llewellyn Falco.

We want to express the condition X is greater than 5 and less than 10.
There are eight different ways that you can express this concept in
code, but they all evaluate the same.

```ruby
5 < x && x < 10
x < 10 && 5 < x
10 > x && x > 5
x > 5 && 10 > x
5 < x && 10 > x
x < 10 && x > 5
10 > x && 5 < x
x > 5 && 10 > x
```

This can be quite confusing, what if we only consider ones that use the
less than `<` sign.

```ruby
5 < x && x < 10
x < 10 && 5 < x
```

If we order the known values, we are only left with one option

```ruby
5 < x && x < 10
```

The consistency of always using `<`, and ordering by value in compound
conditionals creates only a single way to express this sort of code.
This can make code easier to read and reason about.

### DIFF
--- a/style/README.md
+++ b/style/README.md
@@ -54,6 +54,7 @@ Formatting
   `}`.
 * Use [Unix-style line endings][newline explanation] (`\n`).
 * Use [uppercase for SQL key words and lowercase for SQL identifiers].
+* Prefer `<` over `>`. [#556]
 
 
 [dot guideline example]: /style/ruby/sample.rb#L11


style/ruby/README.md
```diff
@@ -22,6 +22,7 @@ Ruby
 * Prefer double quotes for strings. [#174]
 * Prefer `&&` and `||` over `and` and `or`. [#153]
 * Prefer `!` over `not`. [#195]
+* Prefer `<` over `>`
```
mike-burns:
```suggestion
* Prefer `<` over `>`. [#556]
```

mike-burns:
Could this go in the top-level `style/README.md` instead of being Ruby-specific?

eebs:
It could. Which section do you think it should go under, Formatting or Organization?

mike-burns:
Hm, interesting. Maybe it goes under `best-practices/README.md`, General, instead. Or maybe `style/README.md` should have a Principles section. Or maybe it should stay here. I don't know.

iwz:
@eebs that link doesn't give a summary of why this is a good rule-of-thumb. Is there a link to a transcript available?

mike-burns:
That link has a transcript in it.

Despite that, I agree with iwz: do you mind providing an argument for why we should have this as part of this PR description so other people can easily know why we're doing this?

iwz:
Do I have to log in to see the transcript?

mike-burns:
Yes.

iwz:
The transcript is great!

eebs:
Thanks for the feedback @mike-burns and @iwz. I've added the example from the transcript which hopefully illustrates the argument. I've also moved the advice to the style readme.

mike-burns:
Thank you! LGTM but it'd be nice to hear from others since this is top-level.

JoelQ:
Are there other Ruby constructs we might prefer over greater than / less than to describe intervals? For example `Range#cover?`

georgebrock:
What about phrasing this as:

> Consistently use `<` or `>` within a single conditional.

That seems to maintain the benefit of avoiding hard-to-read conditionals that use a mix of `<` and `>`, but also allows the operator to be chosen that best expresses the condition.

For example, I've encountered code this week that checks if a user is 18 years old or older. I find `age >= 18` to be more expressive of that idea than `17 < age`.

eebs:
> For example, I've encountered code this week that checks if a user is 18 years old or older. I find `age >= 18` to be more expressive of that idea than `17 < age`.

That's a great point.

Maybe there isn't a simple way to phrase this, and it's best left to people's own discretion?

jferris:
> Maybe there isn't a simple way to phrase this, and it's best left to people's own discretion?

Yes, I agree with this.

Another common example would be `x > 0`, which reads awkwardly as `0 < x`. I think the larger issues are around ranges, which have better solutions than eliminating an operator.

eebs:
Thanks for the discussion everyone, I'm closing this as it seems best to not try and make it a guide. ❤️

PR #557, opened 2019-07-26
Use trailing commas in ERb

tysongach:
We generally encourage trailing commas throughout our guides. For
example, in Sass maps and for multi-line lists in Ruby.

This establishes that guidance for ERb as well, when we have a
mutli-line method call. The trailing comma makes for clean diffs, when
adding or removing argument.


mike-burns:
Grammar-o in the last sentence of the commit message. Should be:

> The trailing comma makes for clean diffs when adding or removing an argument.

Diff LGTM.

PR #558, opened 2019-07-26
Prefer ES6 to CoffeeScript

tysongach:
We've been discouraging the use of CoffeeScript for a while now, though, I'm lacking the specifics as to why. It would be good to add those details within this commit and the CoffeeScript readme, if we decide to merge this.


mike-burns:
Into it.

Instead of saying don't use it, how about we say what to use? ES6 seems like the preferred answer.

One of the reasons CS has been replaced is that many of the good parts of it have made their way into ES6. ES6 is more widely known and supported, and most JS documentation is described in ES6 now.

The charts in section 3 of https://blog.aha.io/coffescript-to-es6/ show the big picture.

JoelQ:
👍  to something along the lines of "Prefer ES6 to Coffeescript"

tysongach:
Thanks @mike-burns and @JoelQ! I’ve updated this PR with your input and suggestions.

PR #559, opened 2019-07-26
Add some guidelines for Scala

jferris:
This is pretty bare, but it serves as a place to add more guidelines as we agree
on them.

### DIFF
--- a/style/scala/README.md
+++ b/style/scala/README.md
@@ -1,0 +1,10 @@
+# Scala
+
+* Avoid `var`.
+* Avoid implicit conversions.
+* Avoid wildcard imports except for implicits.
+* Avoid `null` except when necessary for Java compatibility.
+* Prefer a lazy effect type like `cats.effect.IO` to `scala.concurrent.Future`.
+* Use [scalafmt] to automatically format as much as possible.
+
+[scalafmt]: https://scalameta.org/scalafmt/


style/scala/README.md
```diff
@@ -0,0 +1,10 @@
+# Scala
+
+* Avoid `var`
+* Avoid implicit conversions
+* Avoid wildcard imports except for implicits
+* Avoid `null` except when necessary for Java compatibility
+* Prefer `IO` to `Future`
```
christoomey: I'm guessing that `IO` here refers to either Cats or Monix, but less so Scalaz. Worth annotating with pointers, or are you purposefully keeping this more general?

jferris: I clarified this a bit in c4ccfa1. I basically want to avoid `Future` and `cats.effect.IO` is an easy choice for many apps, but `monix.eval.Task` or `zio.IO` are also good.

style/scala/README.md
```diff
@@ -0,0 +1,10 @@
+# Scala
+
+* Avoid `var`
+* Avoid implicit conversions
+* Avoid wildcard imports except for implicits
+* Avoid `null` except when necessary for Java compatibility
+* Prefer a lazy effect type like `cats.effect.IO` to `scala.concurrent.Future`
+* Use [scalafmt] to automatically format as much as possible
+
```
johnschoeman: What do you think about adding: `Prefer explicit return types`?

johnschoeman:
> Do we need to add any justification/explanation for these recommendations?

I've found reading the PR comments/discussion to be sufficient for context on why something is in the style guide. Are you suggesting adding explanations to the markdown itself?

jferris:
> Do we need to add any justification/explanation for these recommendations?

We've generally just had lists of guidelines without justification.

PR #563, opened 2019-08-15
Expand GraphQL guide

jferris:
* Shorten tutorial section
* Include reference to GraphQL spec
* Include best practices for status codes, error handling, and mutation
  responses.

### DIFF
--- a/graphql/README.md
+++ b/graphql/README.md
@@ -1,11 +1,9 @@
 # GraphQL
 
-A guide for GraphQL, a place to document questions, answers, best practices,
-smells, solutions to smells, etc. as we discover how to do GraphQL well.
+A guide for building GraphQL servers and clients.
 
-## Resources
 
-### Learning
+## Learning
 
 A curated list of resources for learning GraphQL.
 
@@ -19,7 +17,7 @@ A curated list of resources for learning GraphQL.
   API](https://www.dropbox.com/s/svqe68hpdiixf0g/presentation.pdf?dl=0)**
   Presentation introducing GraphQL to thoughtbot.
 
-### Public GraphQL APIs
+## Public GraphQL APIs
 
 Publicly available GraphQL APIs allowing you to explore how GraphQL is and can
 be used.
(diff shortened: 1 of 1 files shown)


graphql/README.md
```diff
@@ -1,25 +1,12 @@
...
-smells, solutions to smells, etc. as we discover how to do GraphQL well.
+A guide for building GraphQL servers and clients.
 
-## Resources
+If you're just getting started, check out the [official GraphQL tutorial].
 
-### Learning
+[official GraphQL tutorial]: https://graphql.org/learn/
 
-A curated list of resources for learning GraphQL.
```
jferris: I tightened this up a bit as our guides are generally succinct and serve as reference points rather than being tutorials. Some of these original resources also included non-public links, which makes them less valuable as a community reference.

graphql/README.md
```diff
@@ -41,24 +28,22 @@ be used.
...
-
-### Deleting Resources
-
-## Authn, Authz
-
-## Open Questions
-
-
+* Follow the latest version of the [GraphQL specification].
+* When serving over HTTP, respond with a 200 OK status code to all GraphQL
```
JoelQ: We're blocking unauthenticated requests to the GraphQL API on ADS with a 401. Is that a valid case for a non-200 response?

graphql/README.md
mike-burns: We often have people join the `#graphql` channel and ask where to start learning it, how did we learn it, etc.? Having those links somewhere would be useful.

graphql/README.md
mike-burns: [No](https://github.com/rmosolgo/graphql-ruby/issues/1130#issuecomment-347373937):

> Non-200 response codes indicate that some issue occurred at the HTTP transport layer

All GraphQL responses that we are in control of should be 200s.

graphql/README.md
jferris: Okay, that's reasonable. I restored this section in 3c89384.

graphql/README.md
jferris: Yes, after more experience with GraphQL, I agree that we didn't do it the best way on that project.

graphql/README.md
```diff
@@ -41,24 +39,22 @@ be used.
...
+  queries.
+* If a client or server error occurs, use the `errors` key in the GraphQL
+  response.
+* If a user-facing error occurs (such as invalid user input), use the `data` key
+  in the GraphQL response.
+* If a mutation can fail because of a user error, use a union type to describe
+  the possible outcomes.
+* If there is an authenticated user, provide the user in the context for the
```
mike-burns: Is this an implementation detail?

jferris: I don't think it is? I can't find a mention of the ctx/context argument in the spec, but it's been in every GraphQL server implementation I've seen. Specifying where authentication lives seems like a useful thing to decide once instead of on every project.

PR #565, opened 2019-08-19
[RUBY] Single quotes preferred style

ssunday:
Why:
- It's default https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/StringLiterals
- Here it's also default and suggested as default https://github.com/bbatsov/ruby-style-guide#consistent-string-literals (it's his opinion but also adopted by others on a large scale)
- It makes it clearer when interpolation is happening
- Generally cleaner looking

Update:

This: https://github.com/thoughtbot/guides/pull/174#issuecomment-43630455

### DIFF
--- a/style/ruby/.rubocop.yml
+++ b/style/ruby/.rubocop.yml
@@ -346,7 +346,7 @@ Style/SpecialGlobalVars:
 Style/StringLiterals:
   Description: 'Checks if uses of quotes match the configured preference.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#consistent-string-literals'
-  EnforcedStyle: double_quotes
+  EnforcedStyle: single_quotes
   Enabled: true
 
 Style/TrailingCommaInArguments:


iwz:
The change to double quotes was a controversial discussion when introduced. You may find the discussion, reasoning, and eventual decision interesting, @ssunday https://github.com/thoughtbot/guides/pull/174

Personally,  since adopting this change to always use double quotes, I've always appreciated not having to think about which quotes to use.

mike-burns:
Some context on how this PR came to be: @ssunday is collaborating with us on a project in NYC and has had to debug our code.

Some history of the thoughtbot styleguide with a focus on quote marks:

- We started the styleguide so we could stop having discussions about arbitrary style and just pick one thing without ever changing it.
- We [originally](https://github.com/thoughtbot/guides/blob/36491dbb949bac298475a2615eb11232531162b5/README.md#ruby) specified single quotes, as was the Ruby style at the time.
- Someone proposed [changing an arbitrary thing we picked](https://github.com/thoughtbot/guides/pull/169). That was rejected, because that defeats the point of no longer discussing these things.
- Someone [proposed](https://github.com/thoughtbot/guides/pull/237) changing it again, and it was changed.
- We figured that if we can change arbitrary things we might as well [re-discuss quotes](https://github.com/thoughtbot/guides/pull/174).
- We changed to double quotes, as was the Ruby style at the time.

As usual I'd like to promote the idea of deleting the styleguide.

ssunday:
@iwz It's a really interesting thread! Lots definitely changed since then...and not at all! Some of the points brought up are still true (both sides). 

With rubocop autofixing where it is, enforcing single/double quotes removes the Human adoption element to a degree. You could 'forget' to use single and it corrects it if it could/should. (This is assuming rubocop is also being used to correct things). So to me, this then comes to what is easier to read and manage. Are single quotes easier for our minds to process? It takes up less visual space...does it add up? To me, it looks cleaner and seems more elegant. 

Also, which has been said: https://github.com/thoughtbot/guides/pull/174#issuecomment-43630455 

/s: Has there been studies if double quote usage increases Carpal Tunnel Syndrome?? end/s

odlp:
> Personally, since adopting this change to always use double quotes, I've always appreciated not having to think about which quotes to use.

I particularly appreciate this when writing test descriptions. Please excuse the overly specific example - but I hope it provides a useful illustration. 

Say I'm trying to describe the edge-cases where something shouldn't take place:

```ruby
RSpec.describe Foo do
  it 'doesn' # Brain: Uh oh...
end
```

One then has to break their flow and pick an option...

1. Convert to double quotes and continue:

```ruby
it "doesn't notify via Slack"
```

2. Reword to avoid contractions:

```ruby
it 'does not notify via Slack' # Hmm I didn't want to replace "doesn't" with "does not"
```

3. Be lawful evil (aside: Is this how this Alignments thing works?):

```ruby
it 'does\'t not notify via Slack' # Nope
```

JoelQ:
Inspired by @odlp's comment, I threw this together 😂😂😂 

|             | Good                    | Neutral          | Evil            |
|-------------|-------------------------|------------------|-----------------|
| **Lawful**  | `'does not'`            | `'doesn’t'`      | `'doesn\'t'`    |
| **Neutral** | `%q{doesn't}`           | `"doesn't"`      | `'doesnt'`      |
| **Chaotic** | `<<TXT.chomp don't TXT` | `"doesn#{"'"}t"` | `'doesn'"'"'t'` |

mike-burns:
@jferris I don't know how to move this PR forward, in any direction. Do you?

ssunday:
@odlp if you're assuming people are automatically doing this....to me, you could use double quotes and have the linter enforce the proper option. Converting to double quotes for apostrophes is also, I think, acceptable, as sometimes you'll need to use double quotes anyway for other reasons. The goal isn't to eliminate double quotes, but have single quotes be preferred as it's easier to type and parse.

jferris:
I'm going to close this pull request, because I'm not convinced that going back to single quotes will be worth the required changes we'd need to make.

People are welcome to continue discussing quoting styles, but I don't plan on merging this.

ssunday:
@jferris 👍 It's your style guide/linters to run, can't argue with that perspective.

composerinteralia:
More information for the next time this comes up: [Rails itself prefers double quotes](https://github.com/rails/rails/blob/b5f47f755681d055a2b64acf1a091e76298fbce7/.rubocop.yml#L167) and they [plan to clean up any lingering single quotes](https://github.com/rails/rails/pull/38841).

PR #568, opened 2019-09-12
First pass at React Guide

das3in:


protocol/react/README.md
```diff
@@ -0,0 +1,115 @@
+React Protocol
+==============
+
+A guide for writing great front-end web apps.
+
+General Philosophies
+--------------------
+
+* For greenfield `react` projects we like to use `Typescript`. `Typescript` is a
```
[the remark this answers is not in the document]
das3in: I think I saw the backticks in a a couple of blog posts and just thought it was a new convention. It doesn't feel correct to me so I've removed them

protocol/react/README.md
```diff
@@ -0,0 +1,121 @@
...
+libraries to facilitate this.
+
+Add `eslint` to the project
+
+    yarn add eslint @typescript-eslint/parser @typescript-eslint/eslint-plugin --dev
+
+The following is a `.eslintrc.js` file that has some sensible defaults. Add this
+to the root of your project.
+
+<details>
```
christoomey: Typically you'll want to include the summary in `<summary>` element, and that will be rendered as the text with the disclosure triangle (rather than the generic `"Details"`).

The following example shows using it w/ a JS formatted code block (view the raw comment to see the specific markup as it is a bit tricky).

<details>
  <summary>Sample <code>.eslintrc.js</code> file with some sensible defaults</summary>

``` js
module.exports = {
  end: {
    node: true,
    es6: true,
  }
}
```
</details>

protocol/react/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+    module.exports =  {
+      semi:  true,
+      trailingComma:  'all',
+      singleQuote:  true,
+      printWidth:  120,
+      tabWidth:  2,
+    };
```
christoomey: Some of my previous notes on the prettier config seemed to be lost in the shuffle:

- `printWidth` ideally would be 80 as this matches our other guides, and [perttier's own recommendations](https://prettier.io/docs/en/options.html#print-width)
- `semi: true` is default, so we can omit
- `tabWidth: 2` is default, so we can omit

protocol/react/README.md
```diff
@@ -0,0 +1,121 @@
...
+            argsIgnorePattern: '^_',
+          },
+        ],
+      },
+    };
+</details>
+
+And here's a `.prettierrc.js` file with some good default code styles.
+
+<details>
```
christoomey: Similar comment re the details>summary structure as the eslint section.

Alain1405:
@das3in Thank you for the guide! I've noticed "React" items are missing on TOCs [here](https://github.com/thoughtbot/guides/blob/cr-add-react-guide/README.md) and [there](https://github.com/thoughtbot/guides/blob/cr-add-react-guide/protocol/README.md)

das3in:
> @das3in Thank you for the guide! I've noticed "React" items are missing on TOCs [here](https://github.com/thoughtbot/guides/blob/cr-add-react-guide/README.md) and [there](https://github.com/thoughtbot/guides/blob/cr-add-react-guide/protocol/README.md)

Great catch thanks!

PR #578, opened 2019-12-03
Document app/services

mike-burns:
Many -- perhaps all -- of our projects have an `app/services` directory.
This diff documents the Platonic ideal of the practice: it is a place for
command objects that orchestrate interactions between two or more
models.

This is in comparison with `app/models`, which holds both ActiveRecord
objects and also data structures related to the app itself.

This is also distinct from `lib`, which is seldom used and often a good
place to move files before extracting them into a gem.

This PR is the result of an internal discussion that happened because of
https://github.com/thoughtbot/suspenders/pull/1015 .

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -96,6 +96,7 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryBot factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
+* Use `app/models` for model objects and `app/services` for command objects.
 
 [date-block]: /best-practices/samples/ruby.rb#L10
 [fkey]: http://robots.thoughtbot.com/referential-integrity-with-foreign-keys
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -96,7 +96,7 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryBot factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
-* Use `app/models` for model objects and `app/services` for command objects.
+* Use `app/models` for domain objects and `app/services` for command objects.
 
 [date-block]: /best-practices/samples/ruby.rb#L10
 [fkey]: http://robots.thoughtbot.com/referential-integrity-with-foreign-keys


best-practices/README.md
```diff
@@ -96,6 +96,7 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryBot factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
+* Use `app/models` for model objects and `app/services` for command objects.
```
JoelQ: "model" means different things to different people. Depending on who you ask, a model might be only objects that inherit from Active Record or it might be any domain object.

mike-burns: I'll change this to say "domain object" instead of "model object".

best-practices/README.md
```diff
@@ -96,6 +96,7 @@ Rails
   environment variables are detected on deploy.
 * [Use blocks][date-block] when declaring date and time attributes in FactoryBot factories.
 * Use `touch: true` when declaring `belongs_to` relationships.
+* Use `app/models` for domain objects and `app/services` for command objects.
```
sharplet: Do you have a reference for the term _command object_? As a primarily non-Rails dev, I'm unfamiliar with the term.

In regards to the use of `app/services` in particular: the main reason I'm in favour of putting things in `lib` or `app/whatever`, is that I find the use of the term _service_ to represent things other than external APIs and interfaces (web services, databases, background processes, etc.) a little confusing.

jsilasbailey: When I see `services` in rails world I think of service objects which I've seen do a lot of very different things (ie: [this article looking at how different people have referenced them](https://blog.lelonek.me/what-service-objects-are-not-7abef8aa2f99)).

When I see `command object` I assume that is referencing the [command pattern](https://en.wikipedia.org/wiki/Command_pattern).

Most of my experience with an `app/services` directory has been for objects that, don't necessarily follow the command pattern (they may) but that the most important thing about them is they represent a business flow in the application that deals with multiple domain model interaction (ie: a sign up flow).

Because of the general differences in the way that I've seen `service object` referenced in rails world and the meaning of the `command pattern: command object`, I would be in favor of either:

A: Keep the directory as `app/services` but change the wording from `command object` to `service object` and link to a more fleshed out description of what a service object is

OR

B: Change the directory to `app/commands` and keep the `command object` wording

JoelQ: Interesting thoughts on service objects from the classic CodeClimate article on [7 ways to decompose fat active record models](https://codeclimate.com/blog/7-ways-to-decompose-fat-activerecord-models).

>Some actions in a system warrant a Service Object to encapsulate their operation. I reach for Service Objects when an action meets one or more of these criteria:
>
> 1. The action is complex (e.g. closing the books at the end of an accounting period)
> 2. The action reaches across multiple models (e.g. an e-commerce purchase using Order, CreditCard and Customer objects)
> 3. The action interacts with an external service (e.g. posting to social networks)
> 4. The action is not a core concern of the underlying model (e.g. sweeping up outdated data after a certain time period).
> 5. There are multiple ways of performing the action (e.g. authenticating with an access token or password). This is the Gang of Four Strategy pattern.

jsilasbailey: ^ Yes this was the article I was originally looking for 👍

mike-burns:
Some history: https://trello.com/c/K4vAgZVi/38-use-lib-instead-of-app-models-for-the-service-layer

sharplet:
Love this part where @jferris speculates about what "Service-oriented architecture" means: https://trello.com/c/K4vAgZVi/38-use-lib-instead-of-app-models-for-the-service-layer#comment-51115bae07b55bc25c000e11.

mike-burns:
Do we want any part of this diff?

jferris:
It doesn't feel as though we ever reached an agreement on what the directory should contain or what "service object" means, so I don't think we should merge this. The directory feels like a junk drawer, like "utils," which contains objects that you want to abstract but don't follow a particular pattern.

I'm going to close this for now, but feel free to continue discussing.

PR #586, opened 2020-02-22
Move “Releasing a Ruby Gem”

tysongach:
I can never seem to find this section on how to release a Ruby gem.
I think it makes more sense as a ‘how-to.’ It’s also not technically
limited to open-source; you can host your gem server, for example.

### DIFF
--- a/how-to/README.rst
+++ b/how-to/README.rst
@@ -53,3 +53,16 @@ capybara-webkit. For example, in ``spec/features/user_signs_in_spec.rb``:
   end
 
 .. _capybara-webkit: https://github.com/thoughtbot/capybara-webkit
+
+... release a Ruby gem
+--------------------------
+
+* Edit the ``VERSION`` constant.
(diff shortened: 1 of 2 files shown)


composerinteralia:
We also have this releasing template: https://github.com/thoughtbot/templates/blob/master/RELEASING.md.template

Could we link to that instead?

tysongach:
Good question. Linking to a template feels a bit odd - it’s hard to read since it is a `.template` file.

I also sort of think that template is awfully specific to gems. I kinda wonder if it should be generalized a bit to support things like releasing npm packages. If we go down that path, it would be nice to keep a short ‘how-to’ here in this repo on releasing a Ruby gem, and maybe even _add_ a guide on releasing an npm package. Put another way: I see that template as a starting point for a file to add to your repo which should _be customized_ to your project's releasing steps - and that is not necessarily a Ruby gem.

PR #587, opened 2020-02-22
Convert RST to Markdown

tysongach:
We have a lot of Markdown in this repo, this is the only document in
RST format. Let's follow the convention.


mike-burns:
Did we lose the table of contents in this move? That's why the original document was in RST; Markdown doesn't provide the equivalent of the `.. contents::` command.

tysongach:
We did. This was the only file that had a table of contents. Do we feel that’s important to maintain? We also only have 2–3 items in the table of contents.

mike-burns:
It's a long and scattered document, but you're right that it only has a few sections. I'll open a PR with a manual TOC.

PR #588, opened 2020-02-22
Add HTML validation best practice

tysongach:
Running your HTML through a validator can uncover all sorts of issues
and helps you ensure that your markup is accessible and in-line with
Web standards.

### DIFF
--- a/html/README.md
+++ b/html/README.md
@@ -1,5 +1,6 @@
 # HTML
 
+- Use the [W3C's Markup Validation Service][html-validator] to validate HTML
 - Prefer double quotes for attributes.
 - Use lowercase text for elements and attributes
 - Use double quotes to wrap element attributes
@@ -27,5 +28,6 @@
 - Ensure heading elements are used to section content, and heading levels are
   not skipped
 
+[html-validator]: https://validator.w3.org/
 [normal elements]: https://html.spec.whatwg.org/multipage/syntax.html#normal-elements
 [parent elements contain no more than 60 child elements]: https://developers.google.com/web/tools/lighthouse/audits/dom-size


best-practices/README.md
```diff
@@ -217,6 +217,7 @@ JavaScript
 HTML
 ----
 
+* Use the [W3C's Markup Validation Service][html-validator] to validate HTML
```
JoelQ: Are we expected to run this manually on every page of a project?

tysongach:
@JoelQ Good points and questions.

Like accessibility testing, it’s tough to automate HTML validation, in my findings. There are some tools that can help automate validation of HTML in a piecemeal fashion or on the element-level, but I’m unaware of any tools are comprehensive, i.e. validate the whole HTML document. That’s because it’s common for us to use things like layouts and partials to construct an HTML document, and so our tooling isn’t seeing the whole of the end document. We’re also often working in templating languages where the input might look good, but the output HTML has issues.

I tend to validate a document or two a week - just every once in a while I’ll paste a page’s source into the validator or pass it a production or staging URL. While it is manual, it only takes about one minute and I often catch things and learn new things. I also find that when working in an object-oriented fashion with our front-end, testing even _some_ of your HTML documents will identify issues that, if fixed in the component abstraction, fixes the issues on _all_ page’s that use that component.

Does anyone know of any automated tooling in this space?

best-practices/README.md
sneamonitakis: I believe, and @tysongach correct me if I am wrong, it would make the most sense to run it on whatever HTML is being run by the browser. Ideally, I would want to deploy to, say a Heroku Review Branch, and then send that URL over to the W3C's Markup Service.

tysongach:
Anyone up for merging this? cc: @sneamonitakis @JoelQ @nickcharlton

nickcharlton:
I have no objections, I generally feel like this is a good practice to at least aspire to doing.

Assuming no one else any objections, could you rebase it to resolve the conflicts? (Unless you can't now, in which case I'll do it!)

tysongach:
@nickcharlton I don’t have write privileges anymore.

nickcharlton:
Ah yeah of course, your branch isn't from a fork. I rebased so it's in the new place (`html/README.md`).

PR #589, opened 2020-02-22
Move pseudo-element guides to Best Practices

tysongach:
This is not a guideline about code style. Rather, it's a syntax that is
part of the W3C spec. [According to MDN][mdn]:

> As a rule, double colons (`::`) should be used instead of a single
> colon (`:`). This distinguishes pseudo-classes from pseudo-elements.
> However, since this distinction was not present in older versions of
> the W3C spec, most browsers support both syntaxes for the original
> pseudo-elements.

Therefore, this guideline fits better in the Best Practices section
than in the Style section.

[mdn]: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#Syntax

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -265,9 +265,11 @@ CSS
   * Prefer `em` units instead of `px` for breakpoint values
   * Start with the smallest viewport size and work upwards using 
     `min-width`/`min-height`
+* Use [double colon syntax][pseudo-element-syntax] for pseudo-elements
 
 [autoprefixer]: https://github.com/postcss/autoprefixer
 [breakpoints]: http://bradfrost.com/blog/post/7-habits-of-highly-effective-media-queries/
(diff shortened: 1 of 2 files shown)


best-practices/README.md
```diff
@@ -265,6 +265,7 @@ CSS
   * Prefer `em` units instead of `px` for breakpoint values
   * Start with the smallest viewport size and work upwards using 
     `min-width`/`min-height`
+* Use double colons for pseudo-elements
```
JoelQ: Can we link the W3 spec here?

JoelQ: I meant the MDN doc you quoted 😆

tysongach: Good call! Added in 05c913546eca948fc2df68e9b80d88038de53ca4.

PR #597, opened 2020-03-31
Start documenting our core tech stack

jferris:
We've been hashing this out in Basecamp, and have reached some
conclusions. This adds a place to document tech stacks and includes the
decisions we've made so far.

### DIFF
--- a/stack/README.md
+++ b/stack/README.md
@@ -1,0 +1,52 @@
+# thoughtbot Stack
+
+We have a standard technology stack that we use for each new project by default.
+This provides a proven base for rapid, quality development and ensures we have a
+large team of developers ready to jump on new projects using a known technology.
+It helps to avoid decision fatigue at the beginning of each project.
+
+We deviate from this stack when we find something new that we need to evaluate.
+This allows us to practice our value of [Continuous Improvement].
+
+We will also swap in alternate technology as necessary when our default stack is
+inappropriate for the task at hand. This allows us to practice our value of
+[Quality].
+
+[Continuous Improvement]: https://thoughtbot.com/purpose#continuous-improvement
+[Quality]: https://thoughtbot.com/purpose#quality
+
+## Layers
+
+Our stack is broken up into layers. Each layer depends on another layer. This
+allows us to swap out an entire layer when necessary without losing all the
+decisions made for other layers.
+
+### UI
+
+#### Web
+
+* Use server-rendered HTML when possible as a UI layer.
+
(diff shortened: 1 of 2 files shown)


stack/README.md
```diff
@@ -0,0 +1,53 @@
...
+allows us to swap out an entire layer when necessary without losing all the
+decisions made for other layers.
+
+### UI
+
+#### Web
+
+* Use server-rendered HTML when possible as a UI layer.
+
+### Android (Native)
```
mike-burns: @JadeByfield89 can we get your opinions on this? These are being proposed as our defaults.

The problem we're trying to solve is: we'd like to experiment with new technology and ideas, but in order to experiment we first must have a baseline that we are deviating from. This is an attempt to define our base.

JadeByfield89: This looks good, but I don't think WorkManager is a one-size-fits-all for all background tasks. From the docs

> The WorkManager API makes it easy to schedule deferrable, asynchronous tasks that are expected to run even if the app exits or device restarts.

Seems like most of our background tasks(network calls, disk I/O) don't fit this description. We want them to run when the app is in use, and often immediately. Perhaps worth mentioning [Coroutines](https://developer.android.com/kotlin/coroutines) as an option here. We had a good time using them with Room, which might also be worth mentioning as a recommended database layer.

Maybe it's not even worth getting too specific, and leave these choices up to individual teams and developers to implement? I like Kotlin, Gradle, and MVVM as base recommendations though.

mike-burns: Makes sense to me. Let's drop the WorkManager line.

This does leave the question: Kotlin coroutines are an experiment that we tried recently, and were happy with. But what was it an experiment _against_ -- what was the control? What would we have done if we didn't have coroutines?

JadeByfield89: Good question. It seems like it comes down to "How do you handle async work on Android", which feels pretty situational and even philosophical. Maybe we don't need to explicitly recommend any best way here?

After all, coroutines are just _one_ way to handle async stuff. Some folks love RxJava and always use that as their staple. I don't know which is best. Maybe we leave this out and just go with recommending Kotlin, Gradle, and MVVM as defaults.

jferris: We don't have to decide on an option now, but I'd like to come up with a default approach so people know what to expect. I want to have questions to things like: will everyone need to learn RxJava? Will we try to help clients with their async problems by introducing Rx? I'll drop this line for now, though.

PR #599, opened 2020-04-13
Document our Ruby library choices

jferris:
Pull my Suspenders!


composerinteralia:
> I think Suspenders is out of date with our current preferences, but that's a separate problem.

@JoelQ if you have anything specific in mind, could you open an issue on suspenders to capture that?

PR #602, opened 2020-05-22
Expand the stack to include client-side frameworks

jferris:
This documents the conclusions from our mobile and UI strategy sessions.
It also splits out native iOS and Android from the Core Stack,
maintaining them as specialties.

### DIFF
--- a/tech-stack/README.md
+++ b/tech-stack/README.md
@@ -15,46 +15,23 @@ inappropriate for the task at hand. This allows us to practice our value of
 [continuous improvement]: https://thoughtbot.com/purpose#continuous-improvement
 [quality]: https://thoughtbot.com/purpose#quality
 
-## Layers
+## Core Stack
 
-Our stack is broken up into layers. Each layer depends on another layer. This
-allows us to swap out an entire layer when necessary without losing all the
-decisions made for other layers.
+Most developers at thoughtbot learn our Core Stack. The stack is broken up into
+layers and each layer depends on another layer. This allows us to swap out an
+entire layer when necessary without losing all the decisions made for other
+layers. Developers will have varying levels of experience with each layer, but
+the gaps between layers are small enough that developers can learn a new layer
+while building applications.
 
 ### UI
 
-#### Web
-
 - Use server-rendered HTML when possible as a UI layer.
-
-### Android (Native)
-
-- Use Kotlin for writing native app code.
-- Use Gradle with Android Studio for building.
-- Use MVVM to model views.
-- Use Apollo when consuming a GraphQL API.
-
-### iOS (Native)
-
-- Use Swift for writing native app code.
-- Use Xcode and `xcodebuild` for building iOS apps.
-- Use `xcpretty` to format `xcodebuild` output.
-- Use Xcode automatic provisioning when possible.
-- Use UIKit with Storyboards and MVVM for creating UIs.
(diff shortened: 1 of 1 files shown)


stack/README.md
```diff
@@ -15,50 +15,28 @@ inappropriate for the task at hand. This allows us to practice our value of
...
-We recommend learning at least one of the following:
-
-* MapKit or Google Maps
-* Core Location
-* Core Bluetooth
-* PhotoKit
-* UserNotifications
+* Use React when building components with a client-side framework.
+* Use TypeScript when writing client-side code.
+* Avoid building single-page applications for the web.
```
JoelQ, line 32: 💯

stack/README.md
```diff
@@ -15,50 +15,28 @@ inappropriate for the task at hand. This allows us to practice our value of
...
-* Prefer standard UIKit components to custom views.
-
-We recommend learning at least one of the following:
-
-* MapKit or Google Maps
-* Core Location
-* Core Bluetooth
-* PhotoKit
-* UserNotifications
+* Use React when building components with a client-side framework.
```
JoelQ, line 30: Much as I like Elm, I agree that React + Typescript makes more sense as our default.

stack/README.md
```diff
@@ -15,50 +15,28 @@ inappropriate for the task at hand. This allows us to practice our value of
...
+* Use TypeScript when writing client-side code.
+* Avoid building single-page applications for the web.
+* When building a cross-platform mobile app that will be delivered via an app
+  store, use React Native.
 
 ### API
 
 * Use GraphQL as an API layer when connecting a mobile app to a web service.
+* Avoid using GraphQL when connecting from one service to another.
```
JoelQ, line 39: Hmmm I'm not sure what this means? Do we prefer GraphQL for APIs used by frontends but REST for APIs used by other backends?

mike-burns: +1 to removing this, or replacing it with a "Prefer" statement that we can bikeshed instead.

jferris: I removed it in a663ab7.

WilHall:
@jferris Thoughts on merging this once the `README.md` conflicts are resolved?

stack/README.md
```diff
@@ -15,46 +15,23 @@ inappropriate for the task at hand. This allows us to practice our value of
...
-allows us to swap out an entire layer when necessary without losing all the
-decisions made for other layers.
+Most developers at thoughtbot learn our Core Stack. The stack is broken up into
+layers and each layer depends on another layer. This allows us to swap out an
+entire layer when necessary without losing all the decisions made for other
+layers. Developers will have varying levels of experience with each layer, but
+the gaps between layers are small enough that developers can learn a new layer
+while building applications.
 
 ### UI
```
mike-burns, line 27: Should this be "Web UI" instead? I don't expect e.g. Android UI to use server-rendered HTML.

jferris: This is also attempting to say that an HTML-based web UI will be the default UI. If we don't need to build a mobile application, we won't.

This section and the server-rendered HTML guideline weren't added by this pull request, so I would open a new issue or pull request if you think it should change.

iwz:
This is nice. I'm curious about the "components" part. React is very clearly component-based. Is this guideline to differentiate between choosing React and web components? 

What about if not building with components? Where do other tools like Stimulus and Vue fit in?

PR #603, opened 2020-05-22
Add style guides for TypeScript, React, and Jest

WilHall:
This PR is a first attempt at creating style guides for TypeScript, React, and Jest. I added the three together (along with modifications to the JavaScript style guide) since they reference each other.

Tasks remaining:
 - [x] Reference https://github.com/thoughtbot/react-suspenders and move configurations there

General thoughts and questions:
- ~Given that our contribution guidelines for this repository state that `Decisions about which libraries to use should live in template projects such as Suspenders`, some of my included recommendations for libraries may be better suited for a similar React / TypeScript project~ https://github.com/thoughtbot/react-suspenders/pull/1
-  [style/testing](https://github.com/thoughtbot/guides/tree/master/style/testing) exists but it is specific to RSpec. In this PR, I added `style/testing-jest` but I'm not convinced that's the right naming convention. Should we consider having a generic testing guide as well as a style guide for each testing framework we use? https://github.com/thoughtbot/guides/issues/611
- Should we create a separate repository for the TypeScript ESLint configs? [thoughtbot/eslint-config](https://github.com/thoughtbot/eslint-config) exists but it is JavaScript-specific.
- ~Should the example `tsconfig.json` file live here, or would that also be better suited for a Suspenders-like project?~ https://github.com/thoughtbot/react-suspenders/pull/1/files#diff-32ab45b0a7cbf0aa8200c5670863c486


style/javascript/README.md
```diff
@@ -1,8 +1,9 @@
-JavaScript
-==========
+# JavaScript
 
 [Sample](sample.js)
 
+* Use [ESlint] and [Prettier] for auto-formatting and auto-fixing
```
jferris, line 5: Do ESLint and Prettier eliminate the needs for the other guidelines here? (trailing commas, `===`, etc)

style/react/README.md
```diff
@@ -0,0 +1,50 @@
+# React
```
jferris, line 1: I think maybe this should be in best-practices instead (it's not about formatting) but the different between those is fuzzy. I wonder if they should be merged at this point.

jferris:
This is great! Thank you for putting it together. It would be good to get some feedback from folks who have worked more on React/TypeScript projects.

style/testing-jest/README.md
```diff
@@ -0,0 +1,28 @@
...
+See also [Testing](../testing/README.md).
+
+* Use [eslint-plugin-jest] to enforce testing style
+* Use [testing-library/jest-dom] and [jest-community/jest-extended] for
+  supplemental expectation matchers
+* Use [React Testing Library] for testing [React](../react/README.md)
+  components
+* Use [React Hooks Testing Library] for testing [React Hooks]
+* Use [User Event] for simulating DOM events on React components under test
+* Use [Fishery] for building factories
```
stevehanson, line 12: ❤️

style/testing-jest/README.md
```diff
@@ -0,0 +1,28 @@
...
+* Prefer placing test suite files alongside source files (e.g. `Thing.tsx`
```
stevehanson, line 13: A pattern I've seen more often is to put tests in `./__tests__/Thing.test.tsx`. I like that pattern since it keeps tests a bit out of the way but also almost colocated.

style/react/README.md
```diff
@@ -0,0 +1,50 @@
...
+ * Use [Apollo] with [code generation] when working with [GraphQL]
+ * Use [React Router] for routing in SPAs
+ * Avoid deeply nested routing when using React Router
+ * Prefer [Function Components] over [Class Components]
+ * Prefer keeping a single component in each file
+ * Use `PascalCase` for component names and their file names
+ * Use [React Hooks]
+ * Use pre-built hooks when possible (e.g. [streamich/react-use])
+ * Use [custom hooks] to encapsulate stateful logic outside a component
+ * Avoid [Forwarding Refs]
```
stevehanson, line 16: Is this saying to prefer passing a ref as a custom prop like `inputRef` instead of using `forwardRef` or to avoid passing refs altogether?

style/typescript/.eslintrc.json
```diff
@@ -0,0 +1,48 @@
+{
+  "extends": ["@thoughtbot/eslint-config", "plugin:@typescript-eslint/recommended", "plugin:jest/recommended"],
```
stevehanson, line 2: I typically prefer to use [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) along with [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to turn off ESLint rules that overlap with Prettier. Curious to hear your thoughts. We're doing this in the [React Reference app](https://github.com/thoughtbot/react-reference-app/blob/master/.eslintrc.json#L8).

style/typescript/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "arrowParens": "always",
+  "printWidth": 120,
+  "proseWrap": "always",
+  "singleQuote": true,
+  "trailingComma": "all"
```
stevehanson, line 6: I've usually done "es5" for this option so it doesn't use trailing commas in function arguments. [Reference](https://prettier.io/docs/en/options.html#trailing-commas). Curious to hear your thoughts.

style/javascript/README.md
WilHall: My thought was that ESLint and Prettier enforce the style guide, but that it was still worth listing the things we care most about directly in the guide. Alternately we could treat the ESLint and Prettier configs as the definitive source, but I find them to be less digestable for someone coming to the style guides and wanting to quickly read over all the guidelines.

style/react/README.md
WilHall: I agree, almost everything that came to mind for React was a best=practices thing (since style is already covered by the TypeScript guide).

I would be in favor of merging style and best practices guides. It might be worth calling out style guidelines vs best practices where both exist, but it also allows for guidelines which might be in more of a gray area.

style/testing-jest/README.md
WilHall: I really like having the test files as siblings of the component they test, though that may be a result of how I'm used to searching for them.

One benefit of the `__tests__` folder is that it does reduce noise in a directory where there are a lot of components. But I have also found it makes it a little harder to, at a glance, see if a component has a test suite.

I'm also open to either way - curious to hear others' thoughts on this!

style/react/README.md
WilHall: My thought here is probably better summarized as `Avoid nesting Forward Refs`.

The guideline I'd like to include here is less about how you should pass a ref, and more around when their use might point to an overly complex design.

I have seen some cases where there are multiple levels of components forwarding a ref, and it becomes increasingly difficult to trace the ref from its assignment to its usage. Furthermore a ref forwarded multiple layers and used by some far-off component creates a coupling between the two components and also pollutes the component tree when debugging with the React dev tools.

These have just been my observations, curious to hear others' thoughts too.

style/testing-jest/README.md
```diff
@@ -0,0 +1,28 @@
+# Testing with Jest
+
+See also [Testing](../testing/README.md).
+
+* Use [eslint-plugin-jest] to enforce testing style
+* Use [testing-library/jest-dom] and [jest-community/jest-extended] for
+  supplemental expectation matchers
+* Use [React Testing Library] for testing [React](../react/README.md)
```
edwardloveall, line 8: 👍

style/typescript/.eslintrc.json
WilHall: I changed this configuration quite a bit in [the .eslintrc.json for react-suspenders](https://github.com/thoughtbot/react-suspenders/pull/1/files#diff-838f0d5f2e5e1cba5ea6e331d430bd94) - I'd love your thoughts on the file in that PR.

My plan is to delete the configs I added to this repo, and instead link into `react-suspenders`.

style/typescript/.prettierrc
WilHall: 👍

I made the change over in https://github.com/thoughtbot/react-suspenders/pull/1/commits/64d698b111c422390731c3528e899d67f34547bd

style/testing-jest/README.md
johnschoeman: I too prefer the tests as siblings! I find that the specs are better maintained when they arent hidden off in a folder.  It seems to elevate the tests to more of a first class concern, which imo they should be.

style/react/README.md
```diff
@@ -0,0 +1,50 @@
...
+ * Avoid [Higher-Order Components] and [recompose]
+ * Prefer the `children` prop over [render props]
+ * Prefer using [TypeScript prop interfaces] over [PropTypes]
+ * Prefer the [short syntax] when using [Fragments]
+ * Prefer [React Contexts] over [Redux]
```
johnschoeman, line 21: ➕

style/typescript/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "arrowParens": "always",
```
johnschoeman, line 2: I believe that prettier 2.0 makes this the default.

https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev

style/typescript/README.md
```diff
@@ -0,0 +1,57 @@
...
+- Avoid [Overloading Functions]
+- Prefer [async/await] over traditional promise syntax
+- Prefer [Optional Properties] in an interface rather than declaring the property type as `T | undefined`
+- Prefer explicitly defining interfaces over [Extending Interfaces]
+- Avoid the use of the [any] type
+- Avoid the [Non-null assertion operator]
+- Use the [Nullish coalescing operator]
+- Prefer [Type Guards] over [Type Assertions]
+- Prefer the `as`-syntax for [Type Assertions] over the angle-bracket syntax
+- Prefer [Union Types], [Lookup Types], [Mapped Types] and [Const Assertions] over [Enums]
```
johnschoeman, line 27: what do you think about including: `Avoid type casting`

WilHall: 1000% yes

style/typescript/.prettierrc
WilHall: Thanks for catching this 🙂 

I removed this over in https://github.com/thoughtbot/react-suspenders/pull/1/commits/b702b5ec3ea86675365cc0068ec2e84c94e80209

style/testing-jest/README.md
edwardloveall: I am of a third opinion that tests should not be alongside components at all and instead be in their own root level directory that mirrors the components directory. Basically how most Rails projects (along with [many](https://github.com/luckyframework/lucky) [other](https://github.com/lefticus/cpp_starter_project) [languages](https://github.com/rust-lang/rustfmt) and [frameworks](https://github.com/rappasoft/laravel-boilerplate)) treat tests.

Although the entire community seems to be more in favor of tests near components and we'd be against almost every project we're likely to work with. I also see John's point about tests being a first-class citizen and I agree, but I have not found them to be more or less maintained because they are near or far in the file system. That feels like it's been more of a cultural problem than a test placement one.

If these are the options, I would at least prefer they are out of the way a little in a sibling folder. I think it makes them easier to work with which could decrease the friction of maintain them.

style/typescript/README.md
```diff
@@ -0,0 +1,58 @@
...
+- Prefer [async/await] over traditional promise syntax
+- Prefer [Optional Properties] in an interface rather than declaring the property type as `T | undefined`
+- Prefer explicitly defining interfaces over [Extending Interfaces]
+- Avoid the use of the [any] type
+- Avoid the [Non-null assertion operator]
+- Avoid [Type Assertions]
+- Prefer the `as`-syntax for [Type Assertions] over the angle-bracket syntax
+- Use the [Nullish coalescing operator]
+- Prefer [Type Guards] over [Type Assertions]
+- Prefer [Union Types], [Lookup Types], [Mapped Types] and [Const Assertions] over [Enums]
```
edwardloveall, line 28: I was talking about this with a client the other day, but how do you feel about preferring functions like `map`, `reduce`, `filter` as opposed to `for` loops and methods like `filterNulls(...)`? I'm not exactly sure what that category of functions is called...

style/react/README.md
```diff
@@ -0,0 +1,50 @@
+# React
+
+ * Use React in [Strict Mode]
+ * Use React with [TypeScript](../typescript/README.md)
+ * Use [Jest](../testing-jest/README.md) for unit testing
+ * Use [styled-components] for CSS-in-JS
```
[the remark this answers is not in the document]
WilHall: I'm pretty ambivalent about styled components vs scss modules. The `Use` here would be _use this if you need css-in-js_.

It might be the case that we never want to encourage css-in-js, in which case I would be in favor of removing this 🙂 

Some of these specific technology recommendations are also moving to templates in https://github.com/thoughtbot/react-suspenders/pull/1

style/react/README.md
```diff
@@ -0,0 +1,50 @@
...
+ * Use [Apollo] with [code generation] when working with [GraphQL]
+ * Use [React Router] for routing in SPAs
```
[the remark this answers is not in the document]
WilHall: This is similar to the css-in-js discussion above. I didn't mean to imply that we _should_ build SPAs, just that if we are the routing solution of choice would be React Router.

But looking at this from the perspective of someone reading these guides for perhaps the first time without that context, it does feel like a recommendation towards SPAs as a default which shouldn't be our stance.

https://github.com/thoughtbot/react-suspenders/pull/1 handles the SPA case right now since it is based off `create-react-app`. Maybe that, too, should have an option for bootstrapping a non-SPA variant, perhaps using rails suspenders. Though I don't have the expertise using those two together 🙂

style/react/README.md
```diff
@@ -0,0 +1,50 @@
...
+ * Use pre-built hooks when possible (e.g. [streamich/react-use])
+ * Use [custom hooks] to encapsulate stateful logic outside a component
+ * Avoid [Forwarding Refs]
+ * Avoid [Higher-Order Components] and [recompose]
+ * Prefer the `children` prop over [render props]
+ * Prefer using [TypeScript prop interfaces] over [PropTypes]
+ * Prefer the [short syntax] when using [Fragments]
+ * Prefer [React Contexts] over [Redux]
+ * Prefer [Controlled Components] over [Uncontrolled Components]
+ * Avoid using indexes as the value for [keys]
```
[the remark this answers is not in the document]
WilHall: From the React docs (https://reactjs.org/docs/lists-and-keys.html#keys):

> When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort: <example>
> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

Since React uses the keys to identify DOM elements, it won't re-render a DOM element if the key does not change. If you had a list of three items and you added a fourth and removed one at the same time, indexes would not uniquely identify the elements, which may lead to the change not triggering a re-render.

style/react/README.md
```diff
@@ -0,0 +1,50 @@
...
+ * Use [React Hooks]
+ * Use pre-built hooks when possible (e.g. [streamich/react-use])
+ * Use [custom hooks] to encapsulate stateful logic outside a component
+ * Avoid [Forwarding Refs]
+ * Avoid [Higher-Order Components] and [recompose]
+ * Prefer the `children` prop over [render props]
+ * Prefer using [TypeScript prop interfaces] over [PropTypes]
+ * Prefer the [short syntax] when using [Fragments]
+ * Prefer [React Contexts] over [Redux]
+ * Prefer [Controlled Components] over [Uncontrolled Components]
```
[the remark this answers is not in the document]
WilHall: That is correct -- an uncontrolled component will allow the user to change its value directly, where a controlled component will always set and maintain the value that is passed as a prop to the element.

This was an attempt to point out that you almost always want a controlled component, because often you are driving a form element off component state: you have a state variable and you set it `onChange` of the element, and pass the current state as the `value` prop to the element.

An antipattern here would be to try and accomplish the same with an uncontrolled component: you have a state variable and you set it `onChange` of the element, but you **don't** pass the current state as the `value` prop to the element. This might coincidentally work if the form element doesn't get re-rendered, but breaks down once a re-render happens (because the element loses the current value).

But the more I think about it, this situation is super specific and probably not fit for a best-practice.

style/typescript/README.md
WilHall: STRONGLY AGREE. What about something like

> Prefer functional style (`map`/`filter`/`reduce`) over iteration (`for`/`while`/`.each`).

style/react/README.md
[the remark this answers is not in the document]
edwardloveall: Makes sense to me!

style/typescript/README.md
edwardloveall: Love it.

style/javascript/README.md
WilHall: Devil's advocate, @composerinteralia authored https://github.com/thoughtbot/guides/pull/606 today which removes the ruby guidelines in favor of just using and referencing [standard](https://github.com/testdouble/standard).

I like the simplicity of that, and not having to maintain guidelines in both a configuration file and a textual guide.

Curious what others think as well.

style/react/README.md
WilHall: I have opened an issue to discuss combining them: https://github.com/thoughtbot/guides/issues/610

style/testing-jest/README.md
WilHall: I have created https://github.com/thoughtbot/guides/issues/613 to continue this discussion

style/javascript/README.md
WilHall: Based on https://github.com/thoughtbot/guides/pull/606#issuecomment-643308916 and the discussion in https://github.com/thoughtbot/guides/issues/610 it looks like we might be trending towards removing guidelines that we can't automate and don't feel strongly about. Linking those from here so we can continue the conversation when we condense the guides.

PR #606, opened 2020-06-05
Replace Ruby style guide with standard

### LINKED ISSUE #605: Replace Ruby style guide with standard
As we move toward using [standard](https://github.com/testdouble/standard), we can probably replace most or all of our [Ruby style guide](https://github.com/thoughtbot/guides/tree/master/style/ruby) with "• Use standard"

composerinteralia:
closes #605

[standard](https://github.com/testdouble/standard) is opinionated about
style, and some of those opinions are in conflict with our current style
guide.

This commit adds standard to the style guide, and removes any points
that standard already covers. I also ran standard to fix our sample.rb
to match this change.

This commit also removes .rubocop.yml. This will affect projects
which point to this file (e.g. factory_bot). They will need to either
point to a specific commit where the file still exists, or upgrade to
start using standard.


style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
 ====
 
-[Sample](sample.rb)
-
-* Avoid conditional modifiers (lines that end with conditionals). [36491dbb9]
-* Avoid multiple assignments per line (`one, two = 1, 2`). [#109]
-* Avoid organizational comments (`# Validations`). [#63]
-* Avoid ternary operators (`boolean ? true : false`). Use multi-line `if`
-  instead to emphasize code branches. [36491dbb9]
-* Avoid explicit return statements. [#48]
```
mike-burns, line 11: Are we losing this in our move to standard?

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Don't use `self` explicitly anywhere except class methods (`def self.method`)
-  and assignments (`self.attribute =`). [#89]
-* Don't vertically align tokens on consecutive lines.
-* If you break up a chain of method invocations, keep each method invocation on
-  its own line. Place the `.` at the end of each line, except the last.
-* If you break up a hash, keep the elements on their own lines and closing curly
-  brace on its own line.
-* Indent continued lines two spaces.
-* Indent private methods equal to public methods.
-* Name variables created by a factory after the factory (`user_factory`
```
mike-burns, line 24: Some of the things in here are ways for us to avoid thinking about things. Is it reasonable to preserve them?

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-  creates `user`).
-* Prefer nested class and module definitions over the shorthand version
-  [Example][class definition example] [#332]
-* Prefer `detect` over `find`. [0d819844]
```
mike-burns, line 28: Does standard handle this for us?

style/ruby/README.md
composerinteralia: standard covers this via the Style/RedundantReturn rule

style/ruby/README.md
composerinteralia: standard has no preference on detect vs. find

style/ruby/README.md
composerinteralia: I am open to discussing specific things that we should keep if folks feel strongly. If folks to not feel strongly, I plan to remove them all.

For "Name variables created by a factory after the factory" specifically, I am OK with variables names not matching the name of the factory as long as the variable names are contextually meaningful.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-[Sample](sample.rb)
-
-* Avoid conditional modifiers (lines that end with conditionals). [36491dbb9]
-* Avoid multiple assignments per line (`one, two = 1, 2`). [#109]
-* Avoid organizational comments (`# Validations`). [#63]
-* Avoid ternary operators (`boolean ? true : false`). Use multi-line `if`
-  instead to emphasize code branches. [36491dbb9]
-* Avoid explicit return statements. [#48]
-* Avoid using semicolons. [#108]
-* Avoid bang (!) method names. Prefer descriptive names. [#122]
```
georgebrock, line 13: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
georgebrock: I don't feel particularly strongly about this rule, but I think it makes sense for a “use standard” commit to only remove the things that are enforced by or overridden by standard. If we want to also remove this rule, it should be in another commit to make the reasons clear.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* If you break up a chain of method invocations, keep each method invocation on
-  its own line. Place the `.` at the end of each line, except the last.
-* If you break up a hash, keep the elements on their own lines and closing curly
-  brace on its own line.
-* Indent continued lines two spaces.
-* Indent private methods equal to public methods.
-* Name variables created by a factory after the factory (`user_factory`
-  creates `user`).
-* Prefer nested class and module definitions over the shorthand version
-  [Example][class definition example] [#332]
```
georgebrock, line 27: Standard doesn't enforce this one, and I feel pretty strongly about keeping it: there are all kinds of weird constant lookup issues that we can run into when we don't do this.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Prefer `detect` over `find`. [0d819844]
-* Prefer `select` over `find_all`. [0d819844]
-* Prefer `map` over `collect`. [0d819844]
-* Prefer `reduce` over `inject`. [#237]
```
georgebrock, line 31: Standard doesn't enforce any of these alias preferences, so we should keep them.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Prefer double quotes for strings. [#174]
-* Prefer `&&` and `||` over `and` and `or`. [#153]
-* Prefer `!` over `not`. [#195]
-* Prefer `&:method_name` to `{ |item| item.method_name }` for simple method
-  calls. [#183]
```
georgebrock, line 36: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Prefer `if` over `unless`. [#254]
-* Use `_` for unused block parameters. [0d819844]
```
georgebrock, line 38: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Prefix unused variables or parameters with underscore (`_`). [#335]
```
georgebrock, line 39: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Suffix variables holding a factory with `_factory` (`user_factory`).
```
georgebrock, line 40: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Use 2 space indentation (no tabs).
-* Use a leading underscore when defining instance variables for memoization.
-  [#373]
```
georgebrock, line 43: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Use an empty line between methods.
-* Use spaces after commas, after colons and semicolons, around `{` and before
-  `}`.
-* Use `%{}` for single-line strings containing double-quotes that require 
-  interpolation. [36491dbb9]
```
georgebrock, line 48: Standard doesn't enforce this one, so we should keep it (standard does prefer `%()` to `%{}` though)

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Use `{...}` for single-line blocks. Use `do..end` for multi-line blocks.
-  [0d819844]
-* Use `?` suffix for predicate methods. [0d819844]
```
georgebrock, line 51: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Use `CamelCase` for classes and modules, `snake_case` for variables and
-  methods, `SCREAMING_SNAKE_CASE` for constants. [36491dbb9]
-* Use `def self.method`, not `def Class.method` or `class << self`. [40090e22]
```
georgebrock, line 54: Standard enforces the `def Class.method` part of this, but not the `class << self` part.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Use `def` with parentheses when there are arguments. [36491dbb9]
```
georgebrock, line 55: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Don't use spaces after required keyword arguments. [Example][required kwargs]
-  [#205]
-* Use `each`, not `for`, for iteration. [36491dbb9]
-* Use a trailing comma after each item in a multi-line list, including the last
-  item. [Example][trailing comma example] [#167]
-* Use heredocs for multi-line strings. [36491dbb9]
```
georgebrock, line 61: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Prefer `private` over `protected` for non-public `attr_reader`s,
-  `attr_writer`s, and `attr_accessor`s. [#417], [#190]
```
georgebrock, line 63: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Order class methods above instance methods. [#320]
```
georgebrock, line 64: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
```diff
@@ -1,93 +1,4 @@
...
-* Prefer method invocation over instance variables. [#331]
```
georgebrock, line 65: Standard doesn't enforce this one, so we should keep it.

style/ruby/README.md
composerinteralia: OK, I won't delete everything. 😢

composerinteralia:
Wow! Thanks for going through all of these. It is not something I was eager to do myself.

jferris:
For what it's worth, I don't feel as though any of these are worth keeping if they can't be automated. Anything that falls more in the category of a best practice can be moved to best practices. I'd rather drop the others.

I still agree with some of the ones that have been deleted, but they aren't important enough to me that I want to try and enforce them without automation. If there are any we particularly care about, we could try to get them into Standard.

composerinteralia:
> For what it's worth, I don't feel as though any of these are worth keeping if they can't be automated. Anything that falls more in the category of a best practice can be moved to best practices. I'd rather drop the others.

I agree with this, but I also agree with limiting the scope of this PR. I will look into opening separate PRs to move or remove the others.

pablobm:
I'm not 100% sure, but I think this PR may have brought unintended consequences.

If I understand correctly, Hound's configuration across thoughtbot projects uses the file `style/ruby/.rubocop.yml` that was deleted here. With that file missing, Hound is now applying Rubocop's default rules, which are neither like thoughtbot's old rules, nor like StandardRB's.

As a result, I'm seeing unexpected Hound comments over at [Administrate](https://github.com/thoughtbot/administrate).

If my assessment is correct, I guess the fix would be to tell Hound to use StandardRB? I don't know if it's possible to directly tell it to use it. Perhaps instead it can be directed to use these Rubocop config files https://github.com/testdouble/standard/tree/master/config?

composerinteralia:
@pablobm Thanks for the comment. We are moving away from Hound, but in the meantime it should probably still work as before.

I wonder if it is possible to update
https://github.com/thoughtbot/guides/blob/master/.hound.yml#L17 to be a url to [an old commit with the Rubocop config file](https://github.com/thoughtbot/guides/blob/098e6c1925542506177b9729b100c6cf21ad7999/style/ruby/.rubocop.yml) instead of a path. If not, we may need to bring back the Rubocop config for now.

composerinteralia:
I added back .rubocop.yml for now in https://github.com/thoughtbot/guides/commit/9ac4a389e01af5db97fd3528a1d2aaf156efbf74

PR #608, opened 2020-06-12
Remove factory naming style guidline

composerinteralia:
This guideline was introduced in
https://github.com/thoughtbot/guides/pull/607. While I agree that naming
factories in this way makes sense, I think it is fairly nitpicky. We
already have "Name variables, methods, and classes to reveal intent" in
our [naming style guidelines], which I think implicitly covers something
like this. I don't care what the name of a factory is as
long as it is contextually meaningful.

[naming style guidelines]: https://github.com/thoughtbot/guides/tree/master/style#naming

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -11,8 +11,6 @@ Ruby
 * Avoid ternary operators (`boolean ? true : false`). Use multi-line `if`
   instead to emphasize code branches. [36491dbb9]
 * Avoid bang (!) method names. Prefer descriptive names. [#122]
-* Name variables created by a factory after the factory (`user_factory`
-  creates `user`).
 * Prefer nested class and module definitions over the shorthand version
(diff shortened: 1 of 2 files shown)


jferris:
I think it was actually introduced way back in #18.

I agree it's somewhat nitpicky and I don't think it's worth enforcing since it can't be automated.

composerinteralia:
Oops, good catch. I think I copied the url from the wrong tab.

mike-burns:
This is in the category of "decisions that I don't want to make." I'm not interested in thinking about what to make my variable holding the factory object, so having it written down means I get to make one fewer decision.

If you all enjoy making small decisions like this, I suppose I can keep it in my personal notes.

composerinteralia:
A lot has move around since I first opened this, and I also don't feel particularly strongly about it, so I am going to close for now.

PR #609, opened 2020-06-12
Remove ruby alias style guidelines

composerinteralia:
I know we have had some back on forth on these (e.g. #169 and #237).
These preferences were fine when we could automate fixing them with
RuboCop. But as we move to standard I think we should buy fully into
[standard's opinion on these aliases][standard opinion]. If we don't
agree with that opinion, we should discuss the issue on standard instead
of in this guide.

[standard opinion]: https://github.com/testdouble/standard/pull/39#issuecomment-445465868

### DIFF
--- a/style/ruby/README.md
+++ b/style/ruby/README.md
@@ -15,10 +15,6 @@ Ruby
   creates `user`).
 * Prefer nested class and module definitions over the shorthand version
   [Example][class definition example] [#332]
-* Prefer `detect` over `find`. [0d819844]
-* Prefer `select` over `find_all`. [0d819844]
-* Prefer `map` over `collect`. [0d819844]
-* Prefer `reduce` over `inject`. [#237]
 * Prefer `&:method_name` to `{ |item| item.method_name }` for simple method
   calls. [#183]
 * Use `_` for unused block parameters. [0d819844]
@@ -42,7 +38,6 @@ Ruby
 [#122]: https://github.com/thoughtbot/guides/pull/122
 [#183]: https://github.com/thoughtbot/guides/pull/183
 [#190]: https://github.com/thoughtbot/guides/pull/190
-[#237]: https://github.com/thoughtbot/guides/pull/237
 [#320]: https://github.com/thoughtbot/guides/pull/320
 [#331]: https://github.com/thoughtbot/guides/pull/331
 [#332]: https://github.com/thoughtbot/guides/pull/332


composerinteralia:
@jferris as far as I know we haven't added standard to suspenders yet, but we have plans to do so.

Also, I should have been more clear that standard does not fix these aliases one way or the other. The proposal here is that since standard doesn't care one way or the other we shouldn't either.

mike-burns:
This is another place where I don't want to make a decision and instead want it made for me. I think of these four decisions as fundamental to my defense of us having a style guide at all: they allow us to focus on the work instead of whether to use "`#detect` vs `#find`. Changing these -- or thinking about them at all -- is against my interpretation of the guides.

WilHall:
If Standard doesn't have any opinion one way or another, but we do, then I think it's worth keeping these guidelines.

However, I'd question how much we care about something like this. I agree with @mike-burns that I don't want to think about it, but keeping guidelines like this sort of encourages thinking about and enforcing these sorts of rules in the first place.

The [issue](https://github.com/testdouble/standard/pull/39#issuecomment-445465868) @composerinteralia linked explains it best:

> Ruby and (especially) Rails has an utterly massive surface area of methods and aliases to those methods and whether that's a strength or weakness depends largely on one's context. (Been a while since I linked to this, heh.) But regardless of the merits, Rubyland is littered with humane interfaces and it feels like it would be antithetical to what Ruby is to try to consistent-ify that by effectively banning alias use. Even if we agreed that it was a good idea, we'd never find them all.

This was said in the context of Standard of course, but I think it's a good philosophy for out guides as well.

We can't possibly have an opinion on all method aliases, so why are _these specific aliases_ important enough to include?

mike-burns:
We only listed aliases as we knew about them. I'm sure more have come out since then, and we could update our list of preferences.

It's worth noting that this is a "prefers" not a "use" -- basically, unless you have a reason to do otherwise, go with this one.

jferris:
What about:

> When Ruby or a library provides a method with several names or aliases, consistently use one name for that method on each project.

mike-burns:
Is there a place where I would want to use `#collect` and a place where I would want to use `#map`?

If I'm being consistent throughout a project then that would imply that there isn't. How about we just pick one so I don't have to search through the project every time?

composerinteralia:
I think we are thinking about this from two different perspectives, as an author and as a reviewer.

As an author I agree with Mike that having this guide frees me from having to choose between one name or the other.

As a reviewer I don't want this guide. Hound used to comment when this rule was violated, but now having this guide makes me think that it is a human reviewer's responsibility to make that comment.

composerinteralia:
A lot has move around since I first opened this, and I also don't feel particularly strongly about it, so I am going to close for now.

Issue #610, opened 2020-06-12
Combine Best Practices and Style guides


WilHall:
It may make sense to combine Best Practices and style guides.

My thoughts behind this:
 - There is some overlap between the existing best practices and style guides; it still might be worth calling each out separately if they live together, but combining them makes it feel more natural to include guidelines which exist in a gray area between "best practices" and "style"
 - We [split the style guides up by language/framework](https://github.com/thoughtbot/guides/commit/e5532a17a52d6c183ef9b13038db119c3d010ab6) but the best practices guides still live in [a single file](https://github.com/thoughtbot/guides/blob/master/best-practices/README.md). Combining them rectifies that difference in structure
 - It's more cumbersome to browse and maintain guides that exist in two places, but have similar content. I find myself checking both because I don't know which has the guideline I am looking for, or if adding a style guideline will overshadow something in best practices
 - The [graphql](https://github.com/thoughtbot/guides/tree/master/graphql) and [data](https://github.com/thoughtbot/guides/tree/master/data) guides are at the root level of the repo; they don't really fit into style or best-practices since they contain both. They would feel more at home in a structure that combined style and best-practices guides together

I propose a structure similar to that of the existing `style` guides folder, but named `best-practices`. I see "Style" as being in the superset of "Best Practices", and best practices can apply to programming languages, frameworks, and tooling. The proposed structure might look something like:
```
best-practices/
  css/
    README.md
    example.css
  html/
    README.md
    example.html
  ruby/
    . . .
  rails/
    . . .
  git/
    . . .
```

I would love to hear others' thoughts on this proposed structure, alternatives, or reasons against combining these two types of guides.

composerinteralia:
This seems to fit with https://github.com/thoughtbot/guides/pull/606#issuecomment-643308916, and some PRs I opened to start removing things from the Ruby style guide (#607, #608, #609).

jferris:
We could also potentially move them all to the root, since we also have some things like "Protocol" that are hard to differentiate from "Best Practices." The guides originally dealt largely with style concerns and only contained Ruby. They've evolved to the point now where I think a flat structure where everything is assumed to be best practices or recommendations would work best.

WilHall:
> They've evolved to the point now where I think a flat structure where everything is assumed to be best practices or recommendations would work best.

I like this option a lot.

Tangentially related to this issue, I think that suggestion fits well with [your comment](https://github.com/thoughtbot/guides/pull/606#issuecomment-643308916):

[...]

PR #612, opened 2020-06-12
Add language guidance

ericwbailey:
This PR adds a Terminology section to the Guide, reflecting conversation in [this Slack thread](https://thoughtbot.slack.com/archives/C024HFHU9/p1591968247462200).

While language is not technology related per-say, I think it is important to surface technical terms in a technical context. It is also anticipating what is being proposed in https://github.com/thoughtbot/guides/issues/610.

This is by no means a comprehensive list, nor am I the arbiter of what constitutes a problematic term. I feel we should be actively adding to this as we continue our Diversity and Inclusion initiatives.


terminology/README.md
```diff
@@ -0,0 +1,21 @@
+# Terminology
+
+The language we use has power. Being as concise and descriptive as possible helps to lessen ambiguity. This leads to shared understanding and better outcomes.
+
+## Terms
+
+* Use `blocklist`, and not `blacklist`.
+* Use `allowlist`, and not `whitelist`.
+* Avoid using `master` and `slave`. 
+	* For GitHub, use a term such as `main`, or `publish`, or `trunk`.
```
WilHall, line 10:
```suggestion
	* For git, use a term such as `main`, or `publish`, or `trunk`.
```

terminology/README.md
```diff
@@ -0,0 +1,21 @@
...
+	* For other scenarios, use terms such as `parent` and `child` or `leader` and `follower`.
+* Use terms such as `excused` or `exempted` in place of `grandfathered`.
+* Use `negative space`, and not `white space`.
+* Use `billable hours`, and not `man hours`.
+
+[google-io-video]: https://www.youtube.com/watch?v=mLjxXPHuIJo
+[web-dev-fast]: https://web.dev/fast
+[lighthouse-budgets]: https://bitsofco.de/your-first-performance-budget-with-lighthouse/
+[css-wizardry]: https://csswizardry.com/archive/
+[calibre-impact]: https://calibreapp.com/blog/visualise-performance-impact-between-deploys
```
WilHall, line 21: Should these links be removed?

terminology/README.md
samkap, line 10:
```suggestion
	* For GitHub, use a term such as `main`, or `publish`, `production`, or `trunk`.
```

terminology/README.md
ericwbailey: This is what happens when you think you hit push, but don't :sweat\_smile:

terminology/README.md
```diff
@@ -0,0 +1,15 @@
+# Terminology
+
+The language we use has power. Being as concise and descriptive as possible helps to lessen ambiguity. This leads to shared understanding and better outcomes.
+
+## Terms
+
+* Use `blocklist`, and not `blacklist`.
+* Use `allowlist`, and not `whitelist`.
+* Avoid using `master` and `slave`. 
+	* For GitHub, use a term such as `main`, or `publish`, `production`, or `trunk`.
```
composerinteralia, line 10: Some projects use other services for their git remotes, but this applies to git in general.

```suggestion
	* For git, use a term such as `main`, or `publish`, `production`, or `trunk`.
```

composerinteralia: Oops, I see this comment was already made. Don't mind me.

mike-burns, line 10: Can we pick one? I've been using `main` in my personal projects, and much of the discussion online has been around `main` -- can we agree on that one?

terminology/README.md
```diff
@@ -0,0 +1,15 @@
...
+	* For databases, use `write database` and `read replica`.
+	* For other scenarios, use terms such as `parent` and `child` or `leader` and `follower`.
```
mike-burns, line 12: I think `parent` and `child` are great examples but I do want to temper them with: be careful what verbs you apply to those. In POSIX we're stuck sending `kill` signals to `children` and it's awkward every time.

terminology/README.md
ericwbailey: Totally. `main` works for me!

terminology/README.md
ericwbailey: Do you think it's worth including language about that here? Something like:

> For other scenarios, use terms such as `parent` and `child` or `leader` and `follower`. Be mindful of scenarios where other commands will create language like, sending `kill` signals to `children`.

mike-burns: Yes please, and that is perfect!

terminology/README.md
```diff
@@ -0,0 +1,15 @@
...
+The language we use has power. Being as concise and descriptive as possible helps to lessen ambiguity. This leads to shared understanding and better outcomes.
+
+## Terms
+
+* Use `blocklist`, and not `blacklist`.
+* Use `allowlist`, and not `whitelist`.
+* Avoid using `master` and `slave`. 
+	* For git, use `main`.
+	* For databases, use `write database` and `read replica`.
+	For other scenarios, use terms such as `parent` and `child` or `leader` and `follower`. Be mindful of scenarios where other commands will create language like, sending `kill` signals to `children`.
```
nickcharlton, line 12: Should this line have another `*` before it?

ericwbailey: Ah, good catch, thank you. Updated.

terminology/README.md
```diff
@@ -0,0 +1,15 @@
+# Terminology
+
+The language we use has power. Being as concise and descriptive as possible helps to lessen ambiguity. This leads to shared understanding and better outcomes.
```
JoelQ, line 3: This purpose statement is about using more precise technical terms. However, most of the items below are about terms that have problematic racial undertones in a US context. Thoughts on highlighting that here? Otherwise we'll likely end up having discussions about which of two terms is most precise when that's not really the point.

I could see us wanting both types of terminology guidelines here. If so, thoughts on having separate "Non-inclusive terms" and "Ambiguous terms" sections below (we can come up with better header names)? Alternatively, maybe keep both under the same "Terms" section, but add a phrase in parens afterwards explaining why we like one term over the other?

For terms that are merely ambiguous, thoughts on using the weaker verb "Prefer" rather than "Use"?

terminology/README.md
```diff
@@ -0,0 +1,15 @@
...
+## Terms
+
+* Use `blocklist`, and not `blacklist`.
+* Use `allowlist`, and not `whitelist`.
+* Avoid using `master` and `slave`. 
+	* For git, use `main`.
+	* For databases, use `write database` and `read replica`.
+	* For other scenarios, use terms such as `parent` and `child` or `leader` and `follower`. Be mindful of scenarios where other commands will create language like, sending `kill` signals to `children`.
+* Use terms such as `excused` or `exempted` in place of `grandfathered`.
+* Use `negative space`, and not `white space`.
```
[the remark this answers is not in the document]
ericwbailey: It is, at least in a design context.

terminology/README.md
ericwbailey: Thematically, I'm not to keen on separating the terminology into "non-inclusive" and "ambiguous" sections, as that does seem counter to the intersectional goals of inclusivity.

I like your idea about adding language specifically as to why. I'll see what I can whip up.

As for ambiguous terms, I'd like to stay with "Use." I view this as a situation where an employee may arrive here looking for guidance, and this is our collective decision to choose one canonical term. Prefer opens up the opportunity for alternate terms, which potentially dilutes shared understanding.

WilHall:
@ericwbailey What are your thoughts on merging this?

Issue #611, opened 2020-06-12
Our testing guides are ruby/rails-centric


WilHall:
We have a guide on [testing best-practices](https://github.com/thoughtbot/guides/tree/master/best-practices#testing) and [testing style](https://github.com/thoughtbot/guides/tree/master/style/testing), but both contain guidelines specific to ruby/rails.

I noticed this in #603 when adding a [style guide for jest](https://github.com/thoughtbot/guides/pull/603/files#diff-6506acf0845c90847fba11ea15ff40de).

Some of the existing guidelines apply to testing across the board, and could live in a general guide on testing. For guidelines which might be language or framework specific (ruby vs. javascript, rspec vs. jest), where should those live?

Some thoughts on possible solutions:
1.  Promote our general testing guidelines to a top-level folder in this repository, and place language and framework specific guides (e.g. `testing-rspec`, `testing-jest`) into either `best-practices/` or `style/`
1. Add a section for each language/framework to the existing testing guides, and move the non-general guidelines into those sections
1. Only have specific guides -  rename the existing `testing` guide to be specific (`testing-rspec`? `testing-rails`?) and create new `testing-X` guides for other languages and frameworks

JoelQ:
I like the idea of a high-level, framework-agnostic testing guide

edwardloveall:
Agreed! Framework-agnostic guides in general are important to me. Another way to do it would be to make a `testing` directory which has a `README.md` with general guidelines and `rspec.md` and `jest.md` inside of it. The `README.md` could point to the other guides in the folder.

PR #615, opened 2020-06-26
Combine style, best practice, and protocol guides

WilHall:
This change addresses #610.

We remove the arbitrary grouping of guides into "style", "best
practices", and "protocols".

Instead, all guides now live at the top-level of the repository. This
makes it easier to find a guide, and reduces the chance of information
being duplicated between different types of guides for the same topic.

This change also updates the main `README.md` with a grouped table of
contents of all the guides.


README.md
```diff
@@ -1,44 +1,88 @@
...
-  * [React](./protocol/react)
-* [Security](./security)
-* [Style](./style)
-* [Web Performance](./web-performance)
-* [Working Together](./working-together)
-
-High level guidelines:
+Guides for working together, getting things done, programming well, and programming in style.
+
+### High level guidelines
```
ericwbailey, line 7: I like moving this here.

README.md
```diff
@@ -1,44 +1,88 @@
...
 
 * Be consistent.
 * Don't rewrite existing code to follow this guide.
 * Don't violate a guideline without a good reason.
 * A reason is good when you can convince a teammate.
 
-A note on the language:
+### A note on the language
```
ericwbailey, line 14: I also like breaking this out into its own section.

README.md
```diff
@@ -1,44 +1,88 @@
...
 * "Avoid" means don't do it unless you have good reason.
 * "Don't" means there's never a good reason.
 * "Prefer" indicates a better option and its alternative to watch out for.
 * "Use" is a positive instruction.
 
-Contributing
-------------
+## Guides by category
+
+[thoughtbot Tech Stack](/tech_stack.md)
```
ericwbailey, line 23: We'll need to add `-` to create bullets. The rich diff shows they're rendering as one long line of links.

README.md
ericwbailey: I also noticed we're going from a `h1` to a `h3`. Mind updating these to be `h2`s?

README.md
```diff
@@ -1,44 +1,88 @@
-Guides
-======
```
ericwbailey, line 2: I like that you're converting over to GitHub-flavored Markdown. I've had updating our Guides to match this as a thing I wanted to do for awhile now.

web/README.md
```diff
@@ -0,0 +1,7 @@
+Web
```
ericwbailey, line 1: Wow, this one is sparse. I should take a crack at adding some more 😬

README.md
WilHall: I didn't want to do it for all the files as part of this PR, but I want to do a follow-up "formatting only" PR to address this for all files 🙂

web/README.md
WilHall: Yeah -- there are a few like this. Once all the guides are at the root level and have consistent formatting, I want to audit them all for guides that need improvement.

README.md
```diff
@@ -1,44 +1,88 @@
...
 * "Avoid" means don't do it unless you have good reason.
 * "Don't" means there's never a good reason.
 * "Prefer" indicates a better option and its alternative to watch out for.
 * "Use" is a positive instruction.
 
-Contributing
-------------
+## Guides by category
+
+ * [thoughtbot Tech Stack](/tech_stack.md)
```
edwardloveall, line 23: Any reason why you moved `stack/README.md` → `tech_stack.md` but not other things like `/code-review/README.md`?

README.md
WilHall: Just created #617!

README.md
```diff
@@ -1,44 +1,88 @@
...
+ * [General](/general/README.md)
+
+### Collaboration
+
+ * [Code Review](/code-review/README.md)
```
edwardloveall, line 28: I believe that the only difference between paths like `/code-review/README.md` and `/code-review` is:

* Extra characters in the code
* The non-README.md links show the files around them (and also a navigate-one-level-up `..` link)

I guess I don't really mind one way or the other, but I don't understand the intention.

web/README.md
WilHall: Now we have #616 to track this 🙂

email/README.md
```diff
@@ -0,0 +1,12 @@
+Email
+-----
+
+* Use [SendGrid] or [Amazon SES] to deliver email in staging and production
+  environments.
+* Use a tool like [ActionMailer Preview] to look at each created or updated mailer view
```
edwardloveall, line 6: This is probably for another PR but this is a pretty Rails specific point. It could potentially go into a Rails guide either in the base README or in /rails/email.md

README.md
WilHall: I wasn't sure where to put the tech stack document. I put it at the root before I created `general/README.md`.

It probably makes sense to move it to `tech-stack/README.md` so it's like the rest.

README.md
WilHall: No specific intention when I did it. I actually like the behavior where you can see the surrounding files, so I am in favor of changing them to just the folder path.

email/README.md
WilHall: I'll try and catch things like this as part of #616 and create issues for specific changes.

rails/README.md
```diff
@@ -0,0 +1,208 @@
...
+[`.ruby-version`]: https://gist.github.com/fnichol/1912050
+[redirects]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.30
+[spring binstubs]: https://github.com/sstephenson/rbenv/wiki/Understanding-binstubs
+[prevent tampering]: http://blog.bigbinary.com/2013/03/19/cookies-on-rails.html
+[class constant in association]: /rails/sample.rb
+[order-associations]: /rails/sample.rb#L2-L4
+[validations]: /rails/sample.rb#L6
+[`app/views/application`]: http://railscasts.com/episodes/269-template-inheritance
+
+## Starting a new Rails app
```
edwardloveall, line 63: I wonder if these sub headers should be split up into their own files. Maybe not though since people are probably less likely to read the other files and more likely to read things in the same file? What do you think?

react/README.md
```diff
@@ -40,3 +40,22 @@
 [keys]: https://reactjs.org/docs/lists-and-keys.html#keys
 [component composition over component inheritance]: https://reactjs.org/docs/composition-vs-inheritance.html
 [react router]: https://reacttraining.com/react-router/
+
+## General Philosophies
+
+- For greenfield React projects we like to use TypeScript. TypeScript is a typed superset of JavaScript that compiles to
```
edwardloveall, line 46:
```suggestion
- For greenfield React projects we like to use [TypeScript](https://www.typescriptlang.org/). TypeScript is a typed superset of JavaScript that compiles to
```

react/README.md
```diff
@@ -40,3 +40,22 @@
...
+  [TypeScript in 5 minutes](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html).
+- If our app needs to make network requests and hold state outside of a component, we like using GraphQL and our
+  preferred library is [Apollo](https://www.apollographql.com). We have a [section on `GraphQL`](/graphql/) in our
+  Guides.
+- When building React apps with TypeScript and Apollo, we've found working in [VSCode](https://code.visualstudio.com/)
+  to be a mostly-good experience.
+
+## Create App
+
+Use [react-suspenders](https://github.com/thoughtbot/react-suspenders):
```
edwardloveall, line 57: This is still a private project. Did you mean to link it? Or maybe you meant to make react-suspenders public?

relational-databases/README.md
```diff
@@ -0,0 +1,13 @@
+
+Relational Databases
```
edwardloveall, line 2: This feels like it could be combined with postgres?

rails/README.md
WilHall: I think a good delineation could be guidelines vs. how-tos.

I feel like guidelines, even if they're split up by section, should live in the main `README.md` file for the guide.

But how-tos make sense as separate documents, with a `How To...` section in the main `README.md` linking to them.

For this guide, that might look like:

```md

## How To...

 - [Start a new Rails app](./starting_a_new_rails_app.md)
 - [Feature-test JavaScript in a Rails app](./feature_testing_rails_app_javascript.md)

```

Thoughts on that structure?

edwardloveall: I dig it!

shell/README.md
```diff
@@ -0,0 +1,51 @@
+Shell
+=====
```
edwardloveall, line 2: Old style?

relational-databases/README.md
WilHall: What are your thoughts on this when it comes to a topic where there are both specific and general guidelines?

Another good example that's similar to this is testing. We should have a `testing` guide, but we also need a `testing-rspec` and `testing-jest` guide since there are things specific to those frameworks that don't belong in the general guides.

In the case of relational databases, what would we do with the general guidelines if we added a specific `mysql` guide?

#611 goes into this a bit.

One solution besides having separate guides would be to have a single `testing` and `relational-databases` guide that has the framework guidelines as sections.

shell/README.md
WilHall: Yes! My plan is to update the style across the board in #617. I think the style changes will be easier to review in a standalone PR that doesn't move any files or change any content.

react/README.md
WilHall: It is still private, I should not link it.

relational-databases/README.md
edwardloveall: I'll comment there.

rails/README.md
WilHall: Cool 🙂 I split out the `ruby` and `rails` how-tos into sub-files and I also added a `_template` guide we can use as a reference of the expected structure (37b3b7f742ecbf49a06f97f2313c9efbff3c7e69)

composerinteralia:
@WilHall did you intentionally remove .rubocop.yml in this commit? I had removed it in https://github.com/thoughtbot/guides/commit/6d86958c9620d2f302cedf4c994232eae2323dff, but then added it back in https://github.com/thoughtbot/guides/commit/9ac4a389e01af5db97fd3528a1d2aaf156efbf74. It is causing hound to misbehave on some repos.

nickcharlton:
I just opened #623 to put this back, as it's breaking linting suggestions on Administrate and other projects.

PR #621, opened 2020-07-24
Update our Android best practices

mike-burns:
More to come, but these are the basics for Android development in the
year 2020.

### DIFF
--- a/android/README.md
+++ b/android/README.md
@@ -1,7 +1,11 @@
 # Android
 
-In addition to [java](/java/) best practices:
-
 - Properties of views should be alphabetized, with the exception of `id`,
   `layout_width`, and `layout_height` which should be placed first in that
   order.
+- Use Kotlin for all new code.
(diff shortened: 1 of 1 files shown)


mike-burns:
@JadeByfield89 thoughts? is there anything simple and non-controversial that we should add now?

JadeByfield89:
This is a solid list @mike-burns. Do we want to suggest that devs use a design pattern, or is that already implied?

Also, a couple suggestions

- Use string resources wherever possible
- Use vector drawables over PNGs or JPGs wherever possible

mike-burns:
Those two suggestions are perfect. I'll add them.

> Do we want to suggest that devs use a design pattern

I'll make a separate PR declaring that we should use MVVM for everything. There's a chance of disagreement and I'd like to contain that discussion there.

mike-burns:
MVVM: https://github.com/thoughtbot/guides/pull/626

PR #457, opened 2016-12-16
Convert jscsrc config to eslint

derekprior:
The JSCSRC project was absorbed into ESLint. I ran our configuration of
JSCSRC through polyjuice and to convert it.

I don't know our configuration or these tools well enough to know if
this is a complete transformation.

PR #622, opened 2020-07-30
Remove the deprecated jscs configuration

nickcharlton:
In #457, jscs was switched to eslint, but the configuration was still
left around.


nickcharlton:
I'm going to merge this as it's been a week and I think it's likely to be harmless, but it's very hard to tell!

PR #624, opened 2020-07-31
Fix typo in function components URL

JuanitoFatas:


nickcharlton:
@WilHall, I think we can just merge this one instead of waiting some time as it's a typo fix. (Happy to do that, if you don't want to!)

PR #626, opened 2020-07-31
Prefer MVVC on Android

mike-burns:
Google has been pushing MVVC lately and putting effort into their
`ViewModel` class. It's a convenient way to share backend logic between
multiple Fragments and Activities.


mike-burns:
@JadeByfield89 what do you think? Is there more to say here? Any nuance?

mike-burns:
Oh no!

mike-burns:
Fixed, thank you for catching that. Merging!

PR #630, opened 2020-11-20
Avoid returning null from GraphQL operations

mike-burns:
GraphQL mutations and queries can return `null`, but in practice this
only adds complexity to the client code. Prefer to return a rich data
type instead of a null value.

In graphql-ruby set [`null false`][] in your `BaseQuery` and `BaseMutation`.

[`null false`]: https://www.rubydoc.info/github/rmosolgo/graphql-ruby/GraphQL/Schema/Resolver#null-class_method

### DIFF
--- a/graphql/README.md
+++ b/graphql/README.md
@@ -62,5 +62,6 @@ be used.
 - Provide the ID of the deleted object as a field in mutations that delete
   objects.
 - Use JSON as a default transport format.
+- Avoid returning null from operations. [#630]
 
 [graphql specification]: https://graphql.github.io/graphql-spec/


sharplet:
I agree! When I read your explanatory PR description and then look at the diff, it feels like something is missing. Can we include more context? This might be a broader thoughtbot/guides style question.

mike-burns:
That might be a broader thoughtbot/guides question. What would you like to see in the diff?

sharplet:
I'd love to see a snippet of Ruby code showing what to do. My favourite guides include both examples and counterexamples to help illustrate each point.

mike-burns:
I'll do the compromise we've been doing, where I link to this PR from the guide.

PR #633, opened 2020-12-04
Introduce react native guide

johnschoeman:
Why:
We would like to establish some basic common guidance on best practices
in ReactNative.

This commit:
Introduces a guide for ReactNative.

Co-Authored-By: Devin Jameson <devin@thoughtbot.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ programming in style.
 - [iOS](/ios/)
 - [Rails](/rails/)
 - [React](/react/)
+- [React Native](/react-native/)
 - [Testing with Jest](/testing-jest/)
 - [Testing with RSpec](/testing-rspec/)
 
--- a/react-native/README.md
+++ b/react-native/README.md
@@ -1,0 +1,27 @@
+# React Native
+
+- Use React following the [React Guide](/react/)
+- Use [TypeScript](/typescript/)
+- Prefer using [core components and apis] over writing bespoke components.
+
+[core components and apis]: https://reactnative.dev/docs/components-and-apis
+
+## Style
+
+- Prefer using [StyleSheets]
+- Prefer [organized and composable styles]
+- Avoid designing for only one platform.
+- Avoid designing for only one screen size.
+- Prefer common mobile patterns and avoid [non-standard patterns].
+
+[StyleSheets]: https://reactnative.dev/docs/stylesheet
+[organized and composable styles]: https://thoughtbot.com/blog/structure-for-styling-in-react-native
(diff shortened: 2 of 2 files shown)


react-native/READEME.md
```diff
@@ -0,0 +1,26 @@
+# React Native
+
+- Use React following the [React Guide](/react/)
+- Use [TypeScript](/typescript/)
+- Prefer using [core components and apis] over writing bespoke components.
+- Prefer grouping logic by feature rather than implementation detail.
```
purinkle, line 6: Would you be able to expand on this?

react-native/READEME.md
```diff
@@ -0,0 +1,26 @@
...
+
+[core components and apis]: https://reactnative.dev/docs/components-and-apis
+
+## Style
+
+- Prefer [in-line and composable styles]
```
rakeshpetit, line 12: I think the term "in-line" is confusing when used in conjunction with styles. It might be mistaken for in-line styles which is not what we are advocating. Maybe we can reword this ? Also it is worth mentioning that Stylesheets are [performant](https://stackoverflow.com/questions/39336266/react-native-inline-styles-and-performance)

react-native/READEME.md
```diff
@@ -0,0 +1,26 @@
+# React Native
```
rakeshpetit, line 1: Typo in the file name - `READEME.md` :)

rakeshpetit:
I wrote [this](https://thoughtbot.com/blog/best-practices-while-developing-a-react-native-app) few months back and already find myself challenging some of the points. Maybe we can borrow some general points from this.

jferris (approved):
Thank you for putting this together!

react-native/READEME.md
johnschoeman: Yes, but not succinctly. I think it makes sense to remove this line for now and reintroduce this in its own PR after some word smithing.

react-native/READEME.md
johnschoeman: Agreed! What do you think about:
```
Prefer using StyleSheets
Prefer keeping styles colocated with components and composable
```

johnschoeman:
> I wrote [this post](https://thoughtbot.com/blog/best-practices-while-developing-a-react-native-app) few months back and already find myself challenging some of the points. Maybe we can borrow some general points from this.

I think we should! Probably would be best to introduce in a future PR and ship incrementally.

react-native/READEME.md
rakeshpetit: Sounds good to me.

rakeshpetit (approved):

PR #634, opened 2021-02-04
Document preference for ActiveStorage

joshuaclayton:
What?
=====

This documents our preference for ActiveStorage when available.

This also explicitly calls out avoiding interacting with storage
providers in a test environment.

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -112,6 +112,13 @@ others' Rails work, look in particular for:
   is a related todo in the project management system to add it during the
   staging and production deploys.
 
+## Asset Management
+
+- Use [ActiveStorage] to manage file uploads that live on ActiveRecord objects.
+- Don't use live storage backends like S3 or Azure in tests.
+
+[ActiveStorage]: https://guides.rubyonrails.org/active_storage_overview.html
+
 ## How to...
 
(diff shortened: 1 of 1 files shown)


rails/README.md
```diff
@@ -112,6 +112,13 @@ others' Rails work, look in particular for:
   is a related todo in the project management system to add it during the
   staging and production deploys.
 
+## Asset Management
+
+- Use [ActiveStorage] to manage file uploads that live on ActiveRecord objects.
+- Avoid using live storage backends like S3 or Azure in tests.
```
joshuaclayton, line 118: I could be swayed to use the stronger "Don't" language here but left it slightly more permissive given I'd trust if we did "need" to do this, there'd be reasons™.

aledustet: I don't think is that strong since one aspect that we also promote about writing tests is that they should not use the internet

aledustet (approved):

rails/README.md
JoelQ: I'm down for the stronger "Don't" 👍

JoelQ (approved):

mike-burns (approved):
Yup, and I'm also in favor of using "don't".

PR #635, opened 2021-02-08
Collect all Rails migration advice to one location

JoelQ:
Our Rails guide has a section for migration-related advice. However, our
big general advice section still contained several tips about
migrations. This commit moves them under the "migrations" heading to
keep the guide more consistent and easy to scan.


purinkle (approved):
:1234:

jferris (approved):

PR #636, opened 2021-02-08
Sort ActiveRecord queries by SQL execution order

JoelQ:
SQL does not execute the parts of a query in the order it is written.
Instead it follows the [following order]:

* `FROM`/`JOIN` and all the `ON` conditions
* `WHERE`
* `GROUP BY`
* `HAVING`
* `SELECT` (including window functions)
* `ORDER BY`
* `LIMIT`

The idea here is that we can better understand what's happening in a
query if we sort the chain of AR methods in the same order that they
will execute.

Additionally, this order is probably also from "most likely to want to
re-use" down to "most likely to want to break off". It's likely we might
want to have a base scope that does some joining and filtering, but that
other methods might want to control how those results are ordered or how
many get returned.

image

[following order]: https://jvns.ca/blog/2019/10/03/sql-queries-don-t-start-with-select

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -58,6 +58,15 @@
 - [Use blocks](/ruby/sample_2.rb#L10) when declaring date and time attributes in
   FactoryBot factories.
 - Use `touch: true` when declaring `belongs_to` relationships.
+- When building queries with the ActiveRecord interface, sort your method chain
+  as follows since this mirrors the [order SQL actually executes a query]:
+  - `from`
+  - `joins`
+  - `where`
+  - `group`
+  - `having`
+  - `order`
+  - `select`
 
 [add foreign key constraints]: http://robots.thoughtbot.com/referential-integrity-with-foreign-keys
 [`.ruby-version`]: https://gist.github.com/fnichol/1912050
@@ -65,6 +74,7 @@
 [spring binstubs]: https://github.com/sstephenson/rbenv/wiki/Understanding-binstubs
 [prevent tampering]: http://blog.bigbinary.com/2013/03/19/cookies-on-rails.html
 [`app/views/application`]: http://railscasts.com/episodes/269-template-inheritance
+[order SQL actually executes a query]: https://jvns.ca/blog/2019/10/03/sql-queries-don-t-start-with-select
 
 ## Migrations
 


JoelQ:
This PR is inspired by a conversation @edwardloveall and I had on a project where we implemented a lot of query objects. Is this a good idea? I'd love to hear people's thoughts

purinkle:
What impact does the order of chaining have on the SQL query constructed by Rails? Would this formatting be for readability, or does it have a performance impact?

purinkle:
If we merge this pull request, I'd love to read more in a blog post.

jferris:
I think it would be worth trying this as an experiment on a new project before adding it to the guides. I'm generally concerned about having ordering and formatting rules that can't be enforced automatically by tools like Rubocop. It means developers performing reviews will spend time looking at the order of these methods, which I believe is not the best use of a code review.

JoelQ:
@purinkle this has no impact on the generate query. It is purely for readability, and for the ability to decompose it by breaking pieces off the end rather than the middle.

purinkle:
I tried to use this format while writing the following ActiveRecord query yesterday.

```ruby
Order.where(some_identifier: nil).order(created_at: :asc).limit(1)
```

Unfortunately, this query returns a collection instead of an instance.

Ideally, I would have written the following to match the proposed format.

```ruby
Order.find_by(some_identifier: nil).order(created_at: :asc)
```

The above is incorrect syntax, and the correct way to write the query is as follows.


```ruby
Order.order(created_at: :asc).find_by(some_identifier: nil)
```

---

TL; DR — This format won't always work for all queries. We should always use the guides as _a guide_ and shouldn't be beholden to them.

jferris:
@purinkle Instead of `where(...).limit(1)` you can do `where(...).first`, which is another way in Ruby of adding a `LIMIT`.

However, if you start using abstractions like `find_by`, you've left the SQL abstraction layer, so it makes sense that you're not thinking so much about the query anymore.

purinkle:
Huh…

I'm sure `where(...).first` didn't include a `LIMIT` yesterday.

¯\\_(ツ)_/¯

mike-burns:
I like having a default so I don't have to think, but I've never had to think about the ordering of my ARel builders.

JoelQ:
This was an interesting idea but it looks like it solves a problem no one is having. I'm going to close this. If anyone has strong feelings feel free to re-open and champion this style.

PR #637, opened 2021-04-08
Rename some cop parameters in Ruby style guide

skalee:
Rename `ExcludedMethods` cop parameter to `IgnoredMethods`, as it has been changed in Rubocop 1.5.0.

Note: The old parameter name is still supported, though deprecated.

See https://github.com/rubocop/rubocop/pull/9098 for details.


nickcharlton:
Thanks!

PR #645, opened 2021-09-27
Add scopes DSL discussion link to Rails guide

cpjmcquillan:


JoelQ:
I'm curious why this was added. Most lines in our guide could probably have an issue or PR that led to them. Generally I think we try and capture that in the commit message rather than linking directly in the guide. Is this a new pattern we are looking to introduce?

mike-burns:
This follows precedent set in #473.

rails/README.md
```diff
@@ -15,7 +15,7 @@
 - Order i18n translations alphabetically by key name.
 - Order model contents: constants, macros, public methods, private methods.
 - Put application-wide partials in the [`app/views/application`] directory.
-- Use `def self.method`, not the `scope :method` DSL.
+- Use `def self.method`, not the `scope :method` DSL. [Discussion].
```
mike-burns, line 18: What do you think of

```suggestion
- Use `def self.method`, not the `scope :method` DSL. #643
```

As was done in #473.

PR #647, opened 2021-10-29
updating javascript guide

jimmythigpen:
- Added "Use TypeScript" (we already refer to this in the [React guide](https://github.com/thoughtbot/guides/tree/main/react))
- Added Formatting section with Prettier recommendations
- Updated links to [sample.js](https://github.com/thoughtbot/guides/blob/main/javascript/sample.js)

co-authored with Stephen Hanson

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -2,6 +2,7 @@
 
 [Sample](sample.js)
 
+- Use [TypeScript](/typescript/)
 - Use the latest stable JavaScript syntax with a transpiler, such as [babel].
 - Use [ESLint] and [Prettier] for auto-formatting and auto-fixing
 - Use [Jest] for unit testing
@@ -10,8 +11,6 @@
   (`null` or `undefined`).
 - Prefer [arrow functions] `=>`, over the `function` keyword except when
   defining classes or methods.
-- Use semicolons at the end of each statement.
-- Prefer single quotes.
 - Prefer ES6 [destructuring] over object literal notation.
 - Use ES6 [spread] and [rest] operator wherever possible for a cleaner code.
 - Use `PascalCase` for classes, `lowerCamelCase` for variables and functions,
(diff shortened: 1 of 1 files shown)


javascript/README.md
```diff
@@ -23,14 +22,24 @@
...
-- Include a `to_param` or `href` attribute when serializing ActiveRecord models,
-  and use that when constructing URLs client side, rather than the ID.
-- Prefer `data-*` attributes over `id` and `class` attributes when targeting
-  HTML elements. #462
-- Avoid targeting HTML elements using classes intended for styling purposes.
-  #462
+  
+## Formatting
+
+- Use [Prettier defaults](https://prettier.io/docs/en/options.html) with the following additional configuration (.prettierrc):
```
jferris, line 28: I'm curious - is there a strong reason not to just use to Prettier defaults as-is? Will the defaults result in consistent, decent-looking code that's most similar to community code?

jferris (approved):

javascript/README.md
stevehanson: I'm not sure. I think this PR tried to mostly preserve the "Prefer single quotes" rule that was already in place in this guide. I also question if double quotes are more in line with community code. According to [this issue thread](https://github.com/prettier/prettier/issues/1105), it seems most of the JS community uses single quotes (as of 2017), and Prettier started with the double quote default due to jlongster's preferences. According to that thread, the Prettier team wants to move toward a single quote default but the migration path would be difficult for their users.

[This article](https://bytearcher.com/articles/single-or-double-quotes-strings-javascript/) also seems to show that single quotes might be more popular than double quotes, and the React [documentation](https://reactjs.org/docs/components-and-props.html) and [source code](https://github.com/facebook/react/blob/main/packages/react/index.js) both use single quotes.

My preference would be to open a new issue or PR to propose changes to our single/double quote rule, since that wasn't the objective of this PR. There is also some past discussion about single vs. double quotes in the guides [here](https://github.com/thoughtbot/guides/pull/528) (2019) and [here](https://github.com/thoughtbot/guides/pull/336) (2015).

PR #650, opened 2021-12-17
Remote Meeting Guidelines word change

elisuh:
Word change because we cant actually ensure this, per @entcheva's suggestion on #649


entcheva (approved):
Much appreciated @elisuh!

heyvaleria (approved):

PR #652, opened 2022-04-22
Add guide for where to put service classes

sallyhall:
Distilled from a [slack conversation](https://thoughtbot.slack.com/archives/C02C73BAV/p1650307864464899) about where to put service classes

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -15,6 +15,8 @@
 - Order i18n translations alphabetically by key name.
 - Order model contents: constants, macros, public methods, private methods.
 - Put application-wide partials in the [`app/views/application`] directory.
+- Use `lib` for code that is not app-specific and could later be extracted into a gem.
+- Use `app/jobs` for code that doesn't need to return anything and can be run asynchronously.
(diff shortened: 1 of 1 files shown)


JoelQ:
Potentially relevant prior art:
* https://github.com/thoughtbot/guides/pull/578
* https://github.com/thoughtbot/guides/pull/402

rails/README.md
```diff
@@ -15,6 +15,9 @@
 - Order i18n translations alphabetically by key name.
 - Order model contents: constants, macros, public methods, private methods.
 - Put application-wide partials in the [`app/views/application`] directory.
+- Put classes (Service Objects/Interactors/et al) that may later be extracted
```
jferris, line 18: Thanks for distilling this! I have a few thoughts:

- Should this be two bullets? (One about `lib/` and one about `app/jobs`)
-  Maybe something like "Use `lb/` for code that will later be extracted to a gem" to describe `lib`? I think it could be anything generic, it's hard to make a finite list. My opinion is that `lib/` shouldn't have app-specific code.

nickcharlton (approved):

Newman1515 (approved):

PR #654, opened 2022-04-29
Update testing guides to reference system specs

cpjmcquillan:
Before, the guides referred to `feature` specs when outlining best
practices for testing with RSpec.

RSpec now uses [`system specs`][rspec-docs] by default, so this updates the
guides to refer to them.

Guidance around the Capybara `feature/scenario` DSL has been removed,
because it is only an [alias for `describe/it`][capybara-docs]. Removing
that specific guidance, reduces the cognitive overhead of remembering
which style to use when writing different types of RSpec test.

[rspec-docs]: https://relishapp.com/rspec/rspec-rails/docs/system-specs/system-spec
[capybara-docs]: https://github.com/teamcapybara/capybara/blob/master/README.md?plain=1#L222-L224

### DIFF
--- a/rails/how-to/feature_test_javascript_in_a_rails_app.md
+++ b/rails/how-to/feature_test_javascript_in_a_rails_app.md
@@ -21,7 +21,7 @@ capybara-webkit. For example, in `spec/system/user_signs_in_spec.rb`:
 
 ```ruby
 describe "Authentication", :js do
-  it "A user signing in" do
+  it "signs in a user" do
     create(:user, email: "me@example.com", password: "sekrit")
 
     sign_in_as email: "me@example.com", password: "sekrit"
--- a/rails/how-to/feature_test_javascript_in_a_rails_app.md
+++ b/rails/how-to/feature_test_javascript_in_a_rails_app.md
@@ -17,11 +17,11 @@ end
 ```
 
 When writing a spec, you must set the `:js` flag for that test to make use of
-capybara-webkit. For example, in `spec/features/user_signs_in_spec.rb`:
+capybara-webkit. For example, in `spec/system/user_signs_in_spec.rb`:
 
 ```ruby
-feature "Authentication", :js do
-  scenario "A user signing in" do
+describe "Authentication", :js do
+  it "A user signing in" do
     create(:user, email: "me@example.com", password: "sekrit")
 
     sign_in_as email: "me@example.com", password: "sekrit"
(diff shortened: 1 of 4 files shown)


testing-rspec/README.md
```diff
@@ -45,17 +45,16 @@
 
 [Sample](acceptance_test_spec.rb)
```
jared-thoughtbot, line 46: Is it worth calling out somewhere that `system` tests were previously known as `feature` tests in Rspec? I'm thinking about existing repos that will still have the `spec/features` directory and this may confuse some people when they can't find `spec/system`.

cpjmcquillan: Yeah, I think it is. I'm not sure where to put it though, any suggestions?

testing-rspec/README.md
```diff
@@ -45,17 +45,16 @@
...
-  for feature spec file names.
-- Use only one `feature` block per feature spec file.
-- Use scenario titles that describe the success and failure paths.
-- Use spec/features directory to store feature specs.
-- Use spec/support/features for support code related to feature specs.
+  for system spec file names.
+- Use only one `describe` block per system spec file.
+- Use `it` block descriptions that describe the success and failure paths.
+- Use spec/system directory to store system specs.
+- Use spec/support/system for support code related to system specs.
```
jared-thoughtbot, line 57: Maybe?

```suggestion
- Use spec/support/system for support code related to system specs.

> system specs were previously called feature specs and lived in `spec/features`
```

rails/how-to/feature_test_javascript_in_a_rails_app.md
```diff
@@ -17,11 +17,11 @@ end
...
 
 When writing a spec, you must set the `:js` flag for that test to make use of
-capybara-webkit. For example, in `spec/features/user_signs_in_spec.rb`:
+capybara-webkit. For example, in `spec/system/user_signs_in_spec.rb`:
 
 ```ruby
-feature "Authentication", :js do
-  scenario "A user signing in" do
+describe "Authentication", :js do
+  it "A user signing in" do
```
MatheusRich, line 24: I think if we want to use `it`, we should update the description as well "it a user signing it" doesn't read well. Some options:

```suggestion
  it "allows a user to sign in" do
```

```suggestion
  it "signs in a user" do
```

testing-rspec/README.md
JoelQ: My current client just finished upgrading to Rails 4 😅

JoelQ (approved):

PR #655, opened 2022-04-29
Add NIST link to communicating an OTP guidance

cpjmcquillan:
Before, we loosely explained that SMS is not a secure method for
communicating an OTP.

This updates the description to reference guidance from the National
Institute of Standards and Technology (NIST) in the US.


mike-burns (approved):
Heck yeah, thank you!

PR #660, opened 2022-04-30
Add link to related discussion in rails guide

dorianmariefr:


mike-burns (approved):
Huh. I thought GFM autolinked those. Not sure why I thought that.

PR #662, opened 2022-05-09
Acceptance Tests: Use accessible names and descriptions

seanpdoyle:
End-users very rarely interact with our applications based on CSS
selectors or an element's `[id]` attribute.

Whether they're driving their their browser directly or with the support
of assistive technologies, end-user interact with a page's form
controls, buttons, or links based on their visual presentation or their
text content.

In order for our automated test suites to reliably simulate end-user
experiences with the highest fidelity, we should write our instructions
with [accessible names and descriptions][names_and_descriptions] in
mind.

[names_and_descriptions]: https://www.w3.org/TR/wai-aria-practices/#names_and_descriptions

### DIFF
--- a/testing-rspec/README.md
+++ b/testing-rspec/README.md
@@ -45,6 +45,7 @@
 
 [Sample](acceptance_test_spec.rb)
 
+- Use the most specific [selectors][] available.
 - Don't locate elements with CSS selectors or `[id]` attributes.
 - Use [accessible names and descriptions][names_and_descriptions] to locate
   elements, to interact with form controls, buttons, and links, or to scope
@@ -62,6 +63,7 @@
 
 > system specs were previously called feature specs and lived in `spec/features`
 
+[selectors]: https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector
 [names_and_descriptions]: https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/
 
 ## Factories
--- a/accessibility/README.md
+++ b/accessibility/README.md
@@ -12,6 +12,8 @@ accessible.
 
 Automated checks can catch a lot of common issues before they reach production.
 
+- Test the application in a browser (like Capybara-driven [Acceptance
+  Tests](./testing-rspec/README.md#acceptance-tests))
 - Use tools such as [WAVE] or [axe's browser extensions] to run audits on your
   local build
 - Use a CI/CD solution such as [AccessLint] or [axe]
@@ -33,6 +35,7 @@ Automated checks can catch a lot of common issues before they reach production.
   for the background color they are placed over
 - Ensure all interactive content has distinct hover and focus states to help
   indicate interactivity
+- Ensure interactive elements [have a visible text label][rule-2]
 - Ensure color is not the only way to determine meaning
 - Ensure interactive components use common UI affordances where applicable, to
   help users understand how they can be operated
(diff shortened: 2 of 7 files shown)


cpjmcquillan (approved):

thiagoa (approved):
Thank you! I love this guideline, and the docs are very helpful.

testing-rspec/README.md
```diff
@@ -45,6 +45,10 @@
 
 [Sample](acceptance_test_spec.rb)
 
+- Don't locate elements with CSS selectors or `[id]` attributes.
```
mike-burns, line 48: > Don't

Most acceptance tests I see are testing the happy path, and therefore they should definitely follow this advice.

Sometimes I see acceptance tests that are like:

```ruby
scenario "the happy path", :js do
  # great practices in here. stellar work
end

scenario "the user's favorite color is incorrect", :js do
  # ... shortcuts that get us to the page most quickly ..., then
  fill_in "Favorite color", with: "Blue"
  click_on "That is definitely my favorite color. It's absolutely not yellow"
  expect(page).to have_content("You may not cross this bridge")
end
```

Where the happy path follows this advice, but the other test wishes it were a request test but it needs JS.

What do you think of this practice? Do you think it warrants using "avoid" instead of "don't"?

seanpdoyle: I think I'm missing context on the question. Where in this example code is an `[id]` or CSS selector?

mike-burns: Er, sorry about that. The `[id]` happens in here:

```
  # ... shortcuts that get us to the page most quickly ..., then
```

I was unimaginative and couldn't quickly find a real code example where we were sloppy in the lead-up to the error case code path.

(And the JS is whatever is required to get the content on the page. You can imagine a single-page app or some other monstrosity.)

Some meta thoughts:

- maybe an implication here is that error paths should be tested at a totally different layer.
- maybe there's no excuse for circumventing the good, accessible way to navigate, even if it is more verbose.

seanpdoyle: > maybe an implication here is that error paths should be tested at a totally different layer.

Personally, I'm exploring that space with efforts like [thoughtbot/action_dispatch-testing-integration-capybara](https://github.com/thoughtbot/action_dispatch-testing-integration-capybara), but this could be covered with view or component testing, separate from the browser environment.

> maybe there's no excuse for circumventing the good, accessible way to navigate, even if it is more verbose.

There are ways to check for Capybara to locate, interact with, or assert invalid fields. 

For example, the [`:field`](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector) selector accepts a `valid:` Boolean filter and a `validation_message:` String filter to affect query scope based on an element's [validity state](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation) (which requires a Real Browser to function, but doesn't require any JS).

Similarly, the [citizensadvice/capybara_accessible_selectors](https://github.com/citizensadvice/capybara_accessible_selectors) gem extends [`:fillable_field`](https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Selector) and other form control selectors to accept a [described_by:](https://github.com/citizensadvice/capybara_accessible_selectors#described_by-string) filter based on a form control having text content that references it through the [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation) attribute.

mike-burns (approved):
Yeah alright. Let's give this a try!

testing-rspec/README.md
```diff
@@ -58,6 +62,8 @@
 
 > system specs were previously called feature specs and lived in `spec/features`
 
+[names_and_descriptions]: https://www.w3.org/TR/wai-aria-practices/#names_and_descriptions
```
MatheusRich, line 65: It looks like the link has changed

```suggestion
[names_and_descriptions]: https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/
```

testing-rspec/README.md
RockSolid11212: Please let me know how I can help.  I believe I merged.

PR #664, opened 2022-05-10
Accessibility: Accessible names

seanpdoyle:
Whenever an element can receive focus or is otherwise interactive,
application authors are responsible for providing it with an [accessible
name][].

The [WAI ARIA Authoring Practices Guides][wai-aria] dedicates several
sections to outlining _how_ browsers and assistive technologies derive
an element's name. Instead of attempting to outline those methods with
the same level of detail as that document, this commit highlights (and
links to) the [5 Cardinal Rules of Naming][].

[accessible name]: https://www.w3.org/TR/wai-aria-practices/#names_and_descriptions
[wai-aria]: https://www.w3.org/TR/wai-aria-practices/
[5 Cardinal Rules of Naming]: https://www.w3.org/TR/wai-aria-practices/#naming_cardinal_rules

### DIFF
--- a/accessibility/README.md
+++ b/accessibility/README.md
@@ -33,6 +33,7 @@ Automated checks can catch a lot of common issues before they reach production.
   for the background color they are placed over
 - Ensure all interactive content has distinct hover and focus states to help
   indicate interactivity
+- Ensure interactive elements [have a visible text label][rule-2]
 - Ensure color is not the only way to determine meaning
 - Ensure interactive components use common UI affordances where applicable, to
   help users understand how they can be operated
--- a/accessibility/README.md
+++ b/accessibility/README.md
@@ -52,6 +52,13 @@ Automated checks can catch a lot of common issues before they reach production.
 
 ### Development
 
+- Ensure every focusable or interactive element has an [accessible name][]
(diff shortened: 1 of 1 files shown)


stevehanson (approved):
Thank you! That W3 doc was an informative read.

accessibility/README.md
```diff
@@ -52,6 +52,13 @@ Automated checks can catch a lot of common issues before they reach production.
 
 ### Development
 
+- Ensure every focusable or interactive element has an [accessible name][]
+- Follow the [Cardinal Rules of Naming][]:
+  1. [Heed Warnings and Test Thoroughly][rule-1]
+  2. [Prefer Visible Text][rule-2]
```
seanpdoyle, line 58: Would "Prefer Visible Text" also be a guideline worth including under this document's `Design` heading?

stevehanson: I think that would make sense. What would you think of something a little more self-explanatory, like:

> Ensure interactive elements [have a text label][rule-2]

seanpdoyle: Added!

cpjmcquillan (approved):

mike-burns (approved):

PR #667, opened 2022-05-31
Link Acceptance Tests in the Accessibility guides

seanpdoyle:
Link to the [Acceptance
Tests](./testing-rspec/README.md#acceptance-tests) section of the guides
in light of the Accessibility guideline's mention of automated testing
tools.

### DIFF
--- a/accessibility/README.md
+++ b/accessibility/README.md
@@ -12,6 +12,8 @@ accessible.
 
 Automated checks can catch a lot of common issues before they reach production.
 
+- Test the application in a browser (like Capybara-driven [Acceptance
+  Tests](./testing-rspec/README.md#acceptance-tests))
 - Use tools such as [WAVE] or [axe's browser extensions] to run audits on your
   local build
 - Use a CI/CD solution such as [AccessLint] or [axe]


accessibility/README.md
```diff
@@ -12,6 +12,8 @@ accessible.
 
 Automated checks can catch a lot of common issues before they reach production.
 
+- Test the application in a browser (like Capybara-driven [Acceptance
+  Tests](./testing-rspec/README.md#acceptance-tests))
```
seanpdoyle, line 16: The main motivation behind adding this guideline is to draw attention to the Acceptance Tests guidelines.

Those guidelines are currently RSpec and Capybara-specific, but the gist of the guidance is to aim to recreate the experiences of the end users. For mouse users, you should use tools like `click_on`. For keyboard users, you should have test coverage that navigates focus with the keyboard. For assistive tech users, your selectors and helpers should rely on the information that has semantic meaning (`[hidden]`, `[disabled]`, `:focus`, `[aria-selected="true"]`, etc).

I'm not sure about the language to use here to communicate that tersely without also treading the same territory as the Acceptance Tests guidelines.

mike-burns (approved):

jsilasbailey (approved):

PR #671, opened 2022-09-09
Add guidance about confidentiality w/r/t sharing others' pronouns

thoughtbot-summer:
Prompted by <https://github.com/thoughtbot/mkfriend/pull/103> / <https://github.com/thoughtbot/hub/pull/4945#issuecomment-1242469105>.


sallyhall (approved):
❤️

crystal-williams-brown (approved):
This touches on something I hadn't previously thought of but will keep in mind now. Thank you for this

working-together/README.md
```diff
@@ -125,5 +125,6 @@ While we don’t require sharing pronouns, the widespread practice of sharing th
...
 - When everyone in a meeting or group shares their pronouns every time, it can implicitly ‘out’ a person who is the only one not sharing their pronouns. If you’re comfortable, you can help by occasionally not sharing your pronouns, especially when the majority of the group is sharing theirs. This can normalize the idea that pronouns are optional and opting out doesn’t imply a specific identity.
 - If you’re asked to share your pronouns and you’d rather not, you do not need to provide an explanation. A simple “I’m not sharing my pronouns today” or “I’m fine with you addressing me by my name.” is enough.
+- If someone has shared their pronouns with you, think carefully before sharing their pronouns in other contexts, especially when such information might be publicly visible. That person may not be comfortable with other people knowing their pronouns, and/or it could put them in danger to be [outed](https://www.hrc.org/resources/glossary-of-terms) (for example, if their housing is dependent on a transphobic person who might evict them upon learning that they're trans).
```
mike-burns, line 128: Hi I am extremely [nitpicky](https://en.wikipedia.org/wiki/And/or#Criticism) about "and/or". Could we just say "or" instead? I don't think of the English language "or" a strictly "exclusive or" -- you'd need to add "either" to get that.

```suggestion
- If someone has shared their pronouns with you, think carefully before sharing their pronouns in other contexts, especially when such information might be publicly visible. That person may not be comfortable with other people knowing their pronouns, or it could put them in danger to be [outed](https://www.hrc.org/resources/glossary-of-terms) (for example, if their housing is dependent on a transphobic person who might evict them upon learning that they're trans).
```

mike-burns:
How do I put this in use in practice? When referring to someone, what pronouns do I use?

working-together/README.md
Joranhezon: Perhaps `it could also put them in danger...`? In this case I think it passes the definition that 'and/or' is trying to achieve.

Joranhezon:
@mike-burns I think in this scenario it would be good to reference them by name only.

Would like to hear other suggestions too tho.

Joranhezon (approved):
Thanks for this, @thoughtbot-summer!

working-together/README.md
louis-antonopoulos: Can you suggest some more specific guidance around 

>think carefully before sharing their pronouns in other contexts

For example, say Alex has shared with you that their pronouns are "they/them/theirs". 

I think the intent of this change is for people to not do things like send emails saying, "Hey everyone, make sure to use they/them/theirs when talking about Alex."

But what if you're in a client meeting and talking about Alex? Or sharing a story about Alex with a friend? Or talking about Alex on social media ("Alex and I had an awesome lunch together and they really inspired me to try a new menu item")? You have to use their pronouns in these instances so that you're honoring how they want to be referred to.

louis-antonopoulos:
>in this scenario it would be good to reference them by name only

I'm having trouble understanding the `this scenario` part [see above](https://github.com/thoughtbot/guides/pull/671#discussion_r978132374). Maybe it would help to list some scenarios in which you *would* use someone's shared pronouns, and what would be scenarios where you would consider *not* using them?

If a person wants to be referred to by specified pronouns, referring to that person `by name only` kinda feels like you're intentionally ignoring their pronouns by not using them.

I don't know. This is tricky. I understand the intent of the post. Maybe it is about "publicly sharing" vs. "publicly using"? Or "publicly [anything]"?

The biggest areas where clarification would help (on using someone's pronouns) is in work scenarios, such as:
- Client syncups / retro
- PRs
- Project management tools (Linear, Trello, JIRA)
- Client Slack channels / emails
- Third-party communication (e.g., working with the support team at Stripe)

Public-facing work stuff seems more likely that we'd want to exercise more caution, but what does that mean? Spelling  out what level of caution we're trying to achieve would be helpful.
- thoughtbot.com
- thoughtbot.com/blog
- The Bike Shed / Giant Robots / any other podcasts we have
- thoughtbot Twitter / any other

thoughtbot-summer:
@louis-antonopoulos @mike-burns Unfortunately it's a nebulous/nuanced issue, and there's no specific set of scenarios or rules that we could define for certain; it really depends on your relationship with the specific person in question. What if we append this guideline with something like:

> If possible, ask them in advance about how they would like to be referred to; if not, try to imagine what they would say based on what you know about them and their relationships with you, the people you're talking to, and the broader world.

Depending on the individual, this could look like:

- Explicitly stating their name and pronouns.
- Using their name and pronouns without explicitly stating them.
- Using they/them pronouns like you would for someone whose pronouns are unknown.
- Not referring to them by pronouns.
- Deadnaming them.
- Discussing events without revealing that they were involved.

---

> But what if you're in a client meeting and talking about Alex? Or sharing a story about Alex with a friend? Or talking about Alex on social media ("Alex and I had an awesome lunch together and they really inspired me to try a new menu item")? You have to use their pronouns in these instances so that you're honoring how they want to be referred to.

> If a person wants to be referred to by specified pronouns, referring to that person `by name only` kinda feels like you're intentionally ignoring their pronouns by not using them.

You're certainly not alone in this, and I think the reason many people feel this way is because they imagine it as a find-and-replace task, _i.e._ `s/pronoun/name/g`. So they might be about to say something like _"I assigned the task to **Alex**; **they're** really good at refactoring, so it was a great fit for **them**"_ then translate that to _"I assigned the task to **Alex**; **Alex** is really good at refactoring, so it was a great fit for **Alex**"_; however, if they put themself in the mindset of avoiding pronouns before they construct the sentence, then they might come up with something that flows more naturally, such as _"I assigned the task to **Alex**, who's really good at refactoring, so it was a great fit"_. That last version wouldn't feel out of place regardless of who you're talking about, so it wouldn't seem like you're intentionally omitting their pronouns.

louis-antonopoulos:
@thoughtbot-summer thank you so much for your response! It really helps me frame in my mind how we might achieve what you're going for. I'm heading out for the weekend (my brain is mush!) but I'll definitely come back to this next week and see how I can best contribute based on everything you just shared. 

P.S. I really loved the `s/pronoun/name/g` analogy.

mike-burns:
In what circumstances will a coworker have a different set of pronouns with us vs someone else we talk with during the workday?

thoughtbot-summer:
@mike-burns There is by nature no definitive answer to that or specific set of rules that you could apply, but if we replace "will" in your question with "might", then one example could be that someone feels comfortable using new pronouns with their team, a specific group, specific individuals, or only thoughtbotters (_i.e._ not clients). I'll add those examples to this guideline. :slightly_smiling_face:

mike-burns:
This feels hyperspecific at this point. Do we have coworkers or potential coworkers who want to use different pronouns per team?

I feel like if someone were in that position, they'd make it clear when sharing their pronouns. Do we need to document this?

thoughtbot-summer:
> Do we have coworkers or potential coworkers who want to use different pronouns per team?

I either don't know or am keeping that information confidential.

> I feel like if someone were in that position, they'd make it clear when sharing their pronouns.

Maybe, maybe not. Humans are complex.

> Do we need to document this?

We don't need to document any interpersonal objectives; nevertheless, it will certainly help many people. In this case, a specific scenario prompted this, and this new guideline aims to prevent that type of scenario from happening in the future. In addition, this helps fulfill the following [values](https://thoughtbot.com/purpose#values):

> ### Earn, impart, and summon Trust
> 
> […]
> 
> We do all of this while maintaining the privacy and confidentiality that every person deserves.
> 
> ### Practice Continuous Improvement
> 
> We recognize that we can always be better. We confidently go forward with the ideas we currently believe while remaining open and eager to better ideas. We take initiative to improve ourselves, the company, and our community.
> 
> This is a never-ending task, because there is always room for improvement. This week should be better than the last, and we should be optimistic that next week will be better than this week. We learn new things, and share those things with our peers and community.
> 
> […]
> 
> ### Uphold Quality
> 
> […] Beyond that, designing and building quality software means improving the security, privacy, and accessibility of the product; this requires improvements of life and reduction of harm for all users, contributors, and the people they affect.
> 
> […]
> 
> ### Seek Fulfillment in everything we do
> 
> We maintain an inclusive environment where we can thrive professionally and personally. […] We maximize our ability to take on any project by creating a diverse team who can bring their experience and perspectives together to solve problems. […]

sej3506:
> Do we need to document this?

Quote from the change:

> think carefully before sharing their pronouns

The wording of this addition reads to me like a "keep in mind" suggestion as opposed to a hard and fast rule, which I believe is the intention. It can feel scary to not want to make a coworker uncomfortable and put them in danger. It can also feel uncomfortable thinking you might violate a guideline/rule in the future. But we're all human and will all make mistakes; the best we can do is try to keep these things in our mind as important and act accordingly.

I think about our slackbot that reminds us not to use "guys." People will say it sometimes, with no ill intention, and will be gently corrected. The more we think about it, the more it becomes natural to avoid and to use more inclusive language.

I think a good indicator of if a person is no comfortable with their pronouns being shared outside of the context it was given is if they don't like their pronouns anywhere. Zoom username, slack profile, hub profile, etc. Obviously not always going to be true in either direction, but maybe a first place to check.

mike-burns:
> In this case, a specific scenario prompted this, and this new guideline aims to prevent that type of scenario from happening in the future.

Good reason, thanks for explaining.

PR #677, opened 2023-02-01
Removed IgnoredMethods

crnastena:
Removed `IgnoredMethods` parameter from `Metrics/BlockLength` cop for two reasons:

1) Its value is the same as the default `[]`. 
2) In newer versions of rubocop, IgnoredMethods is marked as deprecated (see [rubocop/rubocop#10829](https://github.com/rubocop/rubocop/pull/10829)) and AllowedMethods should be used. Since `AllowedMethods` default is also [] I think we don't need it in this guide.


matsales28:
Looks good to me. Thanks for the contribution!

PR #678, opened 2023-02-07
Add a suggesting change section to code review guidelines

matsales28:
The suggesting changes feature is becoming more popular, and having guidelines can make code reviews more efficient.

I'd appreciate any feedback on the organization structure and wording.

Co-authored-by: Stefanni Brasil <stefannibrasil@gmail.com>

### DIFF
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -65,6 +65,10 @@ experience, refactors the existing code). Then:
 - Seek to understand the author's perspective.
 - Sign off on the pull request with a 👍 or "Ready to merge" comment.
 - Remember that you are here to provide feedback, not to be a gatekeeper.
+- When suggesting changes:
+  - Communicate clearly which lines you suggest adding/removing.
+  - Test the suggested changes to validate it works whenever possible.
+  - When not possible, let the pull request author know that you did not test the suggestion.
 
 ## Style Comments
 
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -69,6 +69,7 @@ experience, refactors the existing code). Then:
   - Communicate clearly which lines you suggest adding/removing.
   - Test the suggested changes to validate it works whenever possible.
   - When not possible, let the pull request author know that you did not test the suggestion.
+  - Provide some context to let the author know why you're suggesting the change.
 
 ## Style Comments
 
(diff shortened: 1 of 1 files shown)


stefannibrasil (approved):
💯

code-review/README.md
```diff
@@ -65,6 +65,10 @@ experience, refactors the existing code). Then:
 - Seek to understand the author's perspective.
 - Sign off on the pull request with a 👍 or "Ready to merge" comment.
 - Remember that you are here to provide feedback, not to be a gatekeeper.
+- When suggesting changes:
```
jutonz, line 68: Is it correct that this advice pertains specifically to the "Add a suggestion" feature? If so, I think it might be helpful to say that explicitly. Otherwise I think people might interpret this as being for any type of suggestion, and not necessarily just this particular situation. What do you think?

```suggestion
- When suggesting changes using the "Add a suggestion" feature:
```

code-review/README.md
```diff
@@ -65,6 +65,10 @@ experience, refactors the existing code). Then:
...
+  - Communicate clearly which lines you suggest adding/removing.
+  - Test the suggested changes to validate it works whenever possible.
+  - When not possible, let the pull request author know that you did not test the suggestion.
```
jutonz, line 71: Nice! Maybe also we could say something about providing some context for a change instead of just suggesting it without explaining why it might be better. Would something like this work?

```suggestion
  - When not possible, let the pull request author know that you did not test the suggestion.
  - Provide some context to let the author know why you're suggesting the change.
```

jutonz (commented):
Nice, I love that this is being included. I've definitely seen inconsistent usage of this feature on client teams, so it'd be great to be able to link to these recommendations. I just had a couple of tiny thoughts. Thanks very much for writing this up :)

code-review/README.md
stefannibrasil: Totally agree, thank you @jutonz

code-review/README.md
stefannibrasil: Good point. At first I felt like it could be repetitive because I take providing context as a guideline already in this document.

However, I agree over-communicating is better, and maybe someone might land just on this section and not read the other guidelines, or simply not get to the same conclusion as I did 😉  Thank you for your suggestion!

jutonz (approved):
I love it. Great work!

neilvcarvalho (approved):
Thank you!

stevehanson (approved):
❤️

ayellapragada (approved):

PR #682, opened 2023-05-25
Add Ignite tech stack doc

adriantoddross:
This is a work in progress to create a guide that will help Ignite team members prescribe tech stacks to projects. I'm not sure if this guide will live here permanently, but it's a start!

### DIFF
--- a/ignite/README.MD
+++ b/ignite/README.MD
@@ -1,0 +1,49 @@
+# Ignite Tech Stack
+
+This doc is meant to help Ignite team members prescribe tech stacks for various projects.
+
+## Choosing a language
+
+...
+
+## Choosing a framework
+
+...
+
+## Choosing a database
+
+...
+
+### Where should backend code run?
+
+...
+
+## Choosing a host
+
+...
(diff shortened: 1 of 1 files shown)


mike-burns:
FWIW, we have a `developing/boost` directory in the Handbook.

I agree that there's not really a good place for team-specific docs like this.

It'd be good for them to live together, though, so please keep us updated on where this ends up.

pangolingo:
I think this _could_ live in the Guides, but it doesn't really fit with anything out there. We'd also have a greater burden of keeping it up-to-date if it's a public document.

I might opt for Mike's suggestion: put it alongside the Boost stuff in the handbook.
If we want to publicize it (and I think we do!) we could do it as a blog post. That would be nice because it's a snapshot of how we build things on Ignite in 2023. Future updates to the doc could result in future blog posts ("here's why we changed X").

adriantoddross:
This doc has been moved to the handbook: https://github.com/thoughtbot/handbook/pull/3308

PR #683, opened 2023-06-16
Update code review guidelines

samithoughtbot:
On my current client project we have been using labels as a prefix to comments we leave on code reviews to clearly display intention and tone. This removes a lot of ambiguity from written text which can be interpreted in many ways.

Our usage of labels has been informed by this article https://conventionalcomments.org/.

This PR would bring this optional way of working to other people within thoughtbot.

### DIFF
--- a/code-review/README.md
+++ b/code-review/README.md
@@ -32,7 +32,7 @@ that covers this material from [Derek Prior at RailsConf
 
 - Be grateful for the reviewer's suggestions. ("Good call. I'll make that
   change.")
-- Be aware that it can be [challenging to convey emotion and intention online]
+- Be aware that it can be [challenging to convey emotion and intention online], you may want to consider [using labels] to convey intention and tone.
 - Explain why the code exists. ("It's like that because of these reasons. Would
   it be more clear if I rename this class/file/method/variable?")
 - Extract some changes and refactoring into future tickets/stories.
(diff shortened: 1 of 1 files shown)


smaboshe:
This sounds interesting. I am not sure if I have misunderstood intent in a PR. I guess I assume positive intent. That said, I'd be open to giving this a try!

_Note to self_: This would be a good point to re-watch, "[Implementing a Strong Code-Review Culture](https://youtu.be/PJjmw9TRB7s)", by Derek Prior

nickcharlton:
I think it's a nice suggestion to have, but I'm not a fan of it myself.

I generally dislike things that make me sound like a robot when I'm writing and so if I had a series of little things (like the nitpick example), I like to write how I'm thinking at the time and that's often going to change. Not only that, but I like to think of reviewing code as much as a conversation as possible and if that means I have to write a lot more to express myself then I'm okay with that.

MatheusRich:
I haven't used this to convey tone, but I'm a fan of using `non-blocking/nitpick: comment`.

MatheusRich (approved):

purinkle (approved):
:bagel:

Love it! I would love it even more if we had a blog post to flesh out our thoughts and feelings about this change. Why is it a good idea? What problem does it solve? How have you seen it solve that problem?

heyvaleria:
> you may want to consider

I'm OK with this phrasing, as it gives everyone the option to use this system, or not.

I am also way more conversational and verbose in all types of communication, so I may embed the fact that it's a curiosity, non-blocking, or a nit, in the sentence, rather than the label.

Before we merge it into our guides, it could be good to hear a broader spectrum of opinions, but then again, you @samithoughtbot  shared about this PR in our slack channels and whatnot, to get reviews from folks.

See how verbose I am?

PR #684, opened 2023-06-18
Fix broken links in Ruby and Rails guides

lnvch:
Several links in Ruby and Rails guides are broken. Replacing them with working ones pointing to the same information.


smaboshe (approved):
Thanks for catching these and thank you Wayback Machine.

MatheusRich (approved):

PR #685, opened 2023-06-29
Remove guideline about using one factories.rb file per project

sarahraqueld:
From my experience working with projects using FactoryBot, this guideline is not something we follow. When asking the opinion of others at thoughtbot, they agreed they prefer to split up factories into their own individual files as this makes it easier to navigate and find them, as opposed to having to look for factories in a big file. 
What are your thoughts?


nickcharlton:
I feel like my opinion on this changes once it gets to a certain size. When it's easy to skim up and down, one file is really helpful. But once it gets over a certain size, it's more difficult to get a feeling of the factories when scanning it. The prior rule of keeping them alphabetically sorted helps this a lot, but is not, in my experience, followed.

How do people feel that we should split the file? By model? Or something else?

thoughtbot-summer (approved):

thoughtbot-summer:
> But once it gets over a certain size

I think that this is inevitable for most projects. I guess if you know your project will remain very small, keeping all of the factories in one file could make things more readable, but I've encountered very few projects like that.

> How do people feel that we should split the file? By model? Or something else?

I think generally one factory per file is good.

mike-burns (approved):
I approve of deleting guidelines in general, so LGTM.

littleforest (approved):

neilvcarvalho (approved):
I also prefer splitting `factories.rb` when it reaches a certain size. As each project will have its own needs, I like the idea of not specifying anything on the guides.

rdnewmanbot (approved):
I agree with this change

PR #686, opened 2023-09-01
Add explanation about what pair programming Ping pong is

dorianmarie:
It's writing test for each other to pass.

So the reader doesn't have to leave the guide and look it up


dorianmarie:
Superseded by https://github.com/thoughtbot/guides/pull/687

PR #689, opened 2023-09-06
Added Black - Python formatter

za:
so the code style is standardize and hopefully it will be more readable by most developers.


pangolingo (approved):
Good addition - Black is great.

PR #699, opened 2024-03-14
Call dynamic-readme reusable workflow

stefannibrasil:
We want to have a way to edit our README footer at one place and have the changes from there be propagated to our repos.

By adding this snippet in the README, we call this reusable workflow: https://github.com/thoughtbot/templates/blob/main/.github/workflows/dynamic-readme.yaml that renders and updates the README footer dynamically.

### DIFF
--- a/.github/worfklows/dynamic-readme.yml
+++ b/.github/worfklows/dynamic-readme.yml
@@ -1,0 +1,17 @@
+name: update-templates
+
+on: 
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  update-templates:
+    permissions:
+      contents: write
+      pull-requests: write
+      pages: write
+    uses: thoughtbot/templates/.github/workflows/dynamic-readme.yaml@main
+    secrets:
+      token: ${{ secrets.GITHUB_TOKEN }}
--- a/README.md
+++ b/README.md
@@ -98,13 +98,10 @@ waiting a week for everybody to leave feedback**.
 Thank you,
 [contributors](https://github.com/thoughtbot/guides/graphs/contributors)!
 
-![thoughtbot](https://thoughtbot.com/brand_assets/93:44.svg)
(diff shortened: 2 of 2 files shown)


matsales28 (approved):
LGTM! Do you know if those are the only necessary changes to implement the dynamic footer? If I do the same thing for `shoulda-matchers`, it'll work?

stefannibrasil:
> LGTM! Do you know if those are the only necessary changes to implement the dynamic footer?

Yes! The additional and final step is to add the repo to this workflow_dispatch line: https://github.com/thoughtbot/templates/blob/main/.github/workflows/trigger-dynamic-readme-update.yaml#L17 I was hoping that wasn't needed, that we could use `github.repository` but GitHub Actions seem to require the repos to be explicitly set there. This workflow dispatch is the one that will trigger the workflow being added in this PR to fetch any updates made to the templates footer file. Without it, when we change the template file, the repos calling the workflow won't be automatically updated.

I will document those when I'm finished. I'm keeping track of the repos to be added to that line to do it once instead of adding one by one. Just to recap: the workflow will render the README once this PR is merged -- we only need to add them to the workflow_dispatch one to keep the footer updated when the original template is changed.

> If I do the same thing for shoulda-matchers, it'll work?

Yep! I'm working on that now and will open one for shoulda-matchers after lunch. Stay tuned!

stefannibrasil:
I wrote a typo so it didn't trigger once it was merged but after fixing the typo, the workflow was ran. The README is now being rendered from the templates 💪 

https://github.com/thoughtbot/guides

matsales28:
@stefannibrasil, just a reminder to delete the directory with the typo `.github/worfklows/`.

stefannibrasil:
thank you! done in https://github.com/thoughtbot/guides/commit/f2d95793d95e502b26b1564b4142457ee9746108

PR #700, opened 2024-04-02
Add React Native tooling preferences to guides

stevehanson:
These are defaults that our mobile team has standardized on over the past few years but not documented here.

### DIFF
--- a/react-native/README.md
+++ b/react-native/README.md
@@ -6,6 +6,22 @@
 
 [core components and apis]: https://reactnative.dev/docs/components-and-apis
 
+## Tooling
+
+* Start new projects with [create-belt-app](https://www.npmjs.com/package/create-belt-app)
+* Use [Expo](https://expo.dev)
+* Use [Expo EAS](https://expo.dev/eas) for continuous deployment
+* Use [Expo Secure Store](https://docs.expo.dev/versions/latest/sdk/securestore/) for storing sensitive data like auth and refresh tokens
+* Use [React Navigation](https://reactnavigation.org/) for routing
+* Use [TanStack React Query](https://tanstack.com/query/v4/docs/framework/react/overview) as API client for REST APIs
+* Use [Apollo Client](https://www.apollographql.com/docs/react/) as API client for GraphQL APIs
+* Use [Redux Toolkit](https://redux-toolkit.js.org/) for global state
+  * Do not use global store for API state -- use API client library instead
+* Use [React Native Firebase](https://rnfirebase.io/) for push notifications
+* Use Sentry for error reporting
+* Prefer [RevenueCat](https://www.revenuecat.com/) for in-app payments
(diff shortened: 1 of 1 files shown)


react-native/README.md
```diff
@@ -6,6 +6,22 @@
...
+* Start new projects with [create-belt-app](https://www.npmjs.com/package/create-belt-app)
+* Use [Expo](https://expo.dev)
+* Use [Expo EAS](https://expo.dev/eas) for continuous deployment
+* Use [Expo Secure Store](https://docs.expo.dev/versions/latest/sdk/securestore/) for storing sensitive data like auth and refresh tokens
+* Use [React Navigation](https://reactnavigation.org/) for routing
+* Use [TanStack React Query](https://tanstack.com/query/v4/docs/framework/react/overview) as API client for REST APIs
+* Use [Apollo Client](https://www.apollographql.com/docs/react/) as API client for GraphQL APIs
+* Use [Redux Toolkit](https://redux-toolkit.js.org/) for global state
+  * Do not use global store for API state -- use API client library instead
+* Use [React Native Firebase](https://rnfirebase.io/) for push notifications
```
rakeshpetit, line 20: I like Firebase but did we discuss why not https://docs.expo.dev/push-notifications/overview/ instead which in turn uses Firebase Cloud Messaging instead of RN Firebase? Are they one and the same?

rakeshpetit (approved):
❤️

react-native/README.md
stevehanson: Interesting. That link uses Expo Notifications, which I believe is Expo's own push notification service ([FAQ here](https://docs.expo.dev/push-notifications/faq/)). And then I'm guessing it uses Firebase Cloud Messaging under the hood? I'm hesitant to tie push notifications directly to Expo and add that extra abstraction layer, personally, and prefer to connect straight to Firebase. Concerns with using Expo's service are:

* Is there lock-in with Expo's service? Eg. if Expo started charging for this service, how would we migrate to using Firebase directly?
* It looks like there's a rate limit
* An extra abstraction layer might make things more difficult to reason about and troubleshoot

On the other hand, Expo usually provides a great dev experience, and if it works well, I can see it being helpful in getting push notifications off the ground easily.

This [other expo doc](https://docs.expo.dev/guides/using-firebase/) also goes into two approaches for using Firebase directly in an Expo project: Firebase JS SDK and React Native Firebase.

My experience and the functionality that @codeofdiego is [building into Belt](https://github.com/thoughtbot/belt/pull/34) are around React Native Firebase. I haven't personally used the Firebase JS SDK or Expo Notifications, but I'm open to discussing more! What would we think about a separate issue or PR to discuss this?

rakeshpetit: Great explanation. I'm not a fan of multiple layers of abstraction as well but only wanted to use Expo as they will make integration a breeze. Your explanation and the integration work on belt convince me enough to directly use the service.

PR #701, opened 2024-04-05
Translations as a default for client projects

laicuRoot:
We would like to enable translations as a default for our clients projects.

For this, we are including guidelines in accessibility about translations.

Related to issue https://github.com/thoughtbot/handbook/issues/3751

### DIFF
--- a/accessibility/README.md
+++ b/accessibility/README.md
@@ -126,13 +126,6 @@ assign it one of four ratings:
 Use the Notes sheet to leave per-cell comments when necessary, referencing them
 with a link. The next steps for an audit are handled on a per-project basis.
 
-## Translations
-
-- Ensure that the application is setup to support multiple locales.
-- Ensure that the application raises an error when a translation is missing for a
-  given locale in development.
-
-
 [accessibility audit template]: https://www.fsb.org.uk/resources-page/accessibility-audit-template.html
 [accesslint]: https://github.com/marketplace/accesslint
 [axe]: https://www.deque.com/axe/axe-for-web/integrations/
--- a/accessibility/README.md
+++ b/accessibility/README.md
@@ -126,6 +126,13 @@ assign it one of four ratings:
 Use the Notes sheet to leave per-cell comments when necessary, referencing them
 with a link. The next steps for an audit are handled on a per-project basis.
 
+## Translations
+
+- Ensure that the application is setup to support multiple locales.
+- Ensure that the application raises an error when a translation is missing for a
+  given locale in development.
+
+
 [accessibility audit template]: https://www.fsb.org.uk/resources-page/accessibility-audit-template.html
 [accesslint]: https://github.com/marketplace/accesslint
 [axe]: https://www.deque.com/axe/axe-for-web/integrations/
--- a/rails/README.md
+++ b/rails/README.md
@@ -55,6 +55,12 @@
   FactoryBot factories.
 - Use `touch: true` when declaring `belongs_to` relationships.
 
+## Translations
+
+- Ensure that the application is setup to support multiple locales.
+- Ensure that the application raises an error when a translation is missing for a
+  given locale in development and tests.
+
 [`.ruby-version`]: https://gist.github.com/fnichol/1912050
 [redirects]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.30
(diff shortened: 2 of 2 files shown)


accessibility/README.md
```diff
@@ -126,6 +126,13 @@ assign it one of four ratings:
 Use the Notes sheet to leave per-cell comments when necessary, referencing them
 with a link. The next steps for an audit are handled on a per-project basis.
 
+## Translations
+
+- Ensure that the application is setup to support multiple locales.
+- Ensure that the application raises an error when a translation is missing for a
+  given locale in development.
+
+
```
mike-burns, line 135: Extra newline.

mike-burns (commented):
Huh, didn't realize we were missing this.

That said, I'd put this in the `rails/README.md`. Maybe I'm opposed to the accessibility section, though.

Something like:

> - Use i18n for all strings, including validation errors.

And the "raise an exception" bit would go in Suspenders (I think it's already there).

What do you think?

minaslater:
> That said, I'd put this in the `rails/README.md`. Maybe I'm opposed to the accessibility section, though.
> Something like:
> 
> > * Use i18n for all strings, including validation errors.

@laicuRoot and I went back and forth a little bit on where to include this. We landed on Accessibility to make this guideline somewhat language-agnostic, but I honestly can't say I've worked with translations in anything other than Rails. 🤔  I'm not opposed to moving it to `rails/README.md` if that makes more sense.
 
> And the "raise an exception" bit would go in Suspenders (I think it's already there).

I'm new to this repo and how it's used. Wondering if it's worth echoing here the default in Suspenders?

mike-burns:
I think for non-Rails code we'll want to provide more explanation around internationalization. For apps that are an API with a React frontend, for example, we might want to come down on how to set up i18n and have a "thoughtbot way" around it. Less so for Rails, since there's one way to do it.

---

This repo has a mysterious, vibe-oriented relationship with Suspenders. In general, if something can be Rails-specific and can be enforced via code, it goes in Suspenders instead of Guides.

---

All that said, I'm not opposed to keeping it as-is in this PR.

accessibility/README.md
```diff
@@ -126,6 +126,13 @@ assign it one of four ratings:
 Use the Notes sheet to leave per-cell comments when necessary, referencing them
 with a link. The next steps for an audit are handled on a per-project basis.
 
+## Translations
+
+- Ensure that the application is setup to support multiple locales.
+- Ensure that the application raises an error when a translation is missing for a
+  given locale in development.
```
MatheusRich, line 133: Should it raise on tests?

minaslater: I think [it does](https://github.com/thoughtbot/suspenders/blob/55cb5c246fe53aea4cf53436c7c208b40ad13e85/NEWS.md?plain=1#L465). We'll have to adjust the wording here.

pangolingo:
Ruby on Rails has a good feature set for translations, but not all frameworks do. The React ecosystem has some popular libraries but they're a little more cumbersome than Rails I18n.

I'd prefer to restrict this guideline to Rails until we ensure that it's feasible for every project.

laicuRoot:
Thank you all for your comments 🙂 

After reading your thoughts, I've thought that it might be better to include this change only in the Rails guide. I agree with @pangolingo in this case. 

That way, we could be more specific in our wording about raising an error if a translation is missing. 

What do you think?

minaslater:
> After reading your thoughts, I've thought that it might be better to include this change only in the Rails guide. I agree with @pangolingo in this case.
> 
> That way, we could be more specific in our wording about raising an error if a translation is missing.
> 
> What do you think?

@laicuRoot I agree!

nickcharlton:
> Ruby on Rails has a good feature set for translations, but not all frameworks do. The React ecosystem has some popular libraries but they're a little more cumbersome than Rails I18n.

I'd look at this a little differently and think of it as the standard we want to uphold (always using translation files) and to then put us in a position where we need to make a decision when we can't fulfil that standard.

That can then make us go, well, if the React libraries are generally not good, what can we do to help improve that so that we can do things the way we'd like to do them?

mike-burns:
When we learn how to use translations with React, we should document them here. I'm not sure the guides repo is the place for values and goals, though -- if so, I'd like it under a new section explicitly dedicated to that.

heyvaleria (approved):
Can we merge this? @laicuRoot

PR #704, opened 2024-04-05
Explain that vague meeting invitations can trigger anxiety attacks

thoughtbot-summer:


working-together/README.md
```diff
@@ -156,6 +156,11 @@ You can [download a double-sided printout] of these guides to display in your me
   - Let everyone know ahead of time what they can expect.
   - Create a calendar invite with links to the video call and dial-in information (if applicable)
   - Create a meeting agenda or note about what the meeting will cover.
+  - Strongly avoid vague meeting invitations. For people with anxiety, vague
```
nickcharlton, line 159: I wonder if we need the clarification "for people with anxiety"? I think that any vague meeting invitations — where the recipient doesn't already expect it because you might've spoken about it previously — can be quite anxiety inducing for all. What do you think?

working-together/README.md
```diff
@@ -156,6 +156,11 @@ You can [download a double-sided printout] of these guides to display in your me
...
+    invitations can cause huge amounts of stress or
+    [panic attacks](https://en.wikipedia.org/wiki/Panic_attack).
+    - If you're incapable of sharing an overview or topic, at least share the
```
nickcharlton, line 162: "incapable" might imply someone doesn't know how, so maybe "unable" is better, given the further explanation?

Another thought: it is possible to make calendar events (at least on Google Calendar) private, so an uninvited colleague shouldn't be able to see it.

working-together/README.md
thoughtbot-summer: Hmmm, good point. I'd like to draw attention to anxiety (in the clinical sense) in particular; maybe:

> Strongly avoid vague meeting invitations. These can be stress-inducing for anyone; and for people with anxiety, they can cause panic attacks.

nickcharlton: Oh yes! That's great!

stevepolitodesign, line 159: [Minor] How do you feel about this? I feel like it's more direct.

```suggestion
  - Avoid vague meeting invitations. For people with anxiety, vague
```

stevepolitodesign (approved):
I like this. I was a recipient of a vague meeting invitation, and was anxious about it.

working-together/README.md
mike-burns: I don't have anxiety around meetings and also I think people should avoid vague meeting invitations for personal reasons.

Specifically: I avoid meetings when I can, so if the meeting invitation is vague then I need to do extra work to learn whether I can skip it. Save me some time, describe your meetings.

skybondsor: I feel attacked, @mike-burns, and I'm here for it!

mike-burns: Always happy to chat!

thoughtbot-summer: Thanks! I decided to say "don't" instead, which I feel is direct and also carries more weight than "avoid".

sarahraqueld (approved):
Nice!

PR #711, opened 2024-07-26
Remove inclusive culture guide

enatario:
We're moving non-technical guides over to our Playbook. [The inclusive culture guide now lives here](https://thoughtbot.com/playbook/best-practices/inclusive-culture).


LindseyChristensen (approved):
Looks good!

PR #715, opened 2024-09-13
Recommend zustand for global state management

MalcolmTomisin:
- Zustand has a similar API to Redux (Selectors etc)
- Way less boilerplate
- Growing popularity as Redux's gets limited to legacy codebases
- New projects get set up with new generation state managers (Tanstack-query, Valtio, MST, Zustand)
- Zustand is lighweight, lightweight is important for React Native
- Performant with transient updates that don't cause re-renders

### DIFF
--- a/react-native/README.md
+++ b/react-native/README.md
@@ -15,7 +15,7 @@
 * Use [React Navigation](https://reactnavigation.org/) for routing
 * Use [TanStack React Query](https://tanstack.com/query/v4/docs/framework/react/overview) as an API client for REST APIs
 * Use [Apollo Client](https://www.apollographql.com/docs/react/) as an API client for GraphQL APIs
-* Use [Redux Toolkit](https://redux-toolkit.js.org/) for global state
+* Use [Zustand](https://github.com/pmndrs/zustand) for global state
   * Avoid storing API data in a global store. Instead, use a dedicated API client.
 * Use [React Native Firebase](https://rnfirebase.io/) for push notifications
 * Use [Sentry](https://docs.sentry.io/platforms/react-native/) for error reporting
--- a/react/README.md
+++ b/react/README.md
@@ -15,7 +15,7 @@
 - Prefer the `children` prop over [render props]
 - Prefer using [TypeScript prop interfaces] over [PropTypes]
 - Prefer the [short syntax] when using [Fragments]
-- Prefer [React Contexts] over [Redux]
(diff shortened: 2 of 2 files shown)


pangolingo:
Is the mobile team using Zustand a lot? I've never worked with it. I'd love to see how we're using it on our client work.

MalcolmTomisin:
@pangolingo I don't think there has been opportunities to experience Zustand on client work. [New projects are more likely to use zustand](https://results.stateofreactnative.com/state-management/).

pangolingo:
I suggest that we convert this PR to a draft, and don't merge until until more thoughtbotters have enough first-hand experience to recommend Zustand over Redux.

codeofdiego:
Hi @pangolingo, yeah I like that suggestion. Ideally, we should get more experience with Zustand and adoption on client projects before changing the guide. @MalcolmTomisin I really liked the idea of writing a blog post about comparing both.

rakeshpetit:
> Zustand is lighweight, lightweight is important for React Native

React Native apps package all code into a single bundle which means the impact on lightweight library is very minimal.

- https://bundlephobia.com/package/@reduxjs/toolkit@1.8.1
- https://bundlephobia.com/package/redux@4.1.2
Both redux and the modern reduxjs/toolkit together are under 15 KB. Zustand is under 1 KB which is extremely tiny, not denying but does 15 KB matter when App bundles are a few MBs usually and is packaged anyways? I'm curious to understand why "lightweight is important for React Native"?

> Performant with transient updates that don't cause re-renders
I'm ready to be on the side of redux with this and fight you :) 

> Growing popularity as Redux's gets limited to legacy codebases
> New projects get set up with new generation state managers (Tanstack-query, Valtio, MST, Zustand)

We use Ruby on Rails as our web stack which is 20 years old, stable and just works. While we should focus on adopting modern solutions, instead of chasing them, we should be carefully choosing them. 

I think we should be introducing new libraries in a client project to test suitability, identify problems, and gather team feedback before recommending broader adoption such as updating our guides.
**Example:** Tanstack query (React query) that has been used in a few client projects reaping benefits with an enjoyable developer experience.

pangolingo:
We're not currently recommending Zustand - we don't feel strongly about it.
[Slack thread form September 2026.](https://thoughtbot.slack.com/archives/C02HVTS7H/p1788558350768959)

PR #716, opened 2024-09-24
Don't use foo, bar, and baz

cpytel:
It is possible that the term foobar has a [questionable
etymology](https://en.wikipedia.org/wiki/Foobar#History_and_etymology),
but in addition, these don't mean anything an in our code, as well as
our documentation, there is almost always a more meaningful example to
use that will better convey intent to other developers, and yourself in
the future.

We already have a line in our guides to "Name variables, methods, and
classes to reveal intent."

However, throughout our open source projects we have used this
placeholder variables in code and documentation when a more meaningful
alternative would have been better.

This adds the point that this guideline also applies to documentation,
and gives the examples of `foo`, `bar`, and `baz`, to hopefully make it
clear what the guideline is saying to not do.

### DIFF
--- a/general/README.md
+++ b/general/README.md
@@ -41,7 +41,7 @@ violations.
 - Prefer naming classes after domain concepts rather than patterns they
   implement (e.g. `Guest` vs `NullUser`, `CachedRequest` vs `RequestDecorator`).
 - Name the enumeration parameter the singular of the collection (`users.each { |user| greet(user) }`).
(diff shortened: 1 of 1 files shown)


thoughtbot-summer (approved):

slickepinne (approved):

nickcharlton (approved):
YES

sej3506 (approved):

MalcolmTomisin (approved):

PR #719, opened 2024-11-13
Require schemas for APIs

JoelQ:
When consuming an API from a client, a schema makes your life easier. Not only does it provide documentation, but it also can integrate into automated tooling such as type-checkers and code generators.

On the back end, it's nice to plug a schema into a unit test to guarantee that the implementation matches the spec. Depending on the system you use, schemas can also be used to generate API docs or test fixtures.

### DIFF
--- a/tech-stack/README.md
+++ b/tech-stack/README.md
@@ -36,6 +36,7 @@ while building applications.
 ### API
 
 - Use GraphQL as an API layer when connecting a mobile app to a web service.
+- Use a schema to describe the shape of the inputs and outputs of an API endpoint.
 
 ### Web
 


nickcharlton (approved):
To bring over my comment from the PR which spawned this one:

Do we have a general preference for tools for doing this?

I think something in the guides which provided specific direction there would be great (and also would make a good blog post!).

tech-stack/README.md
```diff
@@ -36,6 +36,7 @@ while building applications.
 ### API
 
 - Use GraphQL as an API layer when connecting a mobile app to a web service.
+- Use a schema to describe the shape of the inputs and outputs of an API endpoint.
```
pangolingo, line 39: Here are my opinions. This is what I generally do on a mature project. 

```suggestion
- Use an [OpenAPI schema](https://learn.openapis.org/) to describe the shape of the inputs and outputs of an API endpoint.
- When building a REST API, use the [JSON-API](https://jsonapi.org/) spec determine the shape of API responses
- When using GraphQL and Typescript, use [GraphQL Codegen](https://the-guild.dev/graphql/codegen) to generate type definitions for the API 
```

pangolingo: > Ooh JSON-API! I know that can be controversial. Any particular reason you like to default to it?

It's a well-documented opinionated standard, and I haven't come across another public standard that's as fully defined. I actually don't like some of its opinions, but I also don't want to write my own API standards doc on every project.

nickcharlton: Mmm. I think it's the one I've most frequently come across, that's well-supported, and is a good compliment to the HTTP specification.

I've been trying to think of others I've used and enjoyed, but it's been so long I'm coming up short. But I'd be happy with anything which emphasises the spirit of HTTP/the web and provides a standard we can follow.

laicuRoot: > Ooh JSON-API! I know that can be controversial. Any particular reason you like to default to it?

Why is JSON-API controversial @JoelQ ? 👀

emilford: @JoelQ do you have a follow up to @laicuRoot's question on JSON-API?

Anything more we want to consider on this PR before merging?

pangolingo (approved):
I say we go ahead and merge this PR without adding the opinionated JSON-API recommendation. I feel pretty strongly that an API schema should be required whenever we build an API - it's painful without one.

jutonz (approved):

PR #720, opened 2025-02-17
Use Pundit for authorization

nickcharlton:
Authorization is the process of ensuring that users have access to the data that only they should access. It's a common flow found in many different Rails applications.

Pundit is takes the approach of using regular Ruby classes and OO design which makes it pleasant to use.

This recommendation comes after it's been used on many applications, including support included in Administrate.

https://github.com/varvet/pundit
https://administrate-demo.herokuapp.com/authorization

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -54,6 +54,9 @@
 - [Use blocks](/ruby/sample_2.rb#L10) when declaring date and time attributes in
   FactoryBot factories.
 - Use `touch: true` when declaring `belongs_to` relationships.
+- Use [Pundit][] when you need to restrict access to models and data.
+
+[Pundit]: https://github.com/varvet/pundit
 
 ## Translations
 


DoodlingDev:
Love pundit. 👍🏻 

I might even push back on the idea that it's not for Suspenders even. Especially with authentication baked in to Rails 8.. and substantive application is going to need access control of some kind, otherwise one is probably ignoring a lot of the rest of Suspenders anyway

DoodlingDev (approved):

neilvcarvalho (approved):

JoelQ:
Agreed with @DoodlingDev, I think this would probably be a good fit in suspenders. Especially since we now have a generator-based architecture so it can be optional. Do we have a general sense that gem recommendations belong over there? Thoughts @stevepolitodesign?

JoelQ:
I could see a lot of value documenting our best practices around authorization though!

stevepolitodesign (approved):

stevepolitodesign:
> Agreed with @DoodlingDev, I think this would probably be a good fit in suspenders. Especially since we now have a generator-based architecture so it can be optional. Do we have a general sense that gem recommendations belong over there? Thoughts @stevepolitodesign?

For now, we're trying to keep Suspenders as close to `rails new` as possible.

Although *most* applications will benefit from authorization tooling, I'm not sure if it's needed on day 0.

For example, you can get pretty far with scoping queries to the `current_user`

```ruby
# Bad
@project = Project.find(params[:id])

# Good
@project = current_user.projects.find(params[:id])
```

rails/README.md
```diff
@@ -54,6 +54,9 @@
 - [Use blocks](/ruby/sample_2.rb#L10) when declaring date and time attributes in
   FactoryBot factories.
 - Use `touch: true` when declaring `belongs_to` relationships.
+- Use [Pundit][] when you need to restrict access to models and data.
```
stevepolitodesign, line 57: > When you are developing an application with Pundit it can be easy to forget to authorize some action. People are forgetful after all. Since Pundit encourages you to add the authorize call manually to each controller action, it's really easy to miss one.

Would it be too prescriptive to [enforce that policies and scopes are used](https://github.com/varvet/pundit?tab=readme-ov-file#ensuring-policies-and-scopes-are-used)

JoelQ: I'm in favor of that! Although you should be doing one or the other. I've worked in a codebase that does both on every action and it leads to lots of weird code, unnecessary `nil`,  and skipping the before actions 😭

purinkle (approved):

laicuRoot (approved):
✨

nickcharlton:
> For now, we're trying to keep Suspenders as close to rails new as possible.
> 
> Although most applications will benefit from authorization tooling, I'm not sure if it's needed on day 0.

A certain internal application of ours hasn't had any authorization aside from scoping queries until this week!

I think it's something I'd add quite soon, but not immediately in a new application. I wonder if we've got a specific way we'd like to set it up, it would be a good candidate for a generator on Suspenders, just one that we don't run by default?

stevepolitodesign:
> I wonder if we've got a specific way we'd like to set it up, it would be a good candidate for a generator on Suspenders, just one that we don't run by default?

I'd be curious to explore this in a general sense. Things like this might pair well with The Guides.

nickcharlton:
I've pulled out the idea of having future generators into an issue on Suspenders, where it might be nice to start thinking about specifics and other generator ideas: https://github.com/thoughtbot/suspenders/issues/1250

PR #721, opened 2025-03-10
Rails: Update guidance on development environment seed data.

stevepolitodesign:
Relates to [this proposal][1] for Suspenders.

> We have relied on [`dev:prime`][2] for loading data necessary for users
to view most of the features of the app in development.
>
> However, there were a few areas of improvement.
>
> First, rename `dev` namespace to `development`, and rename
[`prime`][1] task to `seed` for improved clarity.
>
> The, introduces `development:seed:replant` take to create parity with
the existing `db:seed:replant` task.

We're already doing this in ChatBot.

https://github.com/thoughtbot/chat_bot/pull/52

[1]: https://github.com/thoughtbot/suspenders/pull/1251
[2]: https://thoughtbot.com/blog/priming-the-pump

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -79,7 +79,7 @@ Guidance on ActiveRecord, ActiveModel, and other model objects.
   suffixes.
 - Keep `db/schema.rb` or `db/development_structure.sql` under version control.
 - Use `db/seeds.rb` for data that is required in all environments.
-- Use `dev:prime` rake task for development environment seed data.
+- Use `development:db:seed` rake task for development environment seed data.
 
 ## Migrations
 
--- a/rails/README.md
+++ b/rails/README.md
@@ -79,7 +79,7 @@ Guidance on ActiveRecord, ActiveModel, and other model objects.
   suffixes.
 - Keep `db/schema.rb` or `db/development_structure.sql` under version control.
 - Use `db/seeds.rb` for data that is required in all environments.
-- Use `development:db:seed` rake task for development environment seed data.
+- Use `development:db:seed` rake task for development environment seed data. [Example](/rails/how-to/seed-data.md).
 
 ## Migrations
 
--- a/rails/how-to/seed-data.md
+++ b/rails/how-to/seed-data.md
@@ -1,0 +1,45 @@
+# How to seed development data
+
+```ruby
+# lib/development/seeder.rb
+
+require "factory_bot"
+
+module Development
+  class Seeder
+    include FactoryBot::Syntax::Methods
+
+    def self.load_seeds
+      new.load_seeds
+    end
+
+    def load_seeds
+      emails = %w[ralph@example.com ruby@example.com]
+
+      emails.each do |email|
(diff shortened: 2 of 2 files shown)


rails/README.md
```diff
@@ -43,7 +43,7 @@
 - Validate the associated `belongs_to` object (`user`), not the database column
   (`user_id`).
 - Use `db/seeds.rb` for data that is required in all environments.
-- Use `dev:prime` rake task for development environment seed data.
+- Use `development:seed` rake task for development environment seed data.
```
vburzynski, line 46: **idea (non-blocking):** What would you think about using `development:db:seed`?

- This would mirror the `db:seed` rake task, but append the environment context to the front.
- would leave room for things like `development:redis:seed`

_note - mentioned this in slack already, adding here for posterity_

JoelQ:
Does this rename blur what seeds actually are? I guess everyone is already confused about the distinction between dev data and seeds. Maybe this is an "if you can't beat them, join them" solution?

stevepolitodesign:
> Does this rename blur what seeds actually are? I guess everyone is already confused about the distinction between dev data and seeds. Maybe this is an "if you can't beat them, join them" solution?

I personally think it's more clear than `dev:prime`.  And, if we use the suggestion made by @vburzynski, it will read `development:db:seed` which I think is even more clear than my proposal.

I just can't think of the better name for what we're doing other than "seed".

vburzynski:
I think it's fine to use `seed` here. In general "seeding" a database is just populating it with some set of data.  The conventions for the specific Ruby on Rails `db:seed` task are to use it for seeding the database with data needed for the application to run in any environment. Which suggests that the are other sets of seed data that may be required for applications to run in specific environments or even in specific contexts. They're all still an act of seeding, just not the primary `db:seed`

I could also see other variants working as well:

- `db:seed:development` 
  - `db` is the primary object we're acting upon
  - `seed` is the action
  - `development` is a targeting modifier in this naming scheme
- `development:db:seed`
  - `development` here is used a prepended namespace or context
  - `db` is once again the primary object we're acting upon
  - `seed` is the action
  - this might make sense if there were additional rake tasks scoped to and specific to the development environment
- `db:development:seed`
  - I don't know if I like this one, but included it for completeness. 🙃
- or even `db:seed:with_development`
  - this would stray from the current implementation but takes a cue from [`data-migrate`](https://github.com/ilyakatz/data-migrate) which extends `db:migrate` with the option `db:migrate:with_data`
  - it's even more clear that this is a modifier; but in this case one that extends the original. I'd expect this to run both the normal seed and the seed data specific to development.

emilford:
@stevepolitodesign what are your thoughts on @vburzynski's suggestions?

stevepolitodesign:
@emilford thank you for bumping this. I just addressed this, and wanted to highlight that we're doing this in ChatBot.

https://github.com/thoughtbot/chat_bot/pull/52

vburzynski (approved):

PR #722, opened 2025-03-14
Use an opinionated set of rules for Rubocop

neilvcarvalho:
Most projects that thoughtbot is working on currently did not start here. The "Use standard" recommendation comes from a time that that wasn't true.

There are a few reasons not to use Standard in existing projects:
- A large project does not already use Standard and retrofitting would be costly
- Need or want Rubocop rules other than Standard's opinionated set

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -2,7 +2,8 @@
 
 [Sample 1](sample_1.rb) [Sample 2](sample_2.rb)
 
-- Use [standard]
+- Use an opinionated set of rules for Rubocop. Prefer [standard] for new projects.
+  - Having an already decided configuration removes bikeshedding on what a Rubocop configuration should be.
 - Avoid conditional modifiers (lines that end with conditionals). [36491dbb9]
 - Avoid multiple assignments per line (`one, two = 1, 2`). [#109]
 - Avoid organizational comments (`# Validations`). [#63]


vburzynski:
Some reasoning behind why I would vote to remove the recommendation of using Standard

- The name has always felt like a form of doublespeak. It seems to deliberately obscure and distort the meaning of the word "standard". The README even furthers the confusion with language such as "Ruby Standard-compliant teams & projects". Which co-opts language used for indicating something is "standards compliant".
- Standard does not always follow community conventions. For example, requiring that hashes remove spaces inside the beginning and end of hash literals. They do have opinionated arguments for why they chose this styling, but it speaks towards how Standard applies non-standards preferences.
- Last year I wrote a ruby script to enumerate all the differences between Standard and the default Rubocop configuration. There is a nontrivial number of useful cops that are disabled. Due to the nature of Standard being "unconfigurable", you are not allowed to re-enable these.
  - One example is that Standard disables all of the `Metrics/*` cops; It would be impossible to set a one-size-fits-all configuration of these rules as they largely all should be tweaked and tuned to individual projects. Standard therefore has to disable them to make them universally applicable. https://github.com/standardrb/standard/issues/102
  - These cops can be highly advantageous to use on projects.
  - I'd have to dig through my findings again to list other examples.
  - Some of the cops fall outside the purview of the Standard project's goals, yet the gem disallows you from benefiting from their use
- While it is currently possible to run both rubocop and standard at the same time, it is NOT supported.
  - The README includes the following warning: "Please note that the following usage is not supported and may break at any time. Use at your own risk and please refrain from opening GitHub issues with respect to loading Standard or its plugins' YAML configurations for use by the rubocop CLI."
  - the project will not support using Rubocop, even when it relies heavily on the gem.
  - running both tools is an added cost that is experienced both during local development and on CI.

My strongest point from the items above is that Standard makes it unnecessarily difficult and risky to use common Rubocop cops that fall outside their preferences which center on styling and layout.

cpytel:
Here is some history on this: https://github.com/thoughtbot/guides/commit/6d86958c9620d2f302cedf4c994232eae2323dff

The config is still here https://github.com/thoughtbot/guides/blob/main/ruby/.rubocop.yml (or was restored, I'm not sure), though it may have fallen out of maintenance or be reduced.

That being said, I'm not sure about the arguments about existing code being a factor here are really applicable? One of the overarching guidelines in the guide is "Don't rewrite existing code to follow this guide."

cpytel:
Ah, it was added back here https://github.com/thoughtbot/guides/commit/9ac4a389e01af5db97fd3528a1d2aaf156efbf74

nickcharlton:
I had restored it in #623, because we used the shared configuration feature of Hound on Administrate (and probably others, I don't recall) and we couldn't move quite so fast on that project (removing it caused hundreds of comments on everyone's PRs, it was unpleasant).

We don't use it any more, although I'm yet to tidy up the final bits of linting to remove it entirely. That said, it's not been used in a while now, because we did switch to Standard.

vburzynski:
I submit a proposal for adding a link to a page discussing the context, problem, and solution around the guideline...

```
# Use an opinionated set of rules for Rubocop

Ruby code can be written in many different styles and still be syntactically correct. The presence of divergent style and
format can disrupt communication as well as complicate feature development. Code that exhibits a consistent style is
easier to understand and maintain. The absence of consistent enforced style can result in unnecessary code churn.

**Therefore:** Use an opinionated set of rules for Rubocop when coding in Ruby. Prefer [standard] for new projects.

- By employing an already decided configuration, you avoid bikeshedding on what the Rubocop configuration should be.
- Code that is consistently formatted and styled will be easier to work with.

There are, however, situations where one might choose not to introduce Standard in a project. After all, change for the sake of change can be disruptive. A few examples include:

- An existing project may already employ it's own opinionated configuration of Rubocop and that set of rules is working.
- A large project, which does not already use Standard, might require a costly amount of time to refactor and retrofit
  existing code to conform to a new set of conventions.
- There exists a need or desire to use Cops which are disabled by Standard

Overall, the goal is to increase the quality and consistency of code while not getting distracted by disproportionally
minor or trivial details. It's more important for an implementation team to agree to follow shared conventions than it 
is to enforce a specific configuration.

[standard]: https://github.com/testdouble/standard
```

I also have a pattern in mind for similar pages that I'm still working on.

We could also include links to many of the pull-requests mentioned in this discussion to cover historical context.

vburzynski:
- expand more of the reasoning document to include the WHYs of using or not using Standard.

PR #726, opened 2025-03-21
Add Markdown Linting

vburzynski:
- Adds `markdownlint-cli2` to lint the markdown files
- configures markdownlint options
- fixes several minor issues and normalizes style to be consistent
- adds `lefthook` and a pre-commit hook to run markdownlint on staged changes

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
   lint:
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -8,4 +8,4 @@ jobs:
     - uses: DavidAnson/markdownlint-cli2-action@v19
       with:
         globs: |
-          *.md
+          ./**/*.md
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,11 @@
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: DavidAnson/markdownlint-cli2-action@v19
+      with:
+        globs: |
+          *.md
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,2 @@
+[tools]
+node = "latest"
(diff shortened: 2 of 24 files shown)


mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+node = "latest"
```
brian-penguin, line 2: neat! I had not heard of mise before. Sort of like a make meets asdf type deal? 
I'm curious what the majority of people use for a language management system at thoughtbot. I think I would assume that asdf is the most default but I have no idea. 

Would you be opposed to also adding a `.tool-versions file?`

rails/how-to/start_a_new_rails_app.md
```diff
@@ -3,11 +3,11 @@
 Use [Suspenders]:
 
 ```sh
-$ gem install suspenders
```
brian-penguin, line 6: This is an interesting one. I have mixed feelings about `$` to prevent whole "copy paste didn't read" interactions. On one hand it's probably good we do that. On the other, it's probably confusing to devs who aren't used to the syntax. Wondering if anyone has stronger opinions on it

brian-penguin (approved):
I like it! the less I have to think about syntax the better.

.github/workflows/linting.yml
```diff
@@ -0,0 +1,11 @@
+on: [push, pull_request]
```
brian-penguin, line 1: Would just pull_request work?

mise.toml
vburzynski: 👍 I wouldn't be opposed to adding `.tool-versions`

[`mise`](https://mise.jdx.dev/) is a drop in replacement for `asdf`. I swapped to it after `asdf` v0.16 broke a whole bunch of things and I couldn't get node or something to installl

rails/how-to/start_a_new_rails_app.md
vburzynski: I compared this to other code blocks marked with `console`/`shell` etc, and all of the other instances did not include the `$` or other marker. My preference would be not adding in `$`, but it's not a strong opinion. Whichever way we went, I'd recommend making it consistent everywhere.

mise.toml
vburzynski: Added 😎

.github/workflows/linting.yml
vburzynski: changed

PR #731, opened 2025-03-28
Remove outdated email guides

nickcharlton:
We were mentioning parts from Rails 4.0 and below, so remove those and point to the current version of Rails guides for previews, which not everyone knows about.


JoelQ (approved):
yay for modernization!

sarahraqueld (approved):

PR #732, opened 2025-03-28
Start recommending baseline feature support

nickcharlton:
Instead of targeting particular browser versions, we should aim to use browser features which are widely supported across major browsers in use. This allows us to balance using new web functionality, whilst avoiding maintaining lists of supported versions.

For example, as of 2025-03-28, CSS Grid Layout (level 1) is marked at widely available on caniuse.com.

https://web.dev/baseline/
https://web-platform-dx.github.io/web-features/
https://caniuse.com/css-grid


JoelQ (approved):

jutonz (approved):
nice!

PR #734, opened 2025-03-28
Treat version history as a resource

nickcharlton:
When composed with care, the version history in a code repository becomes a valuable resource for the future.

For example, it allows you to use `git blame` to see how files changed and understand from the commit messages _why_ such changes were made, allowing you to understand the constraints under what those changes can be made and explain why the code was written that way.

To do this, you should ensure you write enough detail for someone else (or perhaps yourself in future) to understand the change in depth. You should link to external resources, but understand that those can change and so a link alone is not enough. You should rebase to avoid messy ordering, and compose your changes with care so that they can be understood at a glance.

### DIFF
--- a/git/README.md
+++ b/git/README.md
@@ -4,6 +4,7 @@ A guide for programming within version control.
 
 ## Best Practices
 
+- Treat the version history as a resource for the future.
 - Avoid merge commits by using a [rebase workflow].
 - Squash multiple trivial commits into a single commit.
 - Write a [good commit message].


git/README.md
```diff
@@ -4,6 +4,7 @@ A guide for programming within version control.
 
 ## Best Practices
 
+- Treat the version history as a resource for the future.
```
JoelQ, line 7: Can you elaborate a bit more here? I love what you have in the PR description although that's probably too long. Maybe some kind of happy medium?

JoelQ (approved):
Yes! Big fan of this!

git/README.md
nickcharlton: I couldn't work out how to shorten it to fit in a line or two. Any ideas?

stevepolitodesign: Maybe something along the lines of "Explain the _why_ and not the what"?

jutonz: An idea for incorporating some of the content from the PR description

```suggestion
- Treat the version history as a resource for the future. Structure your commits such that they tell the story of _why_ the change was made, so that future developers--including yourself!--can understand the constraints which influenced your decisions at the time. 
```

jutonz (approved):
I love it! tyvm :)

git/README.md
MatheusRich: In #738 we introduced the pattern of adding a markdown file with the reasoning and examples. You might consider doing the same.

Those would be nice blog posts too.

emilford: @nickcharlton what do you think about @jutonz's suggestion and then merging this?

vburzynski, line 7: **suggestion:** I think one part of the previous suggested change highlights the intent and objective in a clear and concrete way:

```suggestion
- Structure commits to tell the story of why the change was made
```

The rest of the reasoning reads as a great start of a "guideline page" to expound on the reasoning behind why it's a good guideline to follow...

> Treat the version history as a resource for the future. Structure your commits such that they tell the story of _why_ the change was made, so that future developers--including yourself!--can understand the constraints which influenced your decisions at the time.

There's a few examples of guideline pages being broken out in the Ruby guides README.md and I think some other recent PRs as well.

vburzynski (commented):
I feel like the PR description and details from the discussion could make for a great "guideline page". What would you think about keeping the bullet point succinct and to the point with a clear actionable phrasing while putting the reasoning into a linked sub-page?

PR #725, opened 2025-03-21
Group Rails guidance into sections

vburzynski:
- diffuses/chunks/reorders the Rails guidance into sections
- normalizes the markdown link references so that they appear under the sections they're used within

PR #735, opened 2025-03-28
Write commit messages for humans

nickcharlton:
You should tell a story for your fellow programmers, not fit within arbitary rules followed in other places.

However, if the repository is already following such a rule, you should do so for consistency.

### DIFF
--- a/git/README.md
+++ b/git/README.md
@@ -7,6 +7,8 @@ A guide for programming within version control.
 - Avoid merge commits by using a [rebase workflow].
 - Squash multiple trivial commits into a single commit.
 - Write a [good commit message].
+- Write commit messages for humans, not machines (i.e.: avoid "chore:" and 
+  other Git patterns).
 
 [rebase workflow]: https://github.com/thoughtbot/guides/blob/main/git/README.md#merge
 [good commit message]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html


nickcharlton:
I wish for a better argument for why here, but I ran out of ideas. Aside from reading fairly "robotic", it's usually meaningless (what even is a chore? isn't all programming at some level, a chore?) and typically not even used for anything. Making the words harder to understand.

git/README.md
```diff
@@ -7,6 +7,8 @@ A guide for programming within version control.
 - Avoid merge commits by using a [rebase workflow].
 - Squash multiple trivial commits into a single commit.
 - Write a [good commit message].
+- Write commit messages for humans, not machines (i.e.: avoid "chore:" and
```
stevepolitodesign, line 10: I actually don't mind those prefixes, since it gives me context as a scan through the commit history.

What I find most valuable (and what I think you are proposing) is that each commit explains the **why**, and not the what. This might relate to #734.

nickcharlton: Yeah, it's all about being purposeful with how you go about your changes, which is why I ended up doing this right after #734.

Does prefixing your commits provide value for other humans? If we're doing that consistently throughout a repository, go ahead. Otherwise consider if it's better to express yourself in a more human way.

vburzynski: I often follow the conventional commit format (https://www.conventionalcommits.org/en/v1.0.0/#summary) and I usually find prefixing commits to be highly valuable.

prefixing and following a standard format facilitates a level of cognitive funneling. When a commit is prefixed with a `type`, a human searching through the history can make a quicker judgement call if the commit is relevant to what they're looking for. If they're looking for new features, they look for the `feat:` prefixes. If it's a bug fix then `fix:`. The format `<type>[optional scope]: <description>` lays out information so that as you read it, you are presented the broadest level of details first and the most narrow and specific last. Including scopes helps with this even more. If what I'm searching for doesn't fit the type or scope, then I don't even have to read the description.

Knowing I need to tag and scope my commit message can also encourage me to keep my commits smaller and more atomic. I shouldn't fix a bug, develop a feature, modify CI config, update docs, add tests, fix style and refactor all in one commit because each of those have their own type tag and I can make granular commits for each of them. Also, if I get carried away making too many changes without committing, the format then encourages me to break up my work into smaller commits to write a commit message that conforms to the format.

vburzynski: Some examples:
- https://github.com/thoughtbot/guides/pull/725
  - I used a `style` type prefix at first because I wasn't adding new content, but working on style and layout
  - I used `refactor` to show I committed some rework
  - I also used `fix` to show I was addressing a linting issue. Which on a code project could be `style`, but the exact mapping of types isn't consequential; it's important that in any one context everyone uses them the same way (a ubiquitous language if you will).
  - _My commit descriptions were still more oriented towards WHAT rather than WHY, but the context is clear (I'm still working on this myself_
- https://github.com/thoughtbot/guides/pull/726
  -  I introduced Markdownlint, so that was a new feature (`feat:`)
  - configuring the linting rules related to conventions and style guide, so I used `style:`
  - I configured the linting to run on github actions -- related to the `build:`
  - lots of `style` commits, one for each rule, all scoped to `markdown`
  - _due to the technical nature, the commit descriptions cover more of WHAT was done_

more references:
- https://github.com/angular/angular/blob/main/contributing-docs/commit-message-guidelines.md#commit-message-header

vburzynski: _addendum — I've focused more on the human aspects above due to the conversation, but following a standard commit structure is also useful for automation and tooling (for example: automating the generation of CHANGELOGs). Though this can again be reconnected to the human aspect, if you have a standard set of tokens/slugs/prefixes in your commits, then a human can more easily use command line tools to filter the git history_

vburzynski (commented):
Is there another way to phrase the intent of the original guideline without stating that we shouldn't use prefixes? There's many benefits (for humans as well) to adding these. I would like to see the orientation of focusing on humans preserved in some way.

Alternatively, is the intent already sufficiently covered by the guideline from #734?

PR #736, opened 2025-03-28
Rails: Prefer Devise for Authentication

stevepolitodesign:
This might be controversial since we maintain [Clearance][], but I've
found that I've only been on projects that use Devise during my time at
thoughtbot.

Although Rails introduced an [authentication generator][], I feel that
Devise it still preferable for the following reasons:

- It receives frequent security updates, whereas you're on your own with
  a generator.
- It's widely adopted in the Rails community.
- It has a [rich ecosystem].

[Clearance]: https://github.com/thoughtbot/clearance
[authentication generator]: https://guides.rubyonrails.org/security.html#authentication
[rich ecosystem]: https://github.com/heartcombo/devise?tab=readme-ov-file#extensions

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -130,6 +130,15 @@ others' Rails work, look in particular for:
 
 [ActiveStorage]: https://guides.rubyonrails.org/active_storage_overview.html
 
+## Authentication
+
+- Prefer [Devise][] over the [authentication generator][], due to its wide
+  adoption, rich [ecosystem][], and security maintenance.
+
+[Devise]: https://github.com/heartcombo/devise
+[authentication generator]: https://guides.rubyonrails.org/security.html#authentication
+[ecosystem]: https://github.com/heartcombo/devise?tab=readme-ov-file#extensions
+
 ## How to...
 
 - [Start a New Rails App](./how-to/start_a_new_rails_app.md)


rails/README.md
```diff
@@ -130,6 +130,15 @@ others' Rails work, look in particular for:
 
 [ActiveStorage]: https://guides.rubyonrails.org/active_storage_overview.html
 
+## Authentication
+
+- Prefer [Devise][] over the [authentication generator][], due to its wide
+  adoption, rich [ecosystem][], and security maintenance.
```
vburzynski, line 136: 1. Could this be expanded to cover the set of architectural characteristics and project needs which would indicate Devise being a better choice and what set would indicate one should use Clearance?
  - For example, a simple forms-over-data CRUD and simple REST based small Rails app might benefit more from Clearance. Clearance might support architectural characteristics like simplicity, maintainability, agility; as it's a more lightweight solution. It could also cost-less to implement if its feature set covers the needs of the app. Yet if an app is likely to expand beyond what Clearance can offer, the implementation team would need to weigh the cost of replacement or enhancing Clearance. 
  - on the other hand, an app with architectural characteristics that prioritize upgradeability, extensibility, configurability, and leverageability/reuse might want to choose Devise. Devise comes with a long list of third party extensions and a larger feature set. It's more commonly used across Rails applications. That large feature set comes with some complexity and challenges, but it's more ubiquitous in the Rails community.
  - so there are tradeoffs to using one or the other.
2. This guideline might also benefit from creating a "guideline page" to expound on the reasoning (including what's mentioned in the PR description).

PR #738, opened 2025-03-28
Limit use of conditional modifiers to short, simple cases.

MatheusRich:
This PR changes the guideline of "Avoiding conditional modifiers", which I believe lacked nuance.

This was product of a group discussion, and we thought it was good to introduce our reasoning as part of the guideline. That way, there's a new markdown file with the reasoning and examples. This might be a new pattern for the guides.


ruby/README.md
```diff
@@ -2,8 +2,12 @@
 
 [Sample 1](sample_1.rb) [Sample 2](sample_2.rb)
 
+> [!TIP]
```
MatheusRich, line 5: This renders as
image

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
...
+`if` with multiple conditions `if a && b`) prefer the multi-line form:
+
+```ruby
+# Avoid
+do_something if condition && condition2
+
+# Prefer
+if condition && condition2
+  do_something
+end
```
MatheusRich, line 33: I'd love a "real world"-like example here. "condition && condition2" might be too abstract for an example

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
...
+
+  destroy_all
+end
+```
+
+## Use your best judgment
+
+There might be cases where the conditional modifier work well with multiple
+conditions, so use your best judgment. Overall, prefer writing code that is easy
+to read and understand, while being idiomatic.
```
MatheusRich, line 87: Is this necessary?

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
+# Limit use of conditional modifiers to short, simple cases
+
+Conditional modifiers (i.e., `if` or `unless` at the end of a line) be
```
rdnewmanbot, line 3:
```suggestion
Conditional modifiers (i.e., `if` or `unless` at the end of a line) can be
```

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
...
+```rb
+if async?
+  do_later
+end
+```
+
+## Complex conditions
+
+However, if the line is too long (around 80 characters) or complex (e.g., an
+`if` with multiple conditions `if a && b`) prefer the multi-line form:
```
rdnewmanbot, line 24:
```suggestion
`if` with multiple conditions like `if a && b`) prefer the multi-line form:
```

ruby/conditional_modifiers.md
rdnewmanbot: In general, I like the advice to use judgement instead of blindly following a role. But other than "the conditional...multiple conditions", that could be generic advice. Perhaps the first sentence is better under `## Complex conditions` ?

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
...
+
+```ruby
+# Avoid
+do_something if condition && condition2
```
rdnewmanbot, line 28:
```suggestion
record_admission if user.admin? && user.active?
```

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Prefer
+if condition && condition2
```
rdnewmanbot, line 31:
```suggestion
if user.admin? && user.active?
```

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,87 @@
...
+  do_something
```
rdnewmanbot, line 32:
```suggestion
  record_admission
```

ruby/conditional_modifiers.md
rdnewmanbot: My suggestion is still pretty contrived. I'll keep thinking on it.

rdnewmanbot (changes requested):
I'm glad we're modeling this!

ruby/conditional_modifiers.md
MatheusRich: I added a new one. LMK what you think

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+## Complex conditions
+
+However, if the line is too long (around 80 characters) or complex (e.g., an
+`if` with multiple conditions like `if a && b`) prefer the multi-line form:
+
+```ruby
+# Avoid
+block_access! if signed_in? && !current_user.active?
```
rdnewmanbot, line 28: yeah, that's great!

rdnewmanbot (approved):
Thank you!!!

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,86 @@
...
+```rb
+if async?
+  do_later
+end
+```
+
+## Complex conditions
+
+However, if the line is too long (around 80 characters) or complex (e.g., an
+`if` with multiple conditions like `if a && b`) prefer the multi-line form:
```
JoelQ, line 24: This is especially true with `unless` since you an introduce a DeMorgan bug

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,86 @@
...
+If the conditions are related, consider extracting a method that groups them.
+This might allow you to use the conditional modifier form again.
+
+```ruby
+def inactive_user?
+  signed_in? && !current_user.active?
+end
+
+block_access! if inactive_user?
+```
```
JoelQ, line 50: This seems to be moving towards the  [Triangle of Separation](https://thoughtbot.com/blog/triangle-of-separation) style of code

ruby/conditional_modifiers.md
```diff
@@ -0,0 +1,86 @@
...
+You can also refactor the code so the less destructive action uses a conditional
+modifier, which pairs well with the informal feel of the modifier form:
+
+```rb
+def action
+  return do_nothing if chill?
+
+  destroy_all
+end
+```
```
JoelQ, line 86: I think this section is really about what is the "main action" of the method, and what is the edge-case blocked by the guard close. It seems less about formality and more about verifying invariants?

JoelQ (commented):
The examples you give all seem to be pre-conditions and invariants. Maybe we just want allow this style for guard clauses? How do we feel about inline conditions in the middle of a method body?

JoelQ:
If we allow inline conditions in the middle of the body, how do we feel about the style that attempts to "flatten" a conditional to fewer lines like:

```ruby
def can_edit_admin_post?
  return owner? unless admin?
  true
end
```
that often ends up more complex than the explicit if/else. The example above is re-implementing boolean `||`

```ruby
def can_edit_admin_post?
  owner? || admin?
end
```

https://thoughtbot.com/blog/back-to-basics-booleans#multiple-variables

JoelQ:
This guideline intersects with some other topics too:
- How do we feel about `if` without a corresponding `else` (now it's implicit, possibly `nil`)?
- How do we feel about early returns?

ruby/conditional_modifiers.md
MatheusRich: I'd say both conditions can be equally valid, which is highlighted in the if/else example? The writer has the power to bring focus to particular branch by *not* using the guard. Maybe that's what I was trying to communicate with formal/informal.

MatheusRich:
@JoelQ it seems to me that inside a method body those would behave the same?

```rb
def brew
  boil_water
  add_coffee_grounds
  pour_water
  stir

  add_sugar if sugar?
  add_cream if cream?

  serve
end
```

ruby/conditional_modifiers.md
vburzynski, line 1: I'd be interested in finding a standardized layout, form, or pattern (informal definition) for guideline pages. Something akin to [Pattern Forms](https://wiki.c2.com/?PatternForms). Though I acknowledge that we aren't generating [formal patterns](https://wiki.c2.com/?WikiPagesAboutWhatArePatterns) here, but more sharing best practices, maxims, preferences, conventions, heuristics, lessons learned, etc.

Still, I would argue that creating a structural consistency across guideline pages would facilitate the readability, scanability, searchability, navigability, and organization of the pages. This would encourage internal consistency across a diverse set of guidelines.

I'm not sure what this form would look like (we might need to build out several pages first and discover the common pattern). Perhaps something like: **problem**, **solution**, **tradeoffs** (indications/contraindications; pros/cons; advantages/disadvantages). 

I could also see us using a variant the portland form as well...
- opening statement — _roughly cover the description, problem, context, and forces to be resolved_
- therefore...  — _thesis, solution, the action to take, resolution, resulting context_
- but... — _antithesis, contraindications, exceptions, things that contrast the concepts already mentioned_

It enacts just enough structure to assist with pages being consistent while still being ad hoc and flexible.

PR #746, opened 2025-04-04
Update general guideline

slickepinne:
We updated the general guidelines and moved language specific guidelines to the relevant folders.

Co-authored-by: Valeria Graffeo <valeria@thoughtbot.com>
Co-authored-by: Rob Whittaker <rob@thoughtbot.com>
Co-authored-by: Trésor Bireke <tresor@thoughtbot.com>


general/README.md
```diff
@@ -12,25 +12,13 @@ Style and best practices that apply to all languages and frameworks.
...
-violations.
-
-[hound]: https://houndci.com
-
 ## Formatting
 
-- Avoid inline comments.
 - Break long lines after 80 characters.
-- Delete trailing whitespace.
+- Delete trailing spaces.
```
JoelQ, line 23: Whitespace is more than just spaces. It includes other invisible characters such as tabs, newlines, carriage returns or the unicode zero-width-joiner.

JoelQ (commented):
This feels like a mix of multiple unrelated changes:

1. Delete Hound guideline
2. Move some guidelines out general
3. Reformat part of general
4. Change guideline on whitespace

Thoughts on making each of these their own commit? Maybe even their own PR? That would allow conversation to happen independently for each, and if one proves controversial then it doesn't block merging the others.

general/README.md
slickepinne: Would "blank spaces" include these characters? Our main goal was to not use "whitespace".

general/README.md
```diff
@@ -19,16 +19,11 @@ violations.
 
 ## Formatting
 
-- Avoid inline comments.
 - Break long lines after 80 characters.
-- Delete trailing whitespace.
+- Delete trailing spaces.
 - Don't misspell.
-- Use empty lines around multi-line blocks.
-- Use spaces around operators, except for unary operators, such as `!`.
```
vburzynski, line 27: I'm curious, why wouldn't one want to use spaces around operators? adding them in can improve readability/legibility. It also applies more broadly than just Ruby.

```ruby
arr << 1
arr<<2
```

```javascript
const result=1+1;
const other = 1 + 2;
```

_EDIT: saw this was moved to ruby; I think this guidance would apply broadly to most if not all programming languages_

general/README.md
JoelQ: Maybe, but I think it's more ambiguous because some of the characters aren't spaces.

"Whitespace" seems to be the precise technical term used in e.g. the [whatwg spec](https://infra.spec.whatwg.org/#ascii-whitespace), the [ECMAScript spec](https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-white-space), or the [Unicode spec](https://www.unicode.org/reports/tr44/#White_Space) to describe the class of characters (although they all define the set slightly differently).

Is "whitespace" in the context of characters and typography considered problematic?

vburzynski: I'd also want to ensure we're covering more than just trailing spaces as there's many other characters that one should avoid trailing a line.

general/README.md
slickepinne: I'm with you and happy to move it back.

general/README.md
vburzynski: TL;DR: Perhaps "delete trailing spaces" is the better phrasing if we do need to get rid of "whitespace". 

Whitespace is the more precise technical term, but there are instances where "trailing whitespace" is shortened to "trailing spaces".  I would, however, strongly argue against using "blank space" or "negative space" over "whitespace". 

In general, I agree with attempting to use more inclusive verbiage where possible. Yet, in regards to this use case, I think the alternatives muddle the clarity of the guideline. There are instances, however, where _trailing whitespace_ is referred to as _trailing spaces_ so that option would be acceptable (see the link to ESLint below for an example).

More often than not, the trailing characters are spaces, but they could also be many other characters. A *whitespace character* (no space), as Joel mentioned, is a specific technical concept. There are 25 Unicode characters that fall under the category (see <https://en.wikipedia.org/wiki/Whitespace_character>)

> Whitespace refers to characters which are used to provide horizontal or vertical space between other characters. Whitespace is often used to separate tokens in HTML, CSS, JavaScript, and other computer language
> — <https://developer.mozilla.org/en-US/docs/Glossary/Whitespace>

On the other hand, [White Space](https://en.wikipedia.org/wiki/White_space_(visual_arts)) (the visual arts concept) is more closely related to the concept of [Negative Space](https://en.wikipedia.org/wiki/Negative_space). There are still slight nuances between the two concepts, but are often interchangeable *in the context of design and art*. Negative space is the unused space in relation to one or more subjects within that space. White space is the unused or unmarked space. So they're talking about the same thing, but using two different frames of reference.

- _trailing whitespace characters_ would be the most precise phrase
- _trailing spaces_ would be less precise, but acceptable and has demonstrable usage
- _negative space_ is not used to describe *whitespace characters* in the context of programming.
- _blank space_ feels too ambiguous
- _white space_ is either a different visual arts concept or a misspelling of whitespace.

example usage of "trailing spaces"
- <https://eslint.org/docs/latest/rules/no-trailing-spaces>

example usage of "trailing whitespace"
- <https://docs.rubocop.org/rubocop/cops_layout.html#layouttrailingwhitespace>
- vscode refers to it as trailing whitespace

general/README.md
slickepinne: Done in https://github.com/thoughtbot/guides/pull/746/commits/94870695ff5a3396ec049206633a2e8a4e5c9425

purinkle (approved):
:person_in_tuxedo: Looks good to me!

general/README.md
heyvaleria: I didn't think whitespace was to be removed as per inclusive language, but if that's the case, I'm in favor of "trailing spaces" as well, FWIW.

heyvaleria (approved):

PR #747, opened 2025-04-04
Rails: Default to "sandbox" mode when accessing production-like environments

stevepolitodesign:
In an effort to avoid accidental writing to the production database, we
recommend setting [config.sandbox_by_default][sandbox] to `true` in
production-like environments.

[sandbox]: https://guides.rubyonrails.org/configuring.html#config-sandbox-by-default

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -81,6 +81,12 @@ Guidance on ActiveRecord, ActiveModel, and other model objects.
 - Use `db/seeds.rb` for data that is required in all environments.
 - Use `dev:prime` rake task for development environment seed data.
 
+## Console
+
+- Access production-like [console][] sessions in "sandbox" mode: `bin/rails console --sandbox`
+
+[console]: https://guides.rubyonrails.org/command_line.html#bin-rails-console
+
 ## Migrations
 
 [Sample](migration.rb)
--- a/rails/README.md
+++ b/rails/README.md
@@ -81,11 +81,11 @@ Guidance on ActiveRecord, ActiveModel, and other model objects.
 - Use `db/seeds.rb` for data that is required in all environments.
 - Use `dev:prime` rake task for development environment seed data.
 
-## Console
+## Security
 
-- Access production-like [console][] sessions in "sandbox" mode: `bin/rails console --sandbox`
+- Set [config.sandbox_by_default][sandbox] to `true` in production-like environments to avoid accidental writing to the production database.
 
-[console]: https://guides.rubyonrails.org/command_line.html#bin-rails-console
+[sandbox]: https://guides.rubyonrails.org/configuring.html#config-sandbox-by-default
 
 ## Migrations
 


brian-penguin:
Mixed on this. In my perfect world the console is always sandboxed and no one attempts to change or alter data. In a world in which I regularly live, stuff changes from the console. Being in sandbox mode and attempting to change records will lock the DB rows until the session is ended and the transaction is rolled back, which can cause problems and potentially crash. It also doesn't prevent side effect calls (changing job states in redis/ external api calls/ etc) made during the session. 

In a lot of applications, there just aren't the tools available to debug, change config, or run code at a specific enough level. Building those tools only usually happens on large teams which have non-technical admin who are taking responsibility for fixing those problems for their users. 

So yeah, torn on the outcome here and not sure there's "generalizable" advice. I think maybe if you had to have console access because you didn't have the other tools for debug-ability (or things like deploying those tools takes way too long) I would push for an audit log tagged with the user, if that's possible. You could add an initializer with a custom logger and an ActiveSupport::Notification subscription for events that change the DB (and maybe custom ones for external side-effects or job enqueueing)? 

```ruby
# config/initializers/console_audit.rb
if defined?(Rails::Console)
  console_audit_logger = ActiveSupport::TaggedLogging.new(
    Logger.new(Rails.root.join("log/console_audit.log"))
  )

  ActiveSupport::Notifications.subscribe(/^sql\.active_record$/) do |*args|
    event = ActiveSupport::Notifications::Event.new(*args)

    if /\A(INSERT|UPDATE|DELETE)/i.match?(event.payload[:sql])
      username = `whoami`.strip # am lazy, could make them say the user or give a token or something

      console_audit_logger.tagged("CONSOLE_AUDIT", username) do
        console_audit_logger.info(event.payload[:sql].gsub(/\s+/, " "))
      end
    end
  end

  # Write the notice out to the console
  puts "Console audit logging enabled. All changes will be recorded in log/console_audit.log"
end
```

brian-penguin:
I know basecamp has some tools for this I've never used 
https://github.com/basecamp/console1984
https://github.com/basecamp/audits1984

rails/README.md
```diff
@@ -81,6 +81,12 @@ Guidance on ActiveRecord, ActiveModel, and other model objects.
 - Use `db/seeds.rb` for data that is required in all environments.
 - Use `dev:prime` rake task for development environment seed data.
 
+## Console
+
+- Access production-like [console][] sessions in "sandbox" mode: `bin/rails console --sandbox`
```
stevepolitodesign, line 86: @brian-penguin would it help if we changed the framing from "don't change data in production" to "exercise caution when using the rails console in production... you wouldn't wanna delete something on accident"?

```suggestion
- Access production-like [console][] sessions in "sandbox" mode when possible: `bin/rails console --sandbox`
```

stevepolitodesign: The spirit of this guidance is to highlight that this is even an option.

nickcharlton: I love the spirit of this, but I share the same thoughts as @brian-penguin. But I do think just pointing towards better approaches is a nice way to do this here.

stevepolitodesign: @nickcharlton @brian-penguin 

What do you think about the changes proposed in 4e4e6bd3e369dddcbbfe66e62e51a1f26e0ed12d?

This way, you need to explicitly pass `--no-sandbox` when entering the Rails console.

nickcharlton: Oh yeah, interesting. I'm a fan of that!

brian-penguin: I think that's a better default!

stevepolitodesign: @nickcharlton / @brian-penguin do you think this looks good to merge?

neilvcarvalho (approved):
I like that as a default.

PR #748, opened 2025-04-04
Rails: Prefer webmock over VCR

stevepolitodesign:
Cassettes are tricky to maintain and are often fragile. Using webmock
allows for more fine-grained control.

Additionally, VCR risks leaking [sensitive data][].

[sensitive data]: https://benoittgt.github.io/vcr/#/configuration/filter_sensitive_data

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -146,6 +146,13 @@ others' Rails work, look in particular for:
 
 [ActiveStorage]: https://guides.rubyonrails.org/active_storage_overview.html
 
+## Testing
+
+- Prefer [webmock][] over [VCR][].
+
+[webmock]: https://github.com/webmock/webmock
+[VCR]: https://github.com/vcr/vcr
+
 ## How To Guides
 
 - [Start a New Rails App](./how-to/start_a_new_rails_app.md)


DoodlingDev:
friends don't let friends use VCR

stevepolitodesign:
Note to self: I want to add my context to this decision, but I need to find the right words.

brian-penguin (approved):
I am against VCR (in the case of 90% of rails integration/unit style specs): 
- When it requires at least one "real" request so someone somewhere needs to hit a production-ish service with real-ish data and if you don't have an accurate sandbox, you just hope that it doesn't matter
- When it's non-specific, what data are we stubbing exactly? Which peices do we care about testing for? Are we sure we are sending the right headers / tokens / full url path? 
- When time matters. If the response came at 9pm on January 5th 2023 and that's the timestamps on the data in the VCR but I have a scope to look within the last month that I'm testing, does that mean I need to hand edit the yaml? Re-run the VCR generation every month?
- A well built mock for a single request is clear, it exists alongside the spec, and it lets you be specific for every bit of data that goes into it and what you expect out.
- Noisy! Re-running the VCR request likely rebuilds the entire file and it's hard to track what, if anything, actually changed
- Sensitive Data. If you're hitting a realish api, you've probably got real api keys/tokens/ids all of which should be filtered out

I am for VCR :
- when you are building a Mock for an external service you control. If I have the microservice to use sitting in the dir next to mine and I want it to respond to a consistent set of data consistently across multiple requests, perfect use case! When the api updates in the microservice, pull the change re-gen the cassette. 
- In the past I've also seen a like "fake" server implementation where your tests all hit a real sinatra app that's faking a real api and I think I would prefer VCR over that unless the api is real simple 
- When you're not using it in specs! Trying to build out an integration but don't the api is down regularly, or inconsistent, or whatever?

JoelQ:
@purinkle I think this intersects with your talk proposal!

jutonz (approved):

PR #750, opened 2025-04-18
Document Examples of using GitHub Alerts (Obsidian Callouts)

vburzynski:
- Adds a template example of using the GitHub Alert or Obsidian Callout Syntax
- Documents the discrepancies between the two implementations


_template/how-to/.DS_Store
JoelQ, line 1: This file should not be committed

JoelQ (commented):
The PR descriptions explains the _what_, but I'd love to see a bit more about the _why_. Can you explain why you think this is needed as part of the guides repo? Are you trying to create some kind of style guide for this repo? For all thoughtbot work on GitHub?

If this is more of a how-to article, maybe it's a better fit for something like the blog?

_template/how-to/.DS_Store
vburzynski: Doh, I always expect those to be git ignore. I'll add that

thoughtbot-summer: You can create a [`~/.config/git/ignore`](https://docs.github.com/en/get-started/git-basics/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) file ([example](https://github.com/smmr0/dotfiles/blob/3f364856b8deae0853734932ca64db3ed2635c61/.config/git/ignore)) (path configurable via [`core.excludesFile`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreexcludesFile)) to automatically ignore it for every repository! 😃

vburzynski:
Not much interest or activity here

PR #752, opened 2025-04-25
A proposal for guideline pages

vburzynski:
A proposal for structuring sub-pages that provide additional details to guidelines.

### DIFF
--- a/_template/Guide To Writing Guideline Pages.md
+++ b/_template/Guide To Writing Guideline Pages.md
@@ -1,0 +1,100 @@
+# Guide To Writing Guideline Pages
+
+## What are guideline pages?
+
+Guideline pages are comprised of the set of pages linked to from the outline-mode pages and which provide additional information and reasoning behind a guideline. The outline-mode pages are those pages which collate lists of guidelines under a specific topic.
+
+A guideline page is a brief synopsis of a convention, standard, heuristic, best practice, or stylistic choice. It may cover the intent, influences, motivation, applicability, consequences, implementation or other particulars behind a guideline. It is not meant to be a comprehensive analysis; rather it should highlight of the most relevant details.
+
+## Why promote a consistent format?
+
+This *guide to writing guideline pages* suggests an informal simple formatted structure. The aim is to promote a small amount of consistency through format and layout while still being oriented towards flexibility. The content can be conversational and expresive through freeform text, while being presented in a familiar way. This consisitency can aid readers in their efforts to navigate the site because they can anticipate what they expect to find on any given page. When presented with information in a nonconventional or atypical way, they'll have to spend more time figuring out what they are looking at.
+
+The selected structure is influenced by the pattern forms explored and discussed across various wikis such as WikiWikiWeb and resources such as the *Design Patterns* book by the Gang of Four. The format is largely borrowed from the Portland Form, also known as "Therefore But" due to it's construction: You summarize the context and forces in play, and then connect that with a bold "therefore" to the guideline. You can then optionally follow that with any constrasting details.
+
+> [!TIP]
+> The guidelines are not formal design patterns, rather they are helpful principles and concepts that are not always well defined. With that in mind, feel free to take liberties with the suggested format. Furthermore, not every guideline requires further explanation.
+
+## Guidelines
+
+- The **file name** should match the **title** of the page exactly
+  - as a result, titles must only use characters that are valid for use in a file name
+- The **title** should be a *proposition* (or claim) which concisely states the guideline
+  - Titles should not be in interrogative form (such a question)
+  - Titles should not be in lexical form (such as a noun or noun phrase)
+- **Guideline Section**
+  - Prefer a **guideline section** that is short and concise but not terse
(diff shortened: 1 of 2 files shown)


_template/Guide To Writing Guideline Pages.md
```diff
@@ -0,0 +1,100 @@
...
+
+The selected structure is influenced by the pattern forms explored and discussed across various wikis such as WikiWikiWeb and resources such as the *Design Patterns* book by the Gang of Four. The format is largely borrowed from the Portland Form, also known as "Therefore But" due to it's construction: You summarize the context and forces in play, and then connect that with a bold "therefore" to the guideline. You can then optionally follow that with any constrasting details.
+
+> [!TIP]
+> The guidelines are not formal design patterns, rather they are helpful principles and concepts that are not always well defined. With that in mind, feel free to take liberties with the suggested format. Furthermore, not every guideline requires further explanation.
+
+## Guidelines
```
MatheusRich, line 18: If we accept these meta-guidelines, we should use a linter of sorts to enforce them (at least the ones we can)

_template/Guide To Writing Guideline Pages.md
```diff
@@ -0,0 +1,100 @@
...
+  - Prefer a **guideline section** that is short and concise but not terse
+  - The first part of this section sets the stage by introducing the problem, context, and forces in play
+  - The conflict is resolved by presenting the solution or guideline along with a discussion of the resulting context.
+  - Then other considerations or counterindications can be reviewed.
+  - It then ends with a summary and discussion of the greater context (to connect back to the broader scope).
+- Additional **headings** and sections can be added between the `## Guideline` and `## References` sections
+  - Prefer repeating patterns found in other guides, but there's no conventions set here yet.
+- A **reference** section includes links to resources and pull-requests
+
+## Page Template
```
MatheusRich, line 35: Once we have one, a real example would be good

_template/Guide To Writing Guideline Pages.md
```diff
@@ -0,0 +1,100 @@
+# Guide To Writing Guideline Pages
+
+## What are guideline pages?
```
MatheusRich, line 3: Should *all* guidelines have a page?

_template/Guide To Writing Guideline Pages.md
vburzynski: That would be a useful tool to have 👍 
I could see the headings being an easy one to enforce.

_template/Guide To Writing Guideline Pages.md
vburzynski: here's a WIP example - https://github.com/thoughtbot/guides/compare/main...vb-modifying-nme-standard

_template/Guide To Writing Guideline Pages.md
vburzynski: 🤔 That's a great question...

- initially I think it'd be unreasonable to enforce this for all existing guidelines. I wouldn't want to rapidly generate a bunch of pages. If we went that direction, I think it should be thoughtful and deliberate. 
- I do think that there are a lot of existing guidelines that could benefit from a revisit and adding a guideline page to flesh out our thoughts on why it's useful and important to follow.
- Having reasoning behind every guideline, even ones that appear obvious, can make them more approachable to everyone and to every level of experience. The work that goes into articulating the reasoning behind a guideline can facilitate analysis of assumptions; avoid miscommunication or confusion; clarify one's thinking; and reveal gaps in our own understanding.
- I could see enforcing this for new guidelines. It can encourage discussion and preserve more context for the next time someone reviews a guideline to see if it's still as relevant and applicable as it once was.

_template/Guide To Writing Guideline Pages.md
```diff
@@ -0,0 +1,100 @@
...
+- Additional **headings** and sections can be added between the `## Guideline` and `## References` sections
+  - Prefer repeating patterns found in other guides, but there's no conventions set here yet.
+- A **reference** section includes links to resources and pull-requests
+
+## Page Template
+
+```markdown
+# Use a prepositional pharse in sentence case for titles
+
+## Guideline
```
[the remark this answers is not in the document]
vburzynski: I've imagined this being more of a starting point to underscore the angles from which we might examine a guideline and to identify the components which could be put into an articulated reasoning. This is more to encourage a bit of internal consistency of approach. The emphasis is not on following a specific structure.

- There's virtue to irregularity. I'd like to encourage common components, but highly encourage creativity and flexibility in how that is composed and presented.
  - https://www.eastgate.com/garden/Virtue_of_Irregularity.html
- at the same time, repetition is a valuable cue
  - https://www.eastgate.com/garden/Repetition.html
- So an effective strategy is the artful combination of irregularity with some order to signal intent. Building more of a garden that composes structural elements with surprise and experimentation. Avoiding the sterility and rigidity of farmland while also avoiding untamed wilderness.
  - https://www.eastgate.com/garden/Gardens.html
  
Also things that are entirely unsurprising yield no information (surprisal being a measure for the amount of information exchanged in a communication between two entities).

So yes, I would love to explore more formats

PR #757, opened 2025-05-09
fix: correct broken links in Ruby guide

vburzynski:
So, looks like GitHub's Markdown parser requires URL encoding for Markdown Links...


MatheusRich:
@vburzynski what about using kebab-case or snake case for file names?

MatheusRich (approved):

PR #758, opened 2025-05-09
Remove SECURITY.md

neilvcarvalho:
Our open source projects include a SECURITY.md file with instructions about supported versions and how to report a vulnerability. While this is an open source project, it's not a runnable app, but a collection of documents.

Recently, as a Markdown linting workflow was added to this project, it started competing with another workflow that keeps SECURITY.md up-to-date on how this file should be formatted.

This commit removes both SECURITY.md and the `update-security` GitHub Actions workflow.


cpytel (approved):

vburzynski (approved):
Makes sense to remove this from the guides repo 👍
There may still be value in removing the extraneous lines in the templates to avoid re-encountering this issue in the future should one of the other repositories choose to add markdown linting.

neilvcarvalho:
For some reason, I had SECURITY.md in mind, but the template changes are currently affecting README.md instead. Anyway, I still think it made sense here.

PR #764, opened 2025-07-25
feat: define thoughtbot rules for AI-enabled IDEs

laicuRoot:
During a Fusion AI task force meeting, we discussed the tools we use and how we enforce rules. As part of that conversation, we decided it would be valuable to define official thoughtbot guidelines for editors that integrate AI, such as Cursor.

This PR opens the discussion by proposing an initial set of rules. This guide is a starting point for how we write and structure Rails apps at thoughtbot. It’s intended as a starting point to ensure consistency, readability, and shared best practices across our codebases. The guide is a living document it.

### DIFF
--- a/rails/ai-rules/rails-rules.md
+++ b/rails/ai-rules/rails-rules.md
@@ -4,7 +4,7 @@ You are an expert in Ruby on Rails, PostgreSQL, and Hotwire (Turbo and Stimulus)
 
 ## Key Conventions
 
-- Follow RESTful routing conventions: Seven restful actions: index, show, new, create, edit, update, delete (https://thoughtbot.com/blog/in-relentless-pursuit-of-rest-ish-routing)
+- Follow RESTful routing conventions: Seven restful actions: index, show, new, create, edit, update, delete ([thoughtbot post REST-ish routing guide](https://thoughtbot.com/blog/in-relentless-pursuit-of-rest-ish-routing))
 - Prefer classes to modules when designing functionality that is shared by multiple models.
 
 ## Data / Models
@@ -20,14 +20,12 @@ You are an expert in Ruby on Rails, PostgreSQL, and Hotwire (Turbo and Stimulus)
 
 - Always check for N+1 queries when rendering collections
 - Prefer includes for eager loading
-- Scope queries to only the fields needed with select
 
 ## Testing
 
 - Always write tests to cover new code generated
 - Use RSpec for testing framework, unless the project already uses minitest
-- Use factories (FactoryBot) (https://thoughtbot.github.io/factory_bot/)
+- Use factories ([FactoryBot](https://thoughtbot.github.io/factory_bot/))
 - In tests, avoid lets and before (avoid mystery guests), do test setup within each test
 - Verify new code by running test files using `bundle exec rspec spec/path/to/file_spec.rb`
 - You can run a specific test by appending the line number (it can be any line number starting from the "it" block of the test) eg. `bundle exec rspec spec/path/to/file_spec.rb:72`
-  


rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,71 @@
...
+Performance Optimization
+- Use database indexing effectively.
+- Implement caching strategies (fragment caching, Russian Doll caching).
+- Use eager loading to avoid N+1 queries.
+- Optimize database queries using includes, joins, or select.
+
+Key Conventions
+- Follow RESTful routing conventions.
+- Use concerns for shared behavior across models or controllers.
+- Implement service objects for complex business logic.
```
laicuRoot, line 45: I'm not 100% sure about this for example. Lately I've been more leaning towards [Service objects are poorly-named models](https://dimiterpetrov.com/blog/service-objects-are-poorly-named-models/).

jaredlt:
@laicuRoot thanks for starting this! I'm curious if you added each line as a consequence of the AI performing poorly on something? Or did you add things from our existing guides that seemed useful?

Some of these things eg. naming conventions I haven't had any trouble with (I'm using Claude Code). I think it would be good if we can avoid adding things that don't improve the situation so we have a list that is maintainable.

Based on my Claude Code usage and observing when it doesn't perform so well I've added the following to my `CLAUDE.md`

```
# Claude memory

## Testing

- Always write tests to cover new code generated
- Prefer RSpec for Rails applications but use the existing test framework if there is one
- In tests, avoid lets and before (avoid mystery guests), do test setup within each test
- Verify new code by running test files using `bundle exec rspec spec/path/to/file_spec.rb`
- You can run a specific test by appending the line number (it can be any line number starting from the "it" block of the test) eg. `bundle exec rspec spec/path/to/file_spec.rb:72`

## Data / Models

- To find model structure look in `db/schema.rb`. The file is big, you'll need to grep
- When working with model attributes don’t guess, grep the schema at `db/schema.rb` to confirm and use only valid attributes

## Commit messages

- Use a loose paragraph format of
  - <Title>
  - Before, <current state context>
  - <Explain the problem (sometimes the before and problem are the same paragraph, other times it is clearer to separate the current state from the problem)>
  - Now, <what's changed and why / the impact>
- Avoid overexplaining individual commits and prefer to explain the overall branch (the problem solved)
- Never mention Claude/AI was used in commit messages eg. do not show co-authored in git commits

```

Some things here, mostly around the commit message guidance is too personal / prescriptive and don't think we'd want that in shared rules.

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,71 @@
...
+- Use CamelCase for class and module names.
+- Follow Rails naming conventions for models, controllers, and views.
+
+Ruby and Rails Usage
+- Use Ruby 3.x features when appropriate (e.g., pattern matching, endless methods).
+- Leverage Rails' built-in helpers and methods.
+- Use ActiveRecord effectively for database operations.
+
+Syntax and Formatting
+- Follow the Ruby Style Guide (https://rubystyle.guide/)
```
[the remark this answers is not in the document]
laicuRoot: I didn't know about hooks 🤔 thank you for sharing this. I guess we can include hooks within the guides?

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,71 @@
...
+
+Key Conventions
+- Follow RESTful routing conventions.
+- Use concerns for shared behavior across models or controllers.
+- Implement service objects for complex business logic.
+- Use background jobs (e.g., Sidekiq) for time-consuming tasks.
+
+Testing
+- Write comprehensive tests using RSpec.
+- Follow TDD/BDD practices.
```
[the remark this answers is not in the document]
JoelQ: I have gotten LLMs to do true red/green/refactor TDD. This has been with one-off agent sessions in my editor (both Claude and GPT) with an initial prompt rather than extracting to a rule though.

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,71 @@
...
+- Use factories (FactoryBot) for test data generation.
```
JoelQ, line 51: Is it worth suggesting that test setup should be local? Tell the AI to optimize for readability and changeability? Otherwise I find that LLMs love to create deeply nested contexts with `let`.

vburzynski: I think it might be worth addressing `let` statements and preferring an approach of keeping test cases self-contained rather than keeping the spec file DRY.

laicuRoot:
I’ve pushed an update that keeps only the suggestions and rules based on my experience using Cursor. I like @jaredlt approach of keeping it concise, focusing just on instructions to correct behaviours we don’t want when using the AI as a pair programming partner 🙂

JoelQ:
How is this mean to be used? Are you expected to copy-paste these into your rule file of choice?

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,33 @@
...
+
+## UI and Styling
+
+- Use Rails view helpers and partials to keep views DRY
+
+## Performance Optimization
+
+- Always check for N+1 queries when rendering collections
+- Prefer includes for eager loading
+- Scope queries to only the fields needed with select
```
MatheusRich, line 23: Does this mean that it will add `.select` to all queries? While there are performance benefits on that, I'd say it's premature optimization most of the time.

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Testing
+
+- Always write tests to cover new code generated
+- Use RSpec for testing framework
```
MatheusRich, line 28:
```suggestion
- Use RSpec for testing framework, unless the project already uses minitest.
```

MatheusRich (approved):
This seems like a good start.

I wonder what would it happen if we feed an LLM with the whole guides repo and ask it to generate the rails-rules.

laicuRoot:
> I wonder if perhaps that is the recommendation here? We recommend feeding these guides into an LLM?

> @laicuRoot what do you think?

I think we could start small as this PR proposes and then iterate considering the information you have shared of our guides digested by LLMs. How do you feel about that @samithoughtbot ? Do you think is a good starting point for this?

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Performance Optimization
+
+- Always check for N+1 queries when rendering collections
+- Prefer includes for eager loading
+- Scope queries to only the fields needed with select
+
+## Testing
+
+- Always write tests to cover new code generated
```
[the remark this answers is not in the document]
laicuRoot: I have struggled with this one personally when working with Cursor. Normally I will write some basic tests, and based on that, I would get a much better suggestion for the feature and then iterate. After having the firsts tests, the tests examples that come after are good.

@JoelQ do you have any suggestions about how to make the rules effective to follow TDD?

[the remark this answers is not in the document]
JoelQ: I've usually done that with an initial prompt. Zed's agent (using Claude) does this well.

MatheusRich:
Some I have on my system prompts is to write tests without `let`/`subject` et al and to use the 3-phase testing pattern (setup, exercise, verify). That gives me tests that much closer to the ones I'd write myself.

laicuRoot:
> Has anyone tested these? It feels like we're just throwing random stuff in a PR. Normally things go into the guides after after we've experimented with them and come up with an agreed-upon best practice.
> 
> I'd love to get some actual experience reports for rules. Some things we should probably capture along with the experience:
> 
> 1. Which agent and which base model? (e.g. I use Zed with Claude as my main agent right now)
> 2. What style of project did you use this on (existing project, greenfield, testing style, etc)
> 3. What rules did the agent not follow? What rules did the agent not need because it knows the use them by default?

I’ve been using this set of rules with Cursor and Claude for a while now in an existing project. One place where they’re particularly useful is when generating tests: if the rules aren’t attached, or if I don’t explicitly state in the prompt how the tests should be written, the output often includes let or before blocks by default.

Similar case when creating new routes, based in my experience, unless rules are attached, it does not follow the RESTful Rails way. This I've seen it in existing and new projects.

In personal projects I've found very helpful to indicate that while dealing with views always consider using helpers and partials, otherwise I've seen that sometimes it would throw everything in the view. This morning for example, in a breakable toy, it prepared data within the view, instead of the Controller. 

Since I set the rules to `always attached` in Cursor in the client project, it consistently follows them. I’ll test without the rules again and share here what the default behaviour looks like out of the box.

It would be good to hear other people experience. I agree that it would be good to have more reports using this set of rules.

rails/ai-rules/rails-rules.md
```diff
@@ -0,0 +1,31 @@
...
+- To find model structure look in `db/schema.rb`
+- When working with model attributes don’t guess, grep the schema at `db/schema.rb` to confirm and use only valid attributes
+
+## UI and Styling
+
+- Use Rails view helpers and partials to keep views DRY
+
+## Performance Optimization
+
+- Always check for N+1 queries when rendering collections
```
louis-antonopoulos: I would remove or change this line. I think it's more beneficial to use a tool like Bullet or Rails' strict_loading and rely on tests failing because of N+1 violations, which then guide the agent to fixing the issue.

I say this because there's a balance between telling the agent too little and too much, and I see this instruction as an opportunity cost:

"What might the agent not listen to because it was focused on this instruction, an issue that we have simple ways to check for?"

So while I would remove it, if kept, maybe something like

```suggestion
- Avoid N+1 queries when rendering collections
```

jaredlt:
@iftheshoefritz I don't think AGENTS.md is officially supported in Claude yet but the common workaround seems to be

```
# In ./CLAUDE.md
@AGENTS.md
```

JoelQ:
@laicuRoot is the intention of #783 to be an alternative to this? Or is it in addition?

gokcelb:
Should we close this PR since I think it's outdated and is being effectively replaced by https://github.com/thoughtbot/guides/pull/783?

PR #765, opened 2025-08-08
Chunk Ruby guides by theme

MatheusRich:
I tried to extract some themes out of the Ruby guides and chunked a few chunks:

- Stuff I didn't touch
- method definitions/signature
- variables/params


MatheusRich:
@vburzynski that's exactly what I aimed for. I considered adding headings too, but that can be a separate step, maybe.

PR #768, opened 2025-08-29
Be cautious about information from unofficial sources

MatheusRich:
Clarify if you haven't tested a code suggestion or got an information from an unofficial source like an LLM or blog post.


sallyhall (approved):
Should we incorporate something specifically about AI? Clarify that AI-generated code or information that you haven't verified should also be disclaimed.

MatheusRich:
@neilvcarvalho noticed this guideline already existed (down below in this guide), so I updated it just to be a bit more explicit.

PR #783, opened 2026-02-18
thoughtbot agents rules

laicuRoot:
## Summary

This proposed change for the rules is based on what I've been reading about the CLAUDE.md/agents.md file, rules, skills in their own documentation and my own testing in client and personal projects with good results.

I think these are good defaults that you can add at the start of any project, or even projects that have been running for a while. Obviously each of us has observed different behaviour from our interactions with the models because all our prompts are different. Some of you might feel these rules are a bit too detailed and that's totally fine, but they have worked well across different projects for me. I'd love to hear what has been working for you!

## What does good results mean? 

- Reduce the need of prompting that we should follow TDD
- Continuously applying the rules
- Despite having other patterns or architecture, flag those as legacy and apply patterns described in rules

## Files included

- Project memory (CLAUDE.md or agents.md): Project architecture, link to rules. Source controlled. 
- Project rules: Language-specific guidelines, testing conventions, and other standards. Source controlled.

## Why not skills? 

Skills are supposed to be invoked on demand. CLAUDE.md and rules should be for every session. We should keep CLAUDE.md and rules under 500 lines.

## Why this project-memory.md format?

The `project-memory.md` template is designed as a **project context file** that describe the stack, list common commands and links to the detail rules.

See also "[give Codex a map, not a 1,000-page instruction manual.](https://openai.com/index/harness-engineering/)"

From Claude documentation chatbot: 
```
Based on the search results, the documentation states:

Keep CLAUDE.md under ~500 lines by including only essentials. This is the ideal length.

The documentation says: "Aim to keep CLAUDE.md under ~500 lines by including only essentials." This applies to CLAUDE.md specifically, and the guidance is to move specialized instructions into skills to keep your base context smaller.

For rules (.claude/rules/*.md), there's no specific line limit mentioned, but since they're also loaded automatically at session start, the same principle applies: keep them concise and focused.*
```

## Test plan

As mentioned above I've been testing this in client and personal projects and I'm happy with the behaviour (with modifications to meet the application differences). 

- [ ] Copy template to a Rails project's `.claude/CLAUDE.md` and make the necessary modifications based on your project
- [ ] Copy `rules/` folder to `.claude/rules/`
- [ ] Verify Claude Code follows thoughtbot conventions when generating code


rails/ai-rules/rules/database.md
```diff
@@ -0,0 +1,10 @@
+# Database & Migrations
+
+- Every foreign key must have an index. Every `where` column should have an index.
+- Always eager-load associations when iterating: `includes`, `preload`, or `eager_load`.
+- Migrations must be reversible. Never modify data in structure migrations.
```
jaredlt: > Never modify data in structure migrations

This bit reads strange to me. I think it's trying to distinguish "structure"/schema migrations from data migrations? I think of all migrations as structure migrations. And consider data changes via rake tasks etc. But I'm not sure what our overall thoughtbot beliefs are here.

rails/ai-rules/rules/models.md
```diff
@@ -0,0 +1,10 @@
+# Models & Domain Objects
+
+- No service objects. All domain classes live in `app/models/` with namespaces, never `app/services/`.
+- Name classes after domain nouns, not actions. No `*Service`, `*Manager`, `*Handler` suffixes.
+- Use `ActiveModel::Model` for POROs that need validation or form integration.
+- Replace `.call` / `.perform` with domain verbs: `#save`, `#complete`, `#submit`, `#deliver`.
+- Extract from ActiveRecord models when they exceed: 200 lines, 15 public methods, or 7 private methods.
```
jaredlt: Is this just for ActiveRecord models or could we remove ActiveRecord and just say models?

Could we say something a bit less resolute here? Maybe:

Look to identify domain models that can be extracted when an existing model exceeds: 200 lines, 15 public methods, or 7 private methods

rails/ai-rules/rules/testing.md
```diff
@@ -0,0 +1,13 @@
+# Testing
+
+- You MUST not use `let` or `before` in specs.
+- Test behaviour, not implementation. Four Phase Test: setup, exercise, verify, teardown.
+- Test pyramid: many model/PORO unit specs, some request specs, few system specs.
+- Every public method on every model and PORO must have at least one spec.
```
jaredlt: What do you think about something similar to this one but talking about testing logical branches? eg. if there's an if / else we typically want to test both

jaredlt (approved):
LOVE this! 🔥 

Left a few comments but I think this looks great and something we can start testing and iterating on over time.

rails/ai-rules/rules/testing.md
```diff
@@ -0,0 +1,13 @@
+# Testing
```
jaredlt: We don't mention TDD in this file? Just in the reference in project-memory.md. I typically say something like:

- Must use TDD. Write tests first and follow red, green, refactor

rails/ai-rules/rules/controllers.md
```diff
@@ -0,0 +1,10 @@
+# Controllers
+
+- Controllers handle HTTP only: receive request, delegate to model, return response.
+- Actions must not exceed 10 lines (excluding strong params).
+- Maximum one instance variable per action.
+- No business logic, calculations, email sending, or multi-object operations in controllers.
+- Always use strong parameters. Never `params.permit!`.
+- Return `status: :unprocessable_entity` on failed form renders (required by Turbo).
+- Prefer RESTful routes. Custom actions often mean a missing resource — e.g., `POST /order_cancellations` not `POST /orders/cancel`.
+- Use `before_action` for auth. Be explicit with `only:` / `except:`.
```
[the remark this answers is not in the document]
laicuRoot: I found it useful on personal projects where most routes are public and auth only applies to a few specific areas. Claude kept forgetting to add it for those cases. But you're right, on a real project we would set `before_action :authenticate_user!` in ApplicationController and skip it where needed. I'll remove it from the rules. Thank you!!

gokcelb (commented):
In general, I really like this setup of the rules files 💯 I found that my client has a similar approach as well. 

One thing I'm not sure about is mentioning 'suspenders' or not, could that be confusing for the agent and something extra to think about. I feel like in a mature project, it doesn't make sense that the agent is searching for 'suspenders' and trying to understand what it is. 

We don't have to answer this quesiton now but I was also wondering how should we flag the lines in specific rules files mentioning technologies that may or may not be used in client projects, e.g. Turbo or Sidekiq?

rails/ai-rules/rules/database.md
```diff
@@ -0,0 +1,11 @@
+# Database & Migrations
+
+- Always use the rails generate migration command to create migration files.
```
cpytel:
```suggestion
- Always use the `rails generate migration` command to create migration files.
```

laicuRoot:
Generally I found that making the LLM to enforce TDD without explicitly saying so in the prompt was difficult but it seems that with the rules in place is much better. Leaving another example in plan mode from some testing I did today.

My prompt was just: `let's start with the recipe model`
 
Screenshot 2026-02-20 at 11 47 20
image
image

rails/ai-rules/rules/views.md
```diff
@@ -0,0 +1,8 @@
+# Views & Presenters
+
+- Views render data. No calculations, queries, or complex conditionals.
+- Use presenters to display logic. Instantiate in controller, use in view.
+- Extract repeated markup into partials. Pass data via `locals:`, not instance variables.
+- Helpers for simple formatting only (dates, currencies). If longer than 5 lines, use a presenter.
+- Turbo: return `status: :unprocessable_entity` on failed forms. Keep Stimulus controllers small.
+- Suspenders provides `_flashes.html.erb` and `_form_errors.html.erb` partials — use them.
```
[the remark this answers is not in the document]
laicuRoot: Yes, I'm also not 100% sure about including Suspenders remarks but as we are thoughtbot, and these are our guides, I thought it made sense. I start any new project with using `suspenders new`. 

Folks can always delete what is not relevant for them. What do you think?

rails/ai-rules/rules/testing.md
JoelQ: The terms here are "line coverage" and "branch coverage" respectively

JoelQ (approved):
Go ahead and merge! I think this is a strong base we can iterate from

rails/ai-rules/rules/views.md
```diff
@@ -0,0 +1,7 @@
+# Views & Presenters
+
+- Views render data. No calculations, queries, or complex conditionals.
+- Use presenters to display logic. Instantiate in controller, use in view.
+- Extract repeated markup into partials. Pass data via `locals:`, not instance variables.
+- Helpers for simple formatting only (dates, currencies). If longer than 5 lines, use a presenter.
+- Turbo: return `status: :unprocessable_entity` on failed forms. Keep Stimulus controllers small.
```
louis-antonopoulos, line 7: Rubocop's [Rails/HttpStatusNameConsistency](https://docs.rubocop.org/rubocop-rails/latest/cops_rails.html#railshttpstatusnameconsistency) recommends usage of `:unprocessable_content`:

```suggestion
- Turbo: return `status: :unprocessable_content` on failed forms. Keep Stimulus controllers small.
```

PR #786, opened 2026-03-31
Add cursor and codex rails ai rules

gokcelb:
Before, we only had Rails AI rules for Claude. This PR adds them for Codex and Cursor as well.


jaredlt:
I can see this getting pretty gnarly to manage as we iteratively make updates. Should we look to make it generic?

Differences I see that we may need to validate:

- CLAUDE.md renamed verbatim to AGENTS.md for Codex - will it follow the rules links that the CLAUDE file specifies? I see Codex has something called [rules](https://developers.openai.com/codex/rules) which is different, but I do wonder if it would "just work"?
- Cursor files are .mdc? Again, would it just work if we referenced the .md files or does it absolutely need to be mdc?

laicuRoot:
Hey folks! 

I think this is worth pursuing but it needs more research. 

Unless I hear otherwise, I will close this PR and move the task associated with it to the next column in the trello board for someone else to pick it up that is interested in this work. 

Thanks.

PR #788, opened 2026-04-15
Add PII storage and PHI specification under Guides > Security > Application

KurodaSteph:
Ref. https://hub.thoughtbot.com/tickets/8715

We are connecting Anthropic Claude Team account to Google Drive, Calendar, Slack, and Figma. Some PII like emails related to sales may live in Google Drive.

While this is not Handbook, in Guides, we want to be explicit that: 

 but we want to be explicit that:

> When you must store PII:

> - Store sensitive data in controlled systems with appropriate access restrictions and encryption. In particular, Protected Health Information (PHI) should not be stored in broad, shared tools such as Google Drive, Slack, or Figma.

While this PR is for Guides, we have a corresponding PR in Handbook where this Guides guidance is a Handbook policy: https://github.com/thoughtbot/handbook/pull/4422

### DIFF
--- a/security/application.md
+++ b/security/application.md
@@ -146,7 +146,7 @@ As much as you can, do not touch any information you don't need.
 
 When you must store PII:
 
-- Store sensitive data in controlled systems with appropriate access restrictions and encryption. In particular, Protected Health Information (PHI) should not be stored in broad, shared tools such as Google Drive, Slack, or Figma.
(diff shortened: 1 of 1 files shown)


JoelQ (approved):

KurodaSteph:
Thanks, @JoelQ!

PR #794, opened 2026-05-22
Update css guides

enatario:
Our CSS guides were pretty stale and not comprehensive. This updates them to include more modern approaches, including recommending native CSS over SCSS (this removes the Sass guides). This also leverages Roux as a baseline for best practices.


css/README.md
```diff
@@ -2,113 +2,370 @@
...
 [double colon syntax]: https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#Syntax
+[popover api]: https://developer.mozilla.org/en-US/docs/Web/API/Popover_API
 
-## Linting
+## Getting started
 
-- Use stylelint to lint CSS & Sass
-  - We maintain a [sharable stylelint configuration] which enforces our guides
-    found in this repo
+For new projects, consider using [Roux] as a structured starting point. It
```
vendelal, line 19: ❤️

css/README.md
```diff
@@ -2,113 +2,370 @@
...
-<details>
+Raw `@import` of CSS will render your CSS, but it's not recommended as a browser will load each `@import` as a separate network request.
 
-#### Code examples
+Use your preferred build tool to bundle CSS into a single output. Options
+may include:
 
-`h1.page-title` carries a specificity of 2, but can be reduced to 1 by removing
-the `h1` type selector:
+- [Lightning CSS] is fast, modern, and light
```
vendelal, line 78: Do these tools have good longevity? This isn't inherently an issue, but just wondering if this landscape changes frequently and whether we want to include specific links or not

css/README.md
```diff
@@ -2,113 +2,368 @@
...
 
-Using an ID in a selector increases its specificity, making it more difficult to
-work with alongside class selectors. Furthermore, because IDs must be unique
-within an HTML document, using them as CSS selectors limits reusability.
+## Responsive design
 
-#### Resources
+Modern CSS layout tools can often eliminate the need for breakpoints entirely.
+Reaching for a media query is reasonable, but it's worth considering whether the
+layout could respond naturally first.
```
vendelal, line 132: This is honestly kind of news to me 😅 I just saw you're including some examples below, that's really helpful!

css/README.md
```diff
@@ -2,113 +2,368 @@
...
+    display: grid;
+    grid-template-columns: auto 1fr;
+  }
 }
 ```
 
-Don't concatenate selector names:
+### Media queries (for larger layout directives)
+
+Media queries can still the right tool for page-level layout changes: things like large
```
vendelal:
```suggestion
Media queries can still be the right tool for page-level layout changes: things like large
```

css/README.md
```diff
@@ -2,113 +2,368 @@
...
 
-```scss
+[container queries]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries
+
+## Linting
+
+[Stylelint] is a good option for enforcing CSS conventions. The
+[stylelint-order] plugin can handle alphabetical declaration ordering. If
+adopting a shared config, it's worth reviewing its rules to make sure they
+reflect native CSS rather than Sass-specific conventions.
```
vendelal, line 389: Do we have a stylelint config we can share here?

css/README.md
```diff
@@ -2,113 +2,368 @@
...
+```
+
+For more on how specificity is calculated, see the [MDN specificity docs].
+
+[mdn specificity docs]: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity
+
+### Naming
+
+- Use lowercase and hyphens (kebab-case) for class names and custom properties
+- Be consistent with whatever naming convention the project is already using. Introducing a
```
vendelal: Should we mention that we prefer BEM if starting fresh or is that too opinionated? Maybe that's covered by the first bullet here?

vendelal (commented):
I'm so grateful you took the time to make these updates! I know I will reference this in the near future knowing it's current and there were some great tips in here that were new to me/that I often forget!

css/README.md
enatario: Possibly! Most of them have been around for long enough that I'd err on the side of recommending them, even if they get stale. Mostly want to ensure folks are bundling instead of using `@import` raw.

css/README.md
enatario: Hahahah breakpoint management is the bane of my existence so the less you have to use it the betterrrrr

css/README.md
enatario: We have our [thoughtbot one](https://github.com/thoughtbot/stylelint-config) but that is also woefully out of date 😭

css/README.md
enatario: Yeah we could be more strong about BEM usage!

css/README.md
vendelal: womp womp haha. Well next step could be to update that and then add it here in a future PR

css/README.md
```diff
@@ -2,113 +2,368 @@
...
-  - Start with the smallest viewport size and work upwards using
-    `min-width`/`min-height`
+- Prefer native HTML elements over custom implementations, like `<dialog>` for
+  modals, `<details>`/`<summary>` for disclosures, and the [Popover API] for
+  tooltips and non-modal overlays
 - Use [double colon syntax] for pseudo-elements
+- Vendor prefixes are rarely needed for modern browsers. If a project requires
+  legacy browser support, consider automating prefixes via a build tool rather
+  than maintaining them by hand
+- Prefer native CSS over preprocessors like SCSS
```
aspencer116, line 13: BIG DEAL

css/README.md
```diff
@@ -2,113 +2,368 @@
...
+Media queries can still be the right tool for page-level layout changes: things like large
+navigation shifts that also should effect other items in the page.
+
+- Start with the smallest viewport and layer upward with `min-width`
+- Prefer `em` units over `px` so breakpoints respect user font size preferences
+- Create breakpoints where the content starts to feel awkward, rather than
+  targeting specific devices
+
+**Note:** Unfortunately, you can't use a CSS custom property as a value within a media query, so be sure
+to keep track of your values here and perhaps scope it per partial.
```
aspencer116, line 189: so annoying!

css/README.md
```diff
@@ -2,113 +2,368 @@
...
+  `span_columns` or `spanColumns`
+- Avoid shorthand properties when setting a single value: `background-color:
+  #ff0000` rather than `background: #ff0000`
+- Use `/* */` for comment blocks
+- Keeping files under 100 lines makes them easier to scan; split larger files
+  into focused partials
+
+### Declaration ordering
+
+[Alphabetical ordering is a reasonable default] because it's predictable and easy to
```
aspencer116, line 407: I like the phrasing "reasonable default" which makes it clear this is a good choice but not the only option

css/README.md
aspencer116: I was going to ask this as well. Personally I think BEM could be a "reasonable-default" also worth linking to a resource here

css/README.md
```diff
@@ -2,113 +2,368 @@
...
+adopting a shared config, it's worth reviewing its rules to make sure they
+reflect native CSS rather than Sass-specific conventions.
+
+[stylelint]: https://stylelint.io/
+[stylelint-order]: https://github.com/hudochenkov/stylelint-order
+
+## Formatting
+
+- Use hyphens when naming classes and custom properties: `span-columns` not
+  `span_columns` or `spanColumns`
```
aspencer116: Use a BEM class format as an example?

aspencer116 (approved):
This is a great update 👏 thanks for working on this! Definitely worth promoting in a blog or social post once it's merged!

css/README.md
enatario: Also happy to add in other preferences like by property type! Thoughts?

css/README.md
enatario: Yup good call!

css/README.md
aspencer116: If you want! I prefer property type but I might be the only one here 😄

css/README.md
enatario: updated!

css/README.md
enatario: updated!

css/README.md
```diff
@@ -2,113 +2,375 @@
...
-  read,
-  - Avoid creating breakpoints that target specific devices
-  - Prefer `em` units instead of `px` for breakpoint values
-  - Start with the smallest viewport size and work upwards using
-    `min-width`/`min-height`
-- Use [double colon syntax] for pseudo-elements
-
-[autoprefixer]: https://github.com/postcss/autoprefixer
-[create breakpoints]: http://bradfrost.com/blog/post/7-habits-of-highly-effective-media-queries/
+- Prefer native HTML elements over custom implementations, like `<dialog>` for
```
pangolingo: This line feels like it belongs in the HTML guide instead of the CSS guide.

css/README.md
```diff
@@ -2,113 +2,375 @@
...
+  modals, `<details>`/`<summary>` for disclosures, and the [Popover API] for
+  tooltips and non-modal overlays
+- Use [double colon syntax] for pseudo-elements, single colon for pseudo-classes
```
pangolingo: An example would be nice:


```suggestion
- Use [double colon syntax] for pseudo-elements (`::after {}`), single colon for pseudo-classes (`button:hover {}`)
```

css/README.md
```diff
@@ -2,113 +2,375 @@
...
+`block_name`
+- Avoid shorthand properties when setting a single value: `background-color:
+  #ff0000` rather than `background: #ff0000`
+- Use `/* */` for comment blocks
+- Keeping files under 100 lines makes them easier to scan; split larger files
+  into focused partials
+
+### Declaration ordering
+
+[Alphabetical ordering is a reasonable default] because it's predictable and easy to
```
pangolingo, line 407: Can we suggest a tool (maybe stylelint) that auto-orders the declarations?

css/README.md
```diff
@@ -2,113 +2,375 @@
...
+[stylelint-order] plugin can handle alphabetical declaration ordering. If
+adopting a shared config, it's worth reviewing its rules to make sure they
+reflect native CSS rather than Sass-specific conventions.
+
+[stylelint]: https://stylelint.io/
+[stylelint-order]: https://github.com/hudochenkov/stylelint-order
+
+## Formatting
+
+- Use hyphens when naming classes and custom properties: `block-name`,
```
pangolingo: This line belongs in the naming section, and might be redundant.

css/README.md
```diff
@@ -2,113 +2,375 @@
...
+## Modern CSS features
+
+Native CSS has made significant advances. Before reaching for a JavaScript
+solution or a complex workaround, it's worth checking whether a native CSS
+approach exists.
+
+### Nesting
+
+Native CSS nesting is well-supported in modern browsers. It's useful for keeping
+component styles self-contained, though deep nesting (more than 2–3 levels) can
```
pangolingo: Can we make this recommendation firmer? "Don't nest more than 2-3 levels deep because it makes specificity and readability harder to manage."

sass/README.md
```diff
@@ -1,143 +1,3 @@
...
-
-- Atom users can use the [Sort Lines package], which provides commands and
-  keybindings for alphabetical sorting.
-- Sublime Text users can use the `Edit > Sort Lines` menu item, or press
-  <kbd>F5</kbd> to sort lines alphabetically.
-
-[sort lines package]: https://github.com/atom/sort-lines
-
-</details>
+We now recommend you use [native CSS](../css/README.md) in lieu of Sass/SCSS.
```
pangolingo, line 3: 💯

pangolingo (approved):
I love this update! 
I left some suggestions, but none of them are critical.

css/README.md
enatario: Yup good point especially now that updating that guide is on the docket!

css/README.md
enatario: Yeah that's a good point about being more opinionated on that (though I sometimes break that rule). I'll adjust

css/README.md
enatario: Yes! I refer to it in the linting section above and I can also ref it here.

brian-penguin (approved):
Great!

PR #802, opened 2026-06-12
Fix staging link in CSS guide

jaredlt:
There was a link to a blog on our staging site.

Now it points to production.


enatario (approved):
oops good catch thank you!

PR #803, opened 2026-06-12
Update CSS file structure to match Roux

jaredlt:
The guide mentions a `reset/` directory but Roux uses `resets/`

Update to match Roux


enatario (approved):
Amaze tysm for catching that

